PR #9, opened 2015-06-21
Registering STL Viewer

jneilliii:
Adding my STL Viewer plugin to the repository.


foosel:
Can you change the image referenced from "assets/..." to "/assets/" to fix the build error please? Then I'll merge. Would do it myself normally but not at a PC at the moment.

jneilliii:
I think I got it fixed.

foosel:
Jup, as you can see the build tool is happy :)

PR #18, opened 2015-09-23
Add EmailNotifier plugin

anoved:
This is my first effort at packaging a plugin. It is loosely based on the Pushbullet plugin but sends notifications via SMTP email instead.


foosel:
Nice! 

One thing I noticed while doing a quick look at the code, you'll want to not set [this line](https://github.com/anoved/OctoPrint-EmailNotifier/blob/master/octoprint_emailnotifier/__init__.py#L94). The Pushbullet plugin needs it since it needs a patched version of the library it depends on (since the version on PyPi has a bug and has still not been update with the fix I PR'd for that), but your plugin doesn't seem to need it (otherwise I'd expect the dependency link property to be set in the repository entry too) and processing with dependency links takes longer, so you really only want to set that if it's absolutely necessary.

anoved:
Thanks - fixed.

PR #20, opened 2015-10-10
Added TouchUI plugin

BillyBlaze:
Should work like:
 - http://billyblaze.github.io/plugins.octoprint.org
 - http://billyblaze.github.io/plugins.octoprint.org/plugins/touchui/


foosel:
Great work!

BillyBlaze:
Thank you very much :)

PR #23, opened 2015-11-09
Added plugin Autocalibration

platsch:
This is a plugin for automatic backlash calibration. Successfully tested on 3 printers so far.


foosel:
Neat!

PR #24, opened 2015-11-10
update autocalibration description

platsch:
Just fixed the multiple extruder offset problem and would like to remove the warning from the description.


foosel:
Sure, no problem

PR #31, opened 2016-01-19
added plugin "Autoscroll"

MoonshineSG:


foosel:
Yikes, I really don't know how that slipped through :/ I'll change the date to today so that it shows up right at the top of the homepage and also gets included in the RSS feed.

MoonshineSG:
👍

PR #33, opened 2016-04-13
Added plugin pushover

thijsbekke:


thijsbekke:
I wanted too receive notifications on my phone, and an e-mail just isn't good enough. So I build a plugin for the app that I use, Pushover.

PR #34, opened 2016-04-25
Added the redeem plugin

eliasbakken:


foosel:
@eliasbakken It's a bit embarrassing but I only noticed just now that you didn't specify an archive URL, making this plugin uninstallable through the plugin manager built into OctoPrint. Was this intentional or should I set this to `https://github.com/eliasbakken/octoprint_redeem/archive/master.zip`?

eliasbakken:
Yes, please do! I had forgotten about it completely, but it would be great to have this searchable.

foosel:
@eliasbakken done

foosel:
@eliasbakken The CI build for the repository is currently failing a check because it the external images from this plugin throw a 401 - any ETA when the server will be back in shape? Alternatively - could you maybe send an update that hosts the relevant images on the repository itself?

eliasbakken:
The server should be back up. I'll see if there is a way to host any images
somewhere else...

foosel:
Perfect, thanks for getting back to me! Re hosting, you could just drop the plugin's related images into `/assets/img/plugins/redeem` here on the repo (via PR). Or I can do it too.

PR #36, opened 2016-06-23
Updated plugin information

donovan6000:
Had to change M3D Fio's name to M33 Fio for legal reasons


foosel:
Hope everything is alright for you...

PR #39, opened 2016-07-04
added "Nautilus" plugin

MoonshineSG:


foosel:
Something is wrong with the linked screenshots there (plugins/marlin vs plugins/nautilus).

foosel:
Oh, also just saw that it requires the development version for now. Same as with the file manager pr I'm a bit reluctant to merge development only plugins into the official repo (especially if the requirements depend on 1.3.0dev instead of just 1.3.0). Then again the version limitation should actually prevent this from showing up out of the blue for people, so I guess I should revisit this stance and maybe just add some special warning to devel only plugins on the page...

MoonshineSG:
Or you can just merge it when the UIPlugin get's promoted to main branch... Any timeline for that ?

foosel:
ASAP ;) Currently hunting down bugs and writing docs. Will need to switch to a different working model for 1.4, too much switching back and forth between maintenance and development producing a lot of overhead all the time, hence everything taking too long for my taste.

PR #40, opened 2016-08-07
M117PopUp & M117NavBar Plugins

jneilliii:
Would like to register my plugin demonstrating the use of _plugin_manager.send_plugin_message and onDataUpdaterPluginMessage.


foosel:
I have NO idea why that PR slipped past me so long. Sorry for that :/

I'll update the dates in your plugin registrations so they show up right on top of the listing.

PR #41, opened 2016-09-01
Added plugin GCodeBar

mtowara:
as requested here
https://github.com/foosel/OctoPrint/issues/1456


foosel:
Hi @mtowara, nice work! Did a quick look at your code and noticed your update check isn't keyed to the plugin identifier but a differently cased variation thereof (gcodebar vs GCodeBar). I think that shouldn't cause problems, but I'd still suggest to adjust that just to make sure.

mtowara:
You mean the one in /extras?
That should be okay in the plugin repo, just need to copy it over to the main repo. (changed it after running cookiecutter)
Or did I miss some other spot?

foosel:
Nah, I meant the key in the get_update_information duct, see https://github.com/mtowara/OctoPrint-Gcodebar/blob/master/octoprint_GCodeBar/__init__.py#L36

mtowara:
ok, now I got it.
And I think it actually broke the update.
Thanks!
Also kudos for that cookiecutter template and the getting started doc, that was really super helpful :)

PR #42, opened 2016-09-02
Added plugin DetailedProgress

dattas:
Added plugin DetailedProgress


foosel:
Nice work! Small suggestion though, instead of using the gcode hook (which should actually not be bogged down with a lot of work, since that's part of code that needs to be fast, any processing delay here can cause some downs), you could just use a RepeatedTimer with 10s started on server startup and have that execute your work loop. See http://docs.octoprint.org/en/master/modules/util.html#octoprint.util.RepeatedTimer

dattas:
@foosel thanks for the recommendation. I've made your recommended changes.

PR #43, opened 2016-09-23
Slicer plugin

kennethjiang:
This plugin provides a GUI to slicer so that you can rotate (for now. scaling the model will be coming soon) the model, and overrides slicing settings.


foosel:
Very nice work!

PR #48, opened 2016-12-09
Nautilus description and links update

MoonshineSG:


MoonshineSG:
that was quick. thanks.

foosel:
Good timing ;)

PR #51, opened 2016-12-28
Feature/smsnotifier plugin

taxilian:
This is a plugin I wrote that uses the Twilio API to send SMS notifications. I'll probably add support for sending via google voice instead one of these days...


foosel:
Sorry for just getting around to this now, I was on a long overdue vacation.

I noticed that your listing is referring to a picture in the `assets` folder, but your commit doesn't contain that. Do you want to add it before I merge this? :)

taxilian:
Good point =] Fixed, and thanks for all your hard work.

foosel:
I took the liberty to set the date to today for some better visibility, since it took me until today to merge this thanks to my vacation. Hope that's alright :) Also: Nice work!

PR #57, opened 2017-02-07
Added plugin slic3r

eyal0:
This adds OctoPrint-Slic3r to the plugin repository.  Tested locally with jekyll following instructions.

Feel well.


foosel:
Three things:

  * Not sure how that empty `screenshots` section will behave and can't check right now (still out of commission), can you remove that just to be safe?
  * Would probably be better to push the linked image (Slic3r logo) into the repo too instead of hotlinking it - URLs not under our control can always change ;)
  * You forgot to put yourself as the author :)

foosel:
Author is still pointing at me (also in the plugin itself I just noticed). Considering that you are now maintaining it it would make more sense to have it state your name (or something like "Eyal Soha, based on work by Gina Häußge" or even just "Various" if that makes you more comfortable). Let's face it, I didn't touch this code in nearly two years, and most of what I did before is obsolete by now probably anyway, so having me started as author feels just wrong, plus it might make people expect support and insight on this plugin that at this point I no longer can provide ;)

eyal0:
* Leaving screenshots empty had no effect on the display but I removed it.
* I downloaded all images and put them in the commit.  Best not to have any external links.
* I changed the authors.  Javier did the most work, really.  I also changed the authors on the plugin itself.

foosel:
Perfect, thanks! :+1:

PR #63, opened 2017-03-10
Adding LEDStripControl Plugin

precision:
A small plugin that implements Marlin's version of the M150 Gcode command.  Intercepts the M150 command and instead controls local GPIOs (attached to Mosfets).


foosel:
🎉 :+1:

PR #78, opened 2017-05-11
Adding Grbl plugin

mic159:


foosel:
Failed check is due to an image from a different plugin that's hosted on a server that's currently apparently down for maintenance.

Ignored this and merged, thanks :)

foosel:
Short heads-up - the entry was incomplete (which I didn't see at first, long day with release ;)), I [fixed that](0aaf94db1dd8ba652632768f224ec206d9f1643f), hope that all is correct. Additionally made sure the `.md` file [has the same id as your plugin](489f73f53fa4993e6a4e8f4d1d0b8f29f8a2d8d8), to avoid conflicts.

PR #82, opened 2017-05-27
New plugin !

art29:
My plugin lights up led depending of the pourcentage that the print is at !

### DIFF
--- a/_plugins/OctoRGB.md
+++ b/_plugins/OctoRGB.md
@@ -1,0 +1,89 @@
+---
+layout: plugin
+
+id: OctoRGB
+title: OctoRGB
+description: This is a plugin that power up rgb leds depending of the pourcentage of where is the print at.
+author: Arthur Fetiveau
+license: AGPLv3
+
+# TODO
+date: today's date in format YYYY-MM-DD, e.g. 2015-04-21
+
+homepage: https://github.com/art29/OctoRGB
+source: https://github.com/art29/OctoRGB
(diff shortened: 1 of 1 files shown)


foosel:
The description of your plugin is still "TODO", the date is also not set properly. Your plugin's README is also completely empty :/

A brief look at your plugin's source code looks like it might need some work. 

As it is implemented now it will not work generally since you utilize the REST API accessed through a local URL that might or might not be available this way. Using the REST API here is like driving around the world to get to a point near you - your plugin runs inside the same process as your OctoPrint instance and can utilize a lot of built in and injected APIs to retrieve the information you are seeking.

Additionally your approach to monitoring the print completion (busy loop in the main thread constantly querying data) isn't the most efficient approach. Take a look at the [ProgressPlugin mixin](http://docs.octoprint.org/en/master/plugins/mixins.html#progressplugin) instead, that's exactly for these kinds of things :)

If your plugin works for you as-is, that is great! But before you publish it on the official repository for others to use I fear it really needs some more work and streamlining, or people installing it as it is now might run into issues.

foosel:
Before I forget: Feel free to get in touch on the mailing list or also the IRC channel if you have questions regarding plugin development - there are people there (including myself) who might be able to help with any questions you might have! You can find the links to that in the footer on http://octoprint.org

kantlivelong:
Closing. Please submit a new PR if desired.

PR #88, opened 2017-06-23
Add PolarCloud plugin

markwal:
If possible, I'd like to defer the merge until June 25, but I submit now so there's a bit of time to look it over.


foosel:
First look looks good. You should have write access to this repository as well, so just merge yourself whenever you feel comfortable about it (June 25th is a sunday and hence "I'm not touching anything OctoPrint-related" weekend time so the earliest I could do it is June 26th).

PR #95, opened 2017-07-04
Remove link to anywhere.io from the doc

kennethjiang:


foosel:
To be honest, personally I think that link was just fine. But the "Get Started" button indeed is a bit confusing, so better err on the side of caution and not risk people getting confused.

PR #98, opened 2017-07-13
BLTouch Plugin

jneilliii:
Plugin that adds BLTouch controls to the Control tab. Let me know if there are any changes that need to be made.


foosel:
Only the date really, you've listed the plugin as published in January ;) I'll fix that to today's date if you don't mind :)

PR #102, opened 2017-07-23
YouTube Live Plugin

jneilliii:
Adding YouTube Live plugin that adds a tab with embedded YouTube Live Stream from configured channel in settings.


foosel:
👍 

Squash-merged because there was this weird merge commit in there.

PR #111, opened 2017-09-02
Added plugin ZERO

gkolozof:
Configures compiling and install FirmWare for any 3D printer in 1 Min.


gkolozof:
Please let me know what should I do to activate my plugin as this is my first  plugin and I am not familiar with the process.

foosel:
I usually take a look at plugins that are submitted by be authors, to make sure there are no obvious issues with the plugin implementation that might produce problems for people out there. Since I'm currently down sick, I haven't yet gotten around to do this and prepare proper feedback. As I wrote in my mail, I'll try to get around to do that as soon as I'm back on my feet.

gkolozof:
Ok thx...
i remove the deletes all logs, is for fast restart the octoprint
 but i need access to haproxy.cfg and rsyslog.conf for monitoring and redicrect url to local

gkolozof:
Ok i.m ready!!! no script no haproxy no rsystem.
I've made all the changes you've asked me,what do I do?

gkolozof:
Hi I followed your instructions and I pushed the changes to "OctoPrint-ZERO
https://github.com/gkolozof/plugins.octoprint.org-1",
Is everything ok or do I need to do an new pull request?
As I haven't got any response yet.

2017-09-07 11:46 GMT+03:00 Gina Häußge <notifications@github.com>:

foosel:
I'm still sick.

gkolozof:
I am sorry to hear you are still sick. Hope not to be something serious and
to get better soon.

2017-09-25 12:52 GMT+03:00 Gina Häußge <notifications@github.com>:

foosel:
I finally got around to take another look at this...

  * Your plugin is still doing stuff directly on plugin load, outside of the plugin framework and plugin folders (see your update and lock stuff, among other things).
  * The JS components of your plugin aren't contained in your plugin's view model. You are cluttering the global namespace with loads of identifiers that might or might not overwrite important other stuff.
  * You redefine `serialList` instead of reusing it. You then abuse it (the first found port must not necessarily be a printer!)
  * Your plugin allows any user to flash a printer, not just admins.
  * Your plugin still fetches data from an arbitrary/obscure IP based host on the internet. This is completely intransparent. Could be anything really.
  * The update configuration is still broken ("systemcommandeditor" as key instead of "ZERO").
  * It still tries to install avrdude for the user (without checking whether this is even necessary OR possible). Again, this is not something your plugin should do. It's something you should instruct the user to do in your README/from a friendly and not annoying message triggered by your plugin. And NOT by sending a AJAX call outside of your view model!
  * There's some avrdude.exe file now in your plugin's sources - what is that supposed to do? Same with the included avrdude.conf
  * You are opening file handles left and right without correct error handling. That will cause issues in various cases and lead to leaking file handles
  * Your plugin reports data to your frontend by writing to a static file. You might not be able to write to that file. And you are effectively blocking the whole server while that file is being read by the frontend. This is not how you push data to the frontend in OctoPrint's plugin infrastructure - that's what [`send_plugin_message`](http://docs.octoprint.org/en/master/modules/plugin.html?highlight=send_plugin_message#octoprint.plugin.core.PluginManager.send_plugin_message) on the backend combined with on `onDataUpdaterPluginMessage(plugin, message)` on the frontend are for.
  * Your plugin reconfigures the frontend's caching strategy to never cache - there is absolutely no reason for that and you are effectively killing performance everywhere this plugin is installed.
  * You still claim it works with any printer even though it only works for Marlin Kimbra firmware. That strongly limits the set of usable printers!
  * Your plugin's listing still doesn't even begin to explain what the plugin actually does. You state it "flashes firmware of every printer in less than a minute" - nowhere does it say that it requires a third party webservice for that, that it is limited to only that specific sub flavor of firmware and that it most definitely will only work with a) printers that are flashable via avrdude (not all are) and b) printers that this firmware supports in the first place.

And that is only the tip of the ice berg after a first look. I lack the time for a deeper review but based on structure and first glance there is a lot more.

As much as I hate to say it, this is not a plugin that is mergeable. There is simply so much to do and fix there to make it fit for the masses that rewriting it would be faster. I understand it works for you, and that's great. But it really is nowhere near ready for general availability on the plugin repository. Sorry.

gkolozof:
Thanks for your valuable support and comments. I'll try to apply your
suggestions.

2017-10-05 17:33 GMT+03:00 Gina Häußge <notifications@github.com>:

foosel:
Hm, I think I need to clarify something here. That list wasn't complete - as I said, those were just the things that jumped to my eye immediately when I took another look at your code, and I lack the time for a full blown review that would unearth even more points. Based on what I saw I do not have the feeling that this plugin in general is in a good shape for general consumption. There are simply too many shortcuts and too many assumptions about the used system and printer 

I'm sorry but I would strongly suggest that you try your hand on something smaller and less complex first, and also take the opportunity and look a bit more into tutorials on writing Python, best practices, the value of code readability and the importance of error handling. Personally I can recommend "Effective Python" by Brett Slatkin or "Fluent Python" by Luciano Ramalho, but there's also a ton of online sources on this topic, in various languages and formats.

From my personal experience I can tell you that writing something that just works for yourself is often very different and also easier than writing something that works well for everyone. So questioning your own assumptions when writing code and making it flexible enough to cope with an environment that is not exactly yours is very important if you plan on sharing your work.

gkolozof:
I agree with U, it was only a test to see the reaction, now I'm rewriting
all the code, 90% I see that it's included in youre functions, sorry that
your manuals have poor examples. thank U for support!!

2017-10-09 17:53 GMT+03:00 Gina Häußge <notifications@github.com>:

gkolozof:
But like after I made all the changes you asked me and no longer have anything to say to me, do you throw me out? and this the spirit of the open sorce ?? you are so scared !!! but have you ever tried it? It's beautiful ...... shame because it was a nice project ....

PR #113, opened 2017-09-05
Attempt to fix install issue by using github tarball

mic159:
Same as all the other plugins.
Seems some people couldn't download the file from pypi.


kantlivelong:
Merged

PR #121, opened 2017-10-29
Added plugin GcodeEditor

ieatacid:


ieatacid:
Did I get it right? >.<

kantlivelong:
PR looks good and this plugin seems useful. The only concern I have is that it seems to load the entire file in memory on the browser which can be a performance problem.

I'll leave this up to Gina to merge.

ieatacid:
> The only concern I have is that it seems to load the entire file in memory on the browser
What other way to do it? Stream only the 30 or so lines being displayed at a time? I don't know how to do that, nor have the time to learn :\  Wouldn't OctoPrint still have to load the entire file?

ieatacid:
Or maybe determine that limit based on system RAM. I haven't had any problems so far, though. It just takes longer to load a larger file.

ieatacid:
I actually came across that issue when I was searching for a gcode editor plugin :)

Thanks for the suggestions and explanations. I'll have more time to work on it this weekend.

PR #123, opened 2017-11-11
Added Themeify plugin

Birkbjo:


Birkbjo:
Seems like the owner of  https://github.com/mikerybka/OctoPrint-Lani removed the repository, causing html-proofer to fail.

foosel:
First of all: Nice work on that plugin! I noticed [this bit in there](https://github.com/Birkbjo/OctoPrint-Themeify/blob/3fc5fcc6f1c24881bc441f241638c70524ab0ab1/octoprint_themeify/static/js/themeify.js#L15-L20) though which kinda seems to do something else? Left-over from personal version? :)

About the build failure, I'll take care of that, feel free to ignore it.

Birkbjo:
Right, I will remove that - that is going to be another plugin :)

foosel:
Yep, perfect! 👍

PR #125, opened 2017-11-14
Withdraw EmailNotifier from plugin registry

anoved:
I cannot maintain [OctoPrint-EmailNotifier](https://github.com/anoved/OctoPrint-EmailNotifier) and regret user frustration caused by its broken/WIP state.


kantlivelong:
Looks like your branch was out of date but I took care of it anyway.

PR #130, opened 2017-11-24
OctoPrint ZERO

gkolozof:
Configures compiling and install FirmWare for 3D printer in 30 Sec.
 3D printer compatibility:
  - CARTESIAN (Prusa,Mendel,Taz,etc)
  - DELTA (Rostock/Max,Kossel,Cerberus,etc)
  - H-Bot/Core
  - Morgan SCARA
  - Makerarm SCARA
  - Wanhao Duplicator
  - Anycubic Linear Plus
  - etc


kantlivelong:
Hi,
I've compared your changes since the last PR and while there are a lot of things to list these stand out the most:

- It looks as if you have moved all of your code to `__plugin_load__` instead of being within `ZEROPlugin`.
- Perhaps I am missing something but it appears that this plugin **re-flashes** firmware on every start of OctoPrint. I can't think of a valid reason for doing this **ever**.
- Your plugin still fetches data from an arbitrary/obscure IP based host on the internet. This is completely in-transparent. Could be anything really.

gkolozof:
I made all the changes you asked me, now what's the problem?

gkolozof:
I made all the changes you asked me, now what's the problem?

gkolozof:
I applied all the changes you suggested so far. Can you please let me know what is left to do?
       THX...

kantlivelong:
Hi @gkolozof,

Generally we don't have to do code reviews for plugin submissions so this is on our list but currently on the back burner since this is is not the first or second iteration of doing this.

kantlivelong:
See https://plugins.octoprint.org/help/registering/ for more information.

kantlivelong:
Additionally https://github.com/gkolozof/OctoPrint-ZERO/blob/c9f9e4cd7d621fe6f230e756275f19bb854f0c3c/octoprint_ZERO/__init__.py#L34 needs some security added to it. From what I understand anyone could call http://[OCTOPRINTSERVER]/fw and initiate a flash.

Findus23:
@gkolozof 
The issues I am having with you plugin: (I know little about 3D printer firmwares, so maybe I am misunderstanding something)
- Which firmware does the plugin install?
- How do I know my printer is supported?
- Why should I trust `http://gkolozof.xyz/0/fw.php to send the correct firmware for my printer?
  - it doesn't return anything for me
- Why is your plugin sending requests to `http://gkolozof.xyz/0/dw.php?dw=[INT]`
  - It only returns `?>`
  - if this is some kind of analytics, it would be nice to inform the user about it.
- How can I be sure that no one is intercepting the download as it isn't using HTTPS?
- (more probable) How can I be sure my internet didn't have a hickup in the wrong moment and the downloaded file is corrupted?

gkolozof:
Thanks for your support, I'm making the changes

gkolozof:
OK I made all the changes you asked me, I hope that now and everything is fine, you are fantastic, I thank you very much for your support

Findus23:
@gkolozof 
I have struck through what has been fixed, but maybe you could answer (or better document it in the description/README) the rest, if you have some time.
Thanks

gkolozof:
I have updated the changelog in Readme.md. Can you let me know if it clear
enough? Is there anything else I need to do?

2018-02-06 11:33 GMT+02:00 Lukas Winkler <notifications@github.com>:

Findus23:
The most important questions I still have are:
- Why are you sending data to `http://gkolozof.xyz/0/dw.php?dw=[INT]`?
- Why are you making it technically possible for your server to take over every Octoprint instance that has the plugin installed? 
(Not saying that you are planning to do so, but you should make it technically impossible by not inserting foreign HTML into the site)

Otherwise you'll need to ask @kantlivelong, I just looked into the code out of interest.

foosel:
As already pointed out by @kantlivelong in an earlier round, your PR still contains stuff in `_site` that don't belong, and some `.bundle/config` file that doesn't belong either. And your images are in the wrong folders too.

But this is only the tip of the iceberg because after having taken another look at the current iteration I come to the same conclusion as @Findus23 - there are still/again some glaring issues here. 

I [said it](https://github.com/OctoPrint/plugins.octoprint.org/pull/111#issuecomment-335182544) in your initial pull request #111 back in September - the task this plugin is supposed to accomplish is somewhat complex and certainly not the best one to get ones feet wet in developing plugins for OctoPrint or really development in general. Errors or bugs can be very very fatal, for the health of the printer, the OctoPrint instances this would be deployed to, and so on and so on. We've now gone through several review rounds, and every time you removed one glaring issue either another one was added or surfaced since it became more visible. And frankly, this simply takes too much time. I don't see that more months of back and forth on this and one review after the other (also with repeated statements that get ignored/overlooked) will get this into a shape that I'd feel comfortable having this on the official repository.

I know this sucks, you have spent a lot of time and effort on this as well, and this is not the outcome you were hoping for here. But I'm sorry, this won't get merged, and please do not open further pull requests for this plugin either. Both me and @kantlivelong and I guess also @Findus23 now have spent several hours on one review after the other, and at this point we all have to face the fact that this is going nowhere.

kantlivelong:
Agreed :disappointed:

gkolozof:
The opt (chk) function sends to the server
http://gkolozof.xyz/0/dw.php?dw=[INT] a number that corresponds to the
status of the Firmware UpLoad process to know if everything went well, see
Table:
# fn opt(chk) send status code
# status code:
# 0-100 % FW UpDate
# 500 = Serial Port Close
# 1000 = Serial port failed
# 1500 = FW DW error
# 1600 = FW error
# 2000 = Proccess faults

THX...


2018-02-06 14:47 GMT+02:00 Lukas Winkler <notifications@github.com>:

PR #132, opened 2017-11-28
OctoPrint-AstroPrint Plugin

CoDanny:
This plugin allows OctoPrint users to print and monitor files from the AstroPrint Cloud


foosel:
Hi! 

I like to give plugins of first time registrants a quick look over, to make sure they don't do anything weird that might cause issues for users out there. I did this here as well and I have a couple of points:

  * [This endpoint looks fishy](https://github.com/AstroPrint/OctoPrint-AstroPrint/blob/master/octoprint_astroprint/__init__.py#L407-L444). I'm not entirely sure what you are checking there (the flow is a bit hard to follow), but considering that you are handing out the global instance API key that has admin rights attached, there should at the very least be a limitation to logged in admin users which I'm not seeing. That has to be changed.
  * Is there a particular reason your plugin re-implements a lot of the core APIs (and doesn't even appear to use those endpoints then)? Has me a bit confused over here (and I'd like to get that clarified before merging this, since it might create problems moving forward, considering that we are looking into tightening down permissions which those endpoints would then circumvent).

    For reference: https://github.com/AstroPrint/OctoPrint-AstroPrint/blob/master/octoprint_astroprint/__init__.py#L488-L806 and https://github.com/AstroPrint/OctoPrint-AstroPrint/blob/master/octoprint_astroprint/__init__.py#L814-L866
  * Your [sent handler](https://github.com/AstroPrint/OctoPrint-AstroPrint/blob/master/octoprint_astroprint/materialcounter/__init__.py#L67) does some regex parsing on the sent commands that adds a certain overhead. As you see from the [handler signature](https://github.com/AstroPrint/OctoPrint-AstroPrint/blob/master/octoprint_astroprint/__init__.py#L301), the gcode is already parsed out for you. Better reuse it instead of parsing it again and slowing down the send loop.

Could you take a look at those? Thanks!

CoDanny:
Hi @foosel,

Thanks for taking the time to look this over and pointing out ways to improve. Here are my answers:

**AccessKeys endpoint**: This is needed in order to give the AstroPrint mobile app local access to the instance. The AstroPrint user logged in the app and the user logged in the plugin are checked to be the same. We can for example, restrict it so that only Admin users can log an AstroPrint user into the instance. 

**Re-implemeting of APIs**: Again this is so the AstroPrint Mobile app can talk to the instance (protected as explained above). We have tried to use as many of the standard APIs as possible but some didn't return all the data we needed. We can revisit this if / when the standard APIs are changed.

**Command parsing**: Our mistake, will improve this.

foosel:
**AccessKeys endpoint**: That sounds good. You could alternatively also use the user API key instead of the global instance API key. Or generate your own (THAT though requires a hook that will only be available starting with OctoPrint 1.3.6). But handing out the global instance API key without checking for login state and admin rights is an accident waiting to happen.

**Re-implementing of APIs**: Hmm 🤔 On a first look through those endpoint implementations I didn't see any additional data, that is the thing. In fact, they look more or less 1:1 identical to the core implementations (or rather, historical versions of those). The only difference I so far found was an additional endpoint `printer/fan` to send fan commands and the printer state being included in the job response. Have you cross referenced with [the current API documentation](http://docs.octoprint.org/en/master/api/index.html) of what is actually missing?

**Command parsing**: :+1:

foosel:
@CoDanny I'm a bit busy with 1.3.6 but:

- Restrict access to the plugin interface to admins: https://github.com/AstroPrint/OctoPrint-AstroPrint/pull/4 - 👎 This still doesn't do anything to protect the `getAccessKeys` method. This is the one that's most critical here since it happily hands out the *global OctoPrint API Key* without checking if the user should have access to it (or is even logged in). This must be restricted to logged in admin users only, and the only thing I see it doing there is check if global access control and your plugin's access control flag are set and if so deny a request from a user not logged into *AstroPrint*, but no check at all for the actual *OctoPrint* login session/current user permissions.

  https://github.com/AstroPrint/OctoPrint-AstroPrint/blob/c1a166af7f8f5ad1ba78cfea581151e029267383/octoprint_astroprint/__init__.py#L419-L456

  Or am I missing something here?

- Removal of API that were already present in OctoPrint: https://github.com/AstroPrint/OctoPrint-AstroPrint/pull/2 - 👍 
- Removal of unnecessary parsing of GCODE commands: https://github.com/AstroPrint/OctoPrint-AstroPrint/pull/3 - 👍

CoDanny:
Thanks for checking @foosel. I have now merged the PRs that you approved. We're getting closer ;) Let me just clarify two things about the use of the API in the remaining PR:

1) The API is called from our mobile app as a REST API call. This means that we don't have an octoprint user session during this call.

2) The code checks that the astroprint user logged in the mobile app is the same astroprint user that the octoprint admin linked their box with. So it doesn't simply hand over the Access Key, it first checks that the astroprint user requesting it is the same astroprint user that the octoprint admin previously allowed access.

Note that with this PR, ONLY admins can link their octoprint boxes to an AstroPrint account. The PR also modifies the copy where they're told what this means.

The image below is taken before the PR:

screen shot 2017-12-05 at 10 46 41 am

The change in copy is this: https://github.com/AstroPrint/OctoPrint-AstroPrint/pull/4/files#diff-c6209f84a8930edfeba32ac33cbdc932L67

foosel:
What I'm not entirely clear here is why you even need an endpoint on OctoPrint's side here that provides your mobile app with an API key (be it admin or something else). I thought the whole point of this plugin was to connect the instance to your cloud and then work over that? I take it your mobile app also is able to leave out the cloud and stay local or something? Could you maybe sketch out the intended work flow a bit here and especially (roughly) what gets communicated where and when by which party? I'm not asking for a full blown sequence diagram but more the 50000 foot view. I'm trying to figure out what you need the key for and if we can't maybe achieve this in a way that doesn't involve handing out the global API key (maybe by handing out a user key instead, or a key generated specifically by your plugin, something like that).

I find myself a bit between a rock and a hard place here to be honest because I understand that you currently have a requirement to expose the API key like this, but I honestly at the current time can't validate that this key can in fact never ever under any circumstances leak, since that relies entirely on whether there exists an Astroprint user session or not and it's not possible for me in the available time to verify that this in fact can never get "left-over" or something like this and such create a problem.

I hope you understand where I'm coming from and what my problem is here.

---

Quick side note because I just stumbled over this, you create your SQLite db in the logs folder: https://github.com/AstroPrint/OctoPrint-AstroPrint/blob/c1a166af7f8f5ad1ba78cfea581151e029267383/octoprint_astroprint/AstroprintDB.py#L11

The plugin data folder definitely would be more fitting here since it was created for this exact kind of purpose. You also might not want people to download that file as is possible with logs in general. See http://docs.octoprint.org/en/master/modules/plugin.html#octoprint.plugin.types.OctoPrintPlugin.get_plugin_data_folder

CoDanny:
@foosel Thanks for the pointer on the sql lite file. Makes total sense, I missed that. I have opened an issue to fix this: https://github.com/AstroPrint/OctoPrint-AstroPrint/issues/6

The plugin's purpose is, as you mentioned, mostly to connect the OctoPrint instance to our cloud, however we do also have a mobile app which can communicate with the printer over local network. Furthermore there are certain operations ( such as moving print head and extrusion ) that we only allow when the phone and the octroprint instance are in the same network. It is for this local network communication between the mobile app and the octoprint instance that use the plugin ( and generic ) APIs and for which we need the access key.

Here's a diagram.
screen shot 2017-12-06 at 10 06 49 pm

Just a little more information about the use of the key: 

Our mobile app, upon detecting the OctoPrint instance via BonJour, will query it to see if the AstroPrint User logged into the instance's AstroPrint Plugin is the same as the one logged into the mobile app, the query is performed using the endpoint under discussion here. 

If the users match, it will show the instance in the list of locally available printers in the "printers" section of the app. When a printer is available locally, some extra operations are available, like moving head and extruding. The access key is retrieved at this point and only stored for that mobile app session, in the app's memory. It's never saved or uploaded anywhere. It's only used for further communication with an authenticated octoprint instance. 

In fact when the user leaves the "printers" section of the mobile app, the access key is removed from memory. Next time the user enters that section, the same sequence is done all over again.

foosel:
Ok, I think I understand now what this endpoint is used for. Huge thanks for clarifying!

How about simplifying the logic in that one a bit more:

``` python
@octoprint.plugin.BlueprintPlugin.route("/accessKeys", methods=["POST"])
def getAccessKeys(self):
    email = request.values.get('email', None)
    accessKey = request.values.get('accessKey', None)

    if not email or not accessKey:
        abort(401) # wouldn't a 400 make more sense here?

    if self.user and self.user.email == email and self.user.accessKey == accessKey and self.user.userId:
        # only respond positively if we have an AstroPrint user and their mail AND accessKey match AND
        # they also have a valid userId
        return jsonify(api_key=self._settings.global_get(["api", "key"], 
                       ws_token=create_ws_token(self.user.userId)))

    # everyone else gets the cold shoulder
    abort(403)
```

As far as I understand you, that would still work fine (the only difference is that the state of whether access control is even enabled or not doesn't play into the result anymore). It limits the return of that data to clients who can provide the correct `accessKey` and account mail address. This of course assumes that the accessKey stays secret.

Which brings me to another question:

https://github.com/AstroPrint/OctoPrint-AstroPrint/blob/c1a166af7f8f5ad1ba78cfea581151e029267383/octoprint_astroprint/__init__.py#L310-L316

If I understand this correctly, this is the method used to set said `accessKey` (or `access_key` in the context of the plugin instance) which then gets attached to the user object in the cloud connection instance. As long as you have an OctoPrint API key, you can freely set this plugin side version of the `access_key`. Any API key, even the anonymous UI api key suffices here, no user or better admin rights check is done. And this is the key that gets copied to the user instance:

https://github.com/AstroPrint/OctoPrint-AstroPrint/blob/c1a166af7f8f5ad1ba78cfea581151e029267383/octoprint_astroprint/AstroprintCloud.py#L150

and against which it's then tested if whatever client has that key and the correct mail address is allowed to get the global API key with admin rights.

What I'm fairly sure here is that it's possible to freely set the plugin's `access_key` due to that endpoint not being limited. What I'm not sure is if it's in fact possible to get `self.user.accessKey` synced to that freely chosen key. If so, that would allow to just POST to that endpoint, set the accessKey to a known value (`test123`), then POST the mail + that chosen accessKey to the other endpoint and now have free reign over the instance through the obtained API key.

So - next workflow question - where does this accessKey usually come from (Cloud, App?), does it even need to come over a REST API and might it not instead come from the Cloud on initial connect, and if it needs to come over that endpoint, can this get tightened down to admin level as well?

CoDanny:
Hi @foosel,

I have another pull request over at the plugin repo: https://github.com/AstroPrint/OctoPrint-AstroPrint/pull/7 which addresses your concerns and suggestions. Thanks again!

Let me know if this is enough and will merge the pull request.

foosel:
@CoDanny First of all sorry for the delay - release overhead.

That looks good now from the perspective of the access key.

I'm slightly confused though because now two API endpoints that are already available in core made their comeback(?). See https://github.com/AstroPrint/OctoPrint-AstroPrint/blob/d9e5c18e114bdaa42203fb6fd4bf6192e9c1e430/octoprint_astroprint/__init__.py#L509 and https://github.com/AstroPrint/OctoPrint-AstroPrint/blob/d9e5c18e114bdaa42203fb6fd4bf6192e9c1e430/octoprint_astroprint/__init__.py#L494. And those and the additional one at https://github.com/AstroPrint/OctoPrint-AstroPrint/blob/d9e5c18e114bdaa42203fb6fd4bf6192e9c1e430/octoprint_astroprint/__init__.py#L476 again lack access control. So with this plugin in place anyone can start/cancel the print job (not just logged in users, anyone) or manipulate the fan settings. I'm pretty sure that's not intentional.

Wasn't that fixed in another PR? Or maybe it wasn't and I missed it - if so, sorry.

Just for clarification, `@restricted_access` only tests if the client has a valid API key, it does NOT do user specific access control of any kind. See also its docstring.

But I think once *that's* solved we'll have it :)

CoDanny:
Thanks @foosel,

My bad on the decorator. I think using ```@admin_permission.require(403)``` instead would be good, correct?

I'll check with the team why those endpoints were left there. They were not removed in the previous pull request. Here's that pull request for reference: https://github.com/AstroPrint/OctoPrint-AstroPrint/pull/2/files

foosel:
Either `@admin_permission.require(403)` or `@user_permission.require(403)`. The "job command" and "send command to printer" API endpoints are both limited to users in core OctoPrint, and I'd classify the fan command one as a very specific "send command to printer" endpoint.

See also http://docs.octoprint.org/en/master/api/printer.html#send-an-arbitrary-command-to-the-printer and http://docs.octoprint.org/en/master/api/job.html#issue-a-job-command

foosel:
Now that is embarrassing:

```
def restricted_access(func):
	If you decorate a view with this, it will ensure that first setup has been
	done for OctoPrint's Access Control plus that any conditions of the
	login_required decorator are met (possibly through a session already created
	by octoprint.server.util.apiKeyRequestHandler earlier in the request processing).

	If OctoPrint's Access Control has not been setup yet (indicated by the "firstRun"
	flag from the settings being set to True and the userManager not indicating
	that it's user database has been customized from default), the decorator
	will cause a HTTP 403 status code to be returned by the decorated resource.
```

Sorry, apparently I was one step ahead of myself here and `restricted_access` does *indeed* check for an existing login session after all (not specific rights, but if I remember correctly an anonymous user would not get a `True` here). It's time to call it a day I think...

CoDanny:
@foosel we have updated the pending pull request at the plugin repo with this commit that should address the remaining issues: https://github.com/AstroPrint/OctoPrint-AstroPrint/commit/3c4746e

foosel:
Looks good!

CoDanny:
Thanks @foosel. Merged!

foosel:
Merged. I also took the liberty to fix a broken image link that was still left over (".jpg.jpg" on featured image).

CoDanny:
Thanks for all your help!

PR #133, opened 2017-12-01
Add 4 plugins extra dist, auto collapse, fan slider, sidebar e-stop

ntoff:
* Extra distance plugin, for more movement distance options
* Auto collapse plugin to automatically  collapse the files accordion after a user configurable timeout
* Fan speed slider plugin to control the speed of the parts cooling fan after seeing a few requests for that ability
* Big ugly red e-stop button in the side bar for, well, emergencies. Also user configurable in case the emergency stop command is something other than M112


foosel:
The check failed because a link in some other plugin had a hiccup, not your fault.

ntoff:
I hadn't even noticed it had failed, so there's nothing I need to do?

foosel:
Nope, all's well & published

PR #139, opened 2017-12-24
Added plugin Power Failure Recovery

pablogventura:
This is a plugin to recover a print after a power supply fail. It generates a new gcode to resume printing after the fail.


kantlivelong:
Is this specific to the Prusa MK3? If so I think it would be good to clarify that in both the name and description.

pablogventura:
It is not specific to mk3. It works with any printer as it recovers the unprinted gcode and injects gcode to leave the printer in the pre-power failure state.

kantlivelong:
Ah I see now that I've looked a little closer. So this is intended to resume from a "close enough" point in time and not exactly when it stopped, correct?

sbts:
@pablogventura After having a look at the code, I can't see very many scenarios where this has any chance of working outside a contrived example.
Only tracking what's been printed once every second will very easily leave you with a completely incorrect restart point.
You MIGHT get closer by tracking which line was last acknowledged by the printer, however, every printer has an internal Queue and you have no way to know what's in the queue that hasn't yet been printed.
In other words, there is absolutely no way to know where to resume from automatically.

You could ask the user to take some exact measurements, then calculate the layer number.
From the layer number you can find the correct point in the gcode for the start of that layer.
That will sort of work. but will always reprint part (or all) of that layer resulting in a blemish on the print.

Even worse, on some (probably many) printers, Z height will be unknown after a power failure and you can't "home" the Z axis if home is when the head is closest to the bed (all printers I'm aware of) as that would crash the head into the work.

The reason Z is unknown after a power fail is that the weight of the moving part of the Z axis (bed in some printers, x/y carriage in others) can cause the axis to move when it's stepper is unpowered.

All in all, a helper tool to allow a user to skip already printed layers may be useful, but beyond that, there are more cases that will fail, than cases that will work for an auto resume tool.


NOTE: all of the above is my personal opinion, @foosel may feel differently and I will be interested in her comments.

pablogventura:
Hello @sbts ! In particular, in cases where I have tried the plugin has behaved very well, almost without leaving any mark on the print. My printer is a handmade CoreXY, running Marlin.

Your idea to track which line was last acknowledged by the printer is very good! Would it not be enough to know the size of the printer's internal queue to know with enough accuracy the line being printed?

On the other hand, perhaps I could add the option to generate a gcode for the continuation of the print, but as a separate part to glue them. In order to avoid the case of printers where the z-axis falls by weight.

Thank you very much for your careful comments!

foosel:
I agree with the points raised by @sbts. The firmware buffer (whose size is usually unknown, as is in most cases its fill rate - and btw, you actually have up to three consecutive buffers here) and also the send queue on OctoPrint's side itself (whose processing relies on getting `ok`s back from the printer) mean there's a delay between your current position in the file at any given moment and what the printer has actually already printed.

Now, with that being said, this plugin's approach - while extremely coarse - may work more or less, depending on the model and printer and a lot of other factors. What you should be aware here is that if this gets published, you can count on very very angry people coming down your throat sooner or later who depended on this to work for their interrupted 50h print but then it didn't. Are you prepared for that? :)

From a review perspective though: Please don't abuse the settings as a data storage. You should persist your recovery data to some file in [your plugin's data folder](http://docs.octoprint.org/en/master/modules/plugin.html#octoprint.plugin.types.OctoPrintPlugin.get_plugin_data_folder), not trigger a settings update across all clients every second (and hence additional completely unnecessary load on the server that should ideally focus on pushing a print to the printer).

foosel:
Closing since no response from the author 😕

pablogventura:
I'm sorry! I've been very busy in college and now I'm moving into a new
apartment. As soon as I get some free time back, I'm still on it.
Thank you for your time!

PR #142, opened 2018-01-02
Added plugin TerminalCommands

ieatacid:
Lets you add custom G-code command buttons to OctoPrint's Terminal view tab (currenty only supports G-code). They're inserted between the terminal command input and the terminal filters.


kantlivelong:
Good idea

PR #147, opened 2018-01-06
Add Firmware Updater plugin

benlye:


foosel:
Great! :+1: And sorry for the delay, vacation + broken build server makes for a tricky mixture ;)

PR #149, opened 2018-01-07
Added plugin M73 Progress

cesarvandevelde:
Most 3D printers offer a progress bar or percentage indicator on
their display. However, by default, this percentage is only updated when
printing from SD card. When printing directly through OctoPrint, the progress bar
remains empty. This small plugin remedies the situation by injecting
`M73 (set build percentage)` commands into the print stream.


foosel:
Very nice! :+1: Apologies for the complaints of the Travis build, something hiccuped hard in there unrelated to this PR... I'll take a proper look and merge this ASAP!

PR #157, opened 2018-01-18
Update Firmware Updater plugin details

benlye:
New MCU and programmer


foosel:
Can you change the date back to initial publication? That should only be set once, it's for ordering on the "recent" page and the RSS feed only. Thanks!

benlye:
> Can you change the date back to initial publication? That should only be set once, it's for ordering on the "recent" page and the RSS feed only. Thanks!

Oops!  My bad - fixed now.

foosel:
Perfect! :+1:

PR #166, opened 2018-02-11
Remove cyborg theme from repository

ntoff:
Since the Cyborg theme is part of the Themeify plugin now, removing the standalone version makes sense.


foosel:
Alright!

ntoff:
Thanks

PR #168, opened 2018-02-12
Updated documentation

banichow:
With the ATXPiHat board release, I needed to update the documentation and installation instructions


foosel:
Something in your PR doesn't look right, it contains the initial listing PR's commits as well as your update commit.

Please make sure to checkout the *current* `gh-pages` branch of this repository and apply *only* your update on top. Do not merge the commit you already created or you'll just recreate the current situation. Also, please leave the `date` as is - that is the date of the initial listing and should never be changed after since it's also used for sorting the "by date" list and RSS feed rendition.

PR #176, opened 2018-03-22
Add Octolapse v0.2.2 to plugin repository

FormerLurker:
Added _plugins/octolapse.md and assets/img/plugins/octolapse/tab_mini.png.  Set the date in the md file to 2018-03-24.


kantlivelong:
Hey @FormerLurker!

I remember seeing this on Reddit. Looks great!

This will be merged once 1.3.7 is released.



Will merge once 1.3.7 is released.

FormerLurker:
Are you saying my compatability settings won't work until 1.3.7 is released?  I tested it running jekyll locally, and it seemed to work with RC1 when I used the plugin manager.  Here's the relevant section:

  octoprint:
  - 1.3.7

I was hoping for get some feedback from the 1.3.7-RC1 users before 1.3.7 drops.  I know it's probably an unusual request, but my plugin requires some 1.3.7 features to work, and these were added in RC1.  If there are any problems with the new stuff hopefully I can detect it before 1.3.7 goes into production.

Thoughts?

foosel:
Huh. I would have thought you'd need to specify 1.3.7rc1 there instead of 1.3.7 for it to be detected as compatible under 1.3.7rc1. But come to think of it I probably only check base versions due to otherwise running into issues with dev versions 💡 Good to know that you tested it locally!

I have no problem with merging it earlier considering that a first RC is already out. If that wasn't the case though, yeah, it would have to wait..

Just thinking about how best to avoid confusion - should we maybe add a small heads-up to the description ("Requires OctoPrint 1.3.7 or higher")? Or maybe I should just add that to the page in a more prominent way in general?

kantlivelong:
I was just basing that off the version requirement specified in `octolapse.md`. In that case I think it would be okay but will have to defer to @foosel for final approval.

kantlivelong:
Oh well speak of the devel (pun intended).

foosel:
The devel 😈 just broke her own rule again and checked mails when she should be sleeping 😆

FormerLurker:
Regarding the notice, I'll add that in plain text for now.  Get some sleep and worry about enhancing the plugin repository later ;)  I'll change it back when it becomes moot.  I'll close this PR and create another with the updated code if that's ok.

It took me a while to get the local test going and working with my octopi instance, but eventually I got there!  I didn't notice any difference at all when I changed from 1.3.7rc1 to 1.3.7, so I though I'd just leave out the bit that didn't seem to make a difference.  It's AWESOME that I can use a local version of the plugin repository and test it against a live instance of OctoPrint!  Very slick.

FormerLurker:
> You can squash and push your changes here as well if you like.

Sorry, didn't squash.  Any way to do it after the fact?

kantlivelong:
`git rebase -i`

In your case:

1. `git rebase -i 4474bd8e2327d63faacf107e0819bf3547632846`
2. Change `pick` to `squash` or `s`.
3. Update the commit message
4. When done push with --force.

FormerLurker:
Thanks for the help, but I'm still a bit lost.  Let me study this command for a while.  Git is mystifying to me sometimes.

FormerLurker:
I seem to have destroyed my repository somehow.  It's so much easier to close this out and recreate it, so I'm going to do that.  Sorry if that causes you anymore work.

PR #186, opened 2018-04-14
Add Bed Level Visualizer plugin.

jneilliii:
Plugin that adds a tab to OctoPrint to visualize the results of gcode `G29 T1` command.


foosel:
Nice! :D

aeropic:
Hi, this is quite a nice plugin, thanks for your work.
I've just a small comment... I wonder why you need to get the "data collector flag" to be typed in by the user ? Why not leaving it empty assuming the firmware is MARLIN and then make a search in the code with a set of keywords (the ones given in the common examples) ?

It could be something like 
if the user has typed something ==> search with the users flag
else search for "common keyword 1" or "common keyword 2" or "common keyword 3"...

Besides this, it's a perfect plugin ;-)

regards
Alain

PR #194, opened 2018-05-06
Adding OctoPrint-Dropbox-Timelapse

jslay88:


foosel:
Hi @jslay88. First of all, awesome idea for a plugin! 👍 

I took a quick look at your code (which I always do for new plugin authors) and spotted one issue though that really should be fixed before this gets merged to the plugin repository. Your `SettingsPluginMixin` should really limit access to the Dropbox API key to admin users via [`get_settings_restricted_paths`](http://docs.octoprint.org/en/master/plugins/mixins.html#octoprint.plugin.SettingsPlugin.get_settings_restricted_paths), otherwise everyone with access to the instance will be able to retrieve it.

foosel:
Hard to say without seeing the code ;) I suggest opening a topic on that on the [forum](https://discourse.octoprint.org/c/development), including a link to the code you are trying.

foosel:
The latter. Though since you probably want to allow admins to access the key, it should rather be

``` python
    return dict(admin=[["api_token"],])
```

See for example [the Pushbullet plugin](https://github.com/OctoPrint/OctoPrint-Pushbullet/blob/e599bd324217c16145c3d095a0e807136be4b723/octoprint_octobullet/__init__.py#L194-L196).

jslay88:
Done. I'll need to work out the wizard later when I have more time to troubleshoot. Thanks for the help and the awesome software!

jslay88:
Merged. Thanks for catching that.

PR #201, opened 2018-05-18
gcodestatEstimator

arhi:


foosel:
Hm, there appears to be a problem with this implementation. Your hooks refer to an undeclared `__plugin_implementation__` and your handler for the software update hook also makes use of `self._plugin_version` which only gets injected into actual plugin implementations (as in subclasses of `octoprint.plugins.OctoPrintPlugin` or any of the mixins, declared as `__plugin_implementation__`).

The way this is implemented now, it doesn't look like it is functional. Could you take a look, fix this and report back once you've done this? I'll happily merge it then, I'm really happy to see that someone is already making use of the new hook(s).

arhi:
the major issue here is - I'm not python dev :D so ... this is how c++ dev fiddles with python :D 
I understood by some automagic __plugin_implementation__ is defined there :D :D :D but I'll handle that differently :D the biggest problem here is I did not test this yet as 1.3.9 is not out, what branch can I test this against? so I can test it and when it works I'll report back

foosel:
You can test this against the `maintenance` branch. `__plugin_implementation__` is not magically defined, it's a variable that you have to define and which gets then evaluated by OctoPrint plugin system. Long story short, you expected magic at the wrong place :D

arhi:
@foosel yeah I see it now (was not really thinking just copy paste :D as I said, not a python person :D ) ... will setup maintenance tonight and test this and when it's working "I'll be back" :D

arhi:
weird, something wrong here... I pulled maintenance branch

```
2018-05-23 05:57:55,456 - octoprint.plugin.core - ERROR - Error loading plugin gcodestatEstimator
  File "/home/pi/OctoPrint/venv/local/lib/python2.7/site-packages/OctoPrint-1.3.9.dev181_g8a1db8e4-py2.7.egg/octoprint/plugin/core.py", line 828, in _import_plugin
    instance = imp.load_module(key, f, filename, description)
  File "/home/pi/OctoPrint/venv/local/lib/python2.7/site-packages/octoprint_gcodestatEstimator/__init__.py", line 6, in <module>
    from octoprint.printer.estimation import PrintTimeEstimator
ImportError: cannot import name PrintTimeEstimator
```

that import that's failing is copy/paste from your example in documentation ?!

foosel:
Working as in "please take another look into merging this PR" kinda working, or more like "I can continue working on this plugin now" kinda working? Asking due to your questions on the [forum](https://discourse.octoprint.org/t/new-time-estimate-hooks-python-questions/1871) suggesting the latter :)

foosel:
Two things:

  * You don't have the update hook defined, meaning that if you push a new version, people won't get notified about it and also will have to apply updates manually. You'd need something like this to fix that:

    ``` python
    # class ...

            return dict(
                gcodestatEstimator=dict(
                    displayName=self._plugin_name,
                    displayVersion=self._plugin_version,
    
                    type="github_release",
                    user="arhi",
                    repo="OctoPrint-gcodestatEstimator",
                    current=self._plugin_version,
    
                    # update method: pip
                    pip="https://github.com/arhi/OctoPrint-gcodestatEstimator/archive/{target_version}.zip"
                )
            )
    
    # [...]

    __plugin_hooks__ = {
        "octoprint.plugin.softwareupdate.check_config": __plugin_implementation__.get_update_information,
        # ...
    }
    ```

  * You are basically instantiating your class multiple times (once as plugin, once per print as estimator) and use class fields for keeping track of values between the two instances - that works, but it is somewhat ugly. I'd actually suggest to have your factory return your instance, or maybe even better, have the instance create a dedicated estimator instance and return it, then have the plugin instance utilize that. [Something like this](https://gist.github.com/foosel/8d8753c145ba1ac00cbbb0f9b8f5d9e6).

If you want, I can merge this now, but I can also wait for you to implement the update hook if you prefer that - just tell me what to do :)

arhi:
@foosel no rush since 1.3.9 is anyhow not out yet :D .. I'll make the update hook (was planning to anyhow) it makes sense to have it. 
WRT second issue, I guessed so (I asked about it on the forum, and how to fix that), that's why I made those static variables in the class so that if there's multiple objects created they share those values :D ... I'll check out the sample you provided, thanks for that, and fix it properly (I hope) :) .. (sorry for wasting your time on these issues btw)

foosel:
You are not wasting my time. I consider this time well invested for everyone.

arhi:
I was head over heals in work these days so didn't have time to work on this, now I see you didn't just give me example but wrote the whole thing for me so I'm speechless !! thanks :D I really appreciate it! I added few more lines, tested it, both estimation functionality and automatic upgrades, everything works superbly :D so I believe it's ready for the merge now :D 

many thanks
b.

PR #202, opened 2018-05-19
Added plugin: DiscordRemote

cameroncros:


foosel:
This appears to be a fork of the already existing OctoRant plugin. Could you explain what differentiates your version from the original upstream one (different features etc)? I don't have an issue with plugins with duplicated functionality on the repo, but if they are even forked off each other it might get very confusing for people.

cameroncros:
The OctoRant plugin uses a webhook to post to Discord, and this plugin uses a full discord bot (which allows for reading messages and parsing them as commands). 
I originally created this as a PR back to OctoRant, but we came to the conclusion that I should release separately, as my changes are not backwards compatible.

The additional features are:

1. Full discord bot
1. Control octoprint via discord messages.

foosel:
Apologies, I completely forgot about this due the usual deluge of tasks.

cameroncros:
All good. Thank you for merging, and thank you for your development of Octoprint itself :)

PR #203, opened 2018-06-01
OctoPrint-SideBarTempGraph

jneilliii:
Small plugin that adds a temperature graph to the sidebar.


foosel:
Paypal's been known to come after entities using their "Donate" buttons (at least in .de jurisdiction) who aren't non-profits, simply due to the "Donate" word having special meaning. Would it be ok for you if we adjusted that link to be behind the "support me" wording? And to also specify something like "I, jneilliii, programmed this plugin [...]"? Just to avoid any kind of confusion or legal consequences for either of us.

jneilliii:
Thanks for the heads up Gina.  I've made the requested change.

foosel:
Thanks! Also just spotted an issue in the update config: key is `tempsgraph` instead of the plugin identifier `sidebartempgraph`. I think it should still work just fine, but to be entirely honest I'm not 100% sure right now 🤔 Want to fix that before I merge this just in case?

jneilliii:
Thanks, forgot to replace that one...actually ended up stripping out your flot version in lieu of using the tempsgraph plotly version anyway...all fixed up...hopefully.

foosel:
So that's odd... I just merged this via the command line. Your commits are also [visible](). But for some reason Github doesn't really have to picked up on this and marked this PR as merged. And so far it's also not rebuilding the page. I'll give it an hour or so and when then nothing has happened I'll see if I can do anything to help it happening...

foosel:
Ok, that "trigger" commit up there apparently did it. Weird.

jneilliii:
I blame Microsoft, thanks Gina.

PR #205, opened 2018-06-08
Update jneilliii plugins

jneilliii:
Updated descriptions and refreshed screenshots for all jneilliii plugins.


foosel:
Hm... your PRs always contain a merge commit from you updating your fork. Somewhere the histories have diverged. Could you force your fork to match the current master so we don't get those merge commits in the future? :)

jneilliii:
I'd be happy to if I knew how. That's not from me pulling from your fork to update mine is it?

foosel:
It is. But if things hadn't diverged between your fork and mine, that would not cause a merge commit to show up and instead do a fast forward.

How are you doing the updates? I also see references to pull requests there... Something like a local checkout with `git remote add upstream https://github.com/OctoPrint/plugins.octoprint.org?` and then a `git pull upstream gh-pages` should usually suffice to keep things up to date, going through Github's PR interface here will usually only complicate things.

jneilliii:
Yeah, just not that crazy familiar with how git works overall, should probably learn more about it.  The process I have been using is utilizing github's interface to do a pull request from your repository to my fork to get it current.  I then use tortoisegit to pull those changes from github to my windows machine to make edits.  Once complete I then do a local commit and push those changes to my fork, and the submit a pull request to your repository.

foosel:
Ah, that would explain this. In that case, I strongly suggest to either use the above commands on your local checkout on the command line, or try to figure out how to tell TortoiseGit to pull from my upstream repository. That PR step via Github is causing the issue here.

jneilliii:
Got it, thanks.  Adding your repo as a remote...pull from that remote to my local working tree, then commit and push to my repo, and then do the pull request.  Should be much cleaner next time.

PR #207, opened 2018-06-12
Changed plugin name, updated screenshots

electr0sheep:


foosel:
Can you please make your assets reside in a folder that matches the plugin's ID? I understand that your plugin now supports more bed leveling scenarios than just the CR10, but in oder to keep some resemblance of order in the assets folder I'd rather not start having multiple IDs possible per plugin there.

foosel:
Ping? :)

foosel:

PR #210, opened 2018-07-05
Added plugin excluderegion

bradcfisher:
Adds info for a new OctoPrint plugin that lets you exclude certain areas from further printing (e.g. if one part of a multi-part print fails).  The new plugin also includes a replacement gcode viewer with improved rendering performance and a couple of extra event hooks.


foosel:
Awesome! 👍

PR #214, opened 2018-07-15
Project Kronos Data Collector plugin

MrBreadWater:


foosel:
Hm... considering the potential sensitivity of the uploaded data, I think this needs a bit more of an opt-in than just a simple install of the plugin. Something like a wizard pane educating the user about what exactly gets uploaded and prompting to definitely confirm this. Otherwise I fear people will install this, forget about it and then get REALLY angry when they realize that it's happily uploading print snapshots and timelapses of failed prints. Don't get me wrong, I understand what you want to do here and it certainly sounds like a great idea, but it is somewhat delicate data and hence people need to be crystal clear on what they are allowing there *before* it actually happens in all cases.

As a sidenote, the way you are fetching the timelapses will only work in the most common setups. Instead of defaulting to `http://localhost:8080/?action=snapshot` you should instead utilize the configured snapshot URL - the user *might* have configured a different webcam server after all. And some error handling would probably also be a good idea just in case. [Take a look here](https://github.com/OctoPrint/OctoPrint-Pushbullet/blob/master/octoprint_octobullet/__init__.py#L341-L360) for an example on how to do this from a plugin.

MrBreadWater:
Is there a plugin I can look at as an example for a wizard? I'm finding the API Docs to be insufficient, and I'm having trouble finding any plugins in the repo that use a wizard.

foosel:
The [Anonymous Usage Tracking plugin](https://github.com/foosel/OctoPrint/tree/maintenance/src/octoprint/plugins/tracking) that is going to be bundled starting with OctoPrint 1.3.10 is a good candidate to copy from since it also utilizes a wizard for an opt-in mechanism.

It basically boils down to implementing the `WizardPlugin` mixin, implementing at least [`is_wizard_required`](https://github.com/foosel/OctoPrint/blob/maintenance/src/octoprint/plugins/tracking/__init__.py#L112_L113) to tell OctoPrint whether your plugin needs some input or not (e.g. you could test here whether your "allow" flag is still at a default value of `None` and if so return True, your wizard would then set either `True` or `False`) and also providing a [`wizard template`](https://github.com/foosel/OctoPrint/blob/maintenance/src/octoprint/plugins/tracking/templates/tracking_wizard.jinja2) to display to the user which you can either bind your own view model to (by setting `custom_bindings=True` in the [template config](https://github.com/foosel/OctoPrint/blob/maintenance/src/octoprint/plugins/tracking/__init__.py#L102) and then [telling OctoPrint to bind your view model to it](https://github.com/foosel/OctoPrint/blob/maintenance/src/octoprint/plugins/tracking/static/js/tracking.js#L83)) or have the wizard bind it to the `WizardViewModel` so you can access your plugin's settings via (IIRC) `settingsViewModel.settings.plugins.your_plugin.your_settings_key`. The wizard dialog will trigger a settings save at its end.

Other examples of various wizards are [the bundled CoreWizard plugin](https://github.com/foosel/OctoPrint/tree/maintenance/src/octoprint/plugins/corewizard). The [OctoPrint-TasmotaMQTT plugin by @jneilliii](https://github.com/jneilliii/OctoPrint-TasmotaMQTT) also uses it, as does the [OctoPrint Anywhere plugin by @kennethjiang](https://github.com/kennethjiang/OctoPrint-Anywhere).

MrBreadWater:
Great! Thank you so much for your help. I'll get right on adding in a Wizard :-)

MrBreadWater:
Reopened on separate pull request with updated information.

PR #223, opened 2018-08-22
Added plugin: OctoKlipper

mmone:
This plugin helps to manage the Klipper Firmware.


foosel:
That looks really great! 👍

PR #226, opened 2018-09-01
Add plugin ProgressBasedOnTime

celogeek:
Plugin ProgressBasedOnTime:

Replace file based progression by time based progression

### DIFF
--- a/_plugins/ProgressBasedOnTime.md
+++ b/_plugins/ProgressBasedOnTime.md
@@ -1,0 +1,81 @@
+---
+layout: plugin
+
+id: ProgressBasedOnTime
+title: OctoPrint-ProgressBasedOnTime
+description: Replace file based progression by time based progression
+author: Celogeek
+license: AGPLv3
+
+date: 2018-09-01
+
+homepage: https://github.com/celogeek/OctoPrint-ProgressBasedOnTime
+source: https://github.com/celogeek/OctoPrint-ProgressBasedOnTime
+archive: https://github.com/celogeek/OctoPrint-ProgressBasedOnTime/archive/master.zip
+
+# Set this to true if your plugin uses the dependency_links setup parameter to include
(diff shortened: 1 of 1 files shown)


celogeek:
They is a hook to replace the progress based on file position?

foosel:
There's a hook to replace progress estimate, but I might be misunderstanding what you are trying to do here.

Could you explain what the goal of your plugin is? What exactly is "time based progression" supposed to mean?

celogeek:
The default implementation of octoprint is to send "progress" data to plugins and apps based on the position in file.
This is really inaccurate and missleading. Because we see for example:
Estimated Total Time: 2H
Print Time: 1H
Progress: 20% (several first layer print slowly)

My plugin will change that, so here we can see "Progress: 50%" as expected.

The Print Time Genius is a really good estimator that give us a really accurate Total Print Time and Print Time Left in real time. So the "Progress" here reflect the progress of Time Left vs Total Time:
Print Time / (Print Time + Time Left).

But in order for the estimator to adjust properly the Time Left, it need the File progression (the current progress of octoprint).

That's why my patch go this way. It will call the estimator with the progression in the file, and the plugins (on_progress_data) with the progress based on the Time Left vs Total Time. It also return the progress with that information, so the "apps" and the "web interface" can change the "Progress" information properly.

celogeek:
any progress on this pr ?

foosel:
I have my plate full with a ton of stuff at the moment, hence everything is a bit slower than usual.

foosel:
Now with some fresh eyes I get what you did there... I was going in a completely wrong direction there, sorry :)

Considering that you are replacing a protected core method you really should pay close attention on future RCs of OctoPrint with this plugin published - I can't guarantee that these implementation details don't change down the road so please test against new RCs as soon as they become available and take the necessary steps to ensure compatibility.

And thanks for your patience. Life's a bit crazy currently.

PR #231, opened 2018-09-17
Add new filament sensor plugin.

Red-M:
The other two don't seem to work as the original plugin does not work and the newer one segfaults.
I've fixed this and added functionality from both into one along with my own additions.

### DIFF
--- a/_plugins/filament_sensor_ng.md
+++ b/_plugins/filament_sensor_ng.md
@@ -1,0 +1,30 @@
+---
+layout: plugin
+
+id: filamentsensorng
+title: Filament Sensor NG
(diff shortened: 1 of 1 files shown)


foosel:
You forgot to adjust the update information within your plugin code, its key is still `filamentsensor` but should be `filamentsensorng` :)

Red-M:
I believe I've update that key and I've pushed a release for that as well.
Thanks for spotting that.

Red-M:
Did I update the wrong thing?

kantlivelong:
Hi @Red-M,

We don't get notifications for external projects. I've taken a look and it appears to be good now.

Thanks for your patience.

PR #233, opened 2018-09-23
Add Project Kronos Data Collector

MrBreadWater:
Fixed Travis-CI Build error.


foosel:
Your plugin still sadly doesn't do an opt-in since your "enable" flag defaults to True even without the user's confirmation, see https://github.com/MrBreadWater/project-kronos-data-collector/blob/a3f789e7eaa026398f189d4b81bfe3ecf95a28f5/octoprint_kronos_data_collector/__init__.py#L26

I'd also be way more comfortable with this if the wizard stated the collection of images **& timelapses** in bold face and on its own paragraph, so that people really can't miss that.

The opt-in check box also should make it crystal clear that people are enabling significant data collection here, not just a plugin. 

Don't get me wrong please, I applaud the effort, really think it's a great idea and I understand that image data is crucial here to get a neural network trained for the intended goal. But I also have made the experience the past couple of years that you have to be *very clear* on something like the data collection necessary for this if you don't want people to later scream their heads off at you that they didn't know and feel spied on. So this has to be handled really really carefully and fully transparently and wording is crucial here. 

Ideally you also want to have a privacy statement that people can read that tells them a bit more about your data retention habits, what kind of data get collected by the plugin *exactly* and all that. Also keep in mind such things as the GDPR which is probably also going to affect you here. 

I faced a similar situation with the bundled tracking plugin that will be part of OctoPrint starting with 1.3.10 - opt-in, privacy statement, GDPR considerations, the whole nine yards. The [page I set up for this](http://tracking.octoprint.org/) might be interesting here.

MrBreadWater:
@foosel Alright, I've changed it to default to False. I've added extra clarity to what it's doing in the Wizard, and added a link to a Privacy Policy. Will these changes work?

PR #235, opened 2018-09-28
Plugin OctoAutoremote

trunzoc:
Added plugin OctoAutoremote


trunzoc:
Added plugin OctoAutoRemote

kantlivelong:
Hi @trunzoc,

Thanks for the PR. I have everything and just have a few questions/comments:
* Can you remove the assets/img/plugins/OctoAutoremote/test directory?

* In your setup instructions you have a link to the deprecated Wiki outlining how to install via zip. While not required I'd suggest just instructing to use the plugin manager without the zip URL. If you do plan on keeping that please use the current URL @ [http://docs.octoprint.org/en/master/bundledplugins/pluginmanager.html](http://docs.octoprint.org/en/master/bundledplugins/pluginmanager.html)

* I'm not sure if you wrote the server side portion but if so then I would recommend adjusting the code to send/accept the [autoremotekey](https://github.com/trunzoc/Octoprint_AutoRemote/blob/0b45e39ea0bdb5edde3bade892a2f0ce614f0041/octoprint_AutoRemote/__init__.py#L117-L123) as POST data instead of in the REQUEST itself. That way the key is not visible to anyone logging/sniffing traffic.

trunzoc:
Thank you for the feedback and suggestions. 

All 3 issues have been addressed and corrected

kantlivelong:
Thanks!

Merged

trunzoc:
Thank you!

PR #244, opened 2018-11-05
Added plugin TuyaSmartplug

ziirish:


foosel:
Neat, just when I look into cheap smart plugs backed by Tuya, a corresponding plugin pops up in the OctoPrint Plugin Repo 👍 

One thing I noted while giving your source a quick look - you probably should make the `localKey` an admin only attribute in the settings via [`get_settings_restricted_paths`](http://docs.octoprint.org/en/master/plugins/mixins.html#octoprint.plugin.SettingsPlugin.get_settings_restricted_paths). I can merge this like it is now, but wanted to give you a heads-up just in case you want to change this.

ziirish:
Awesome, glad someone find this useful ;)

I'll give a look at the *admin only* attribute then.

Just for your record, it was not as easy as I though to get those id/keys. I had to install a very old apk on a very old android device I got in my basement to be able to extract them. So now I can only control this smartplug through octoprint and not with my smartphone, but it answers my initial need which was to be able to turn my printer off when my print finishes.

foosel:
> Just for your record, it was not as easy as I though to get those id/keys.

Yeah, I read online that it is a bit of a PITA, which has so far kept me from looking further into this.

ziirish:
I have just implemented the `get_settings_restricted_paths` function.

PR #252, opened 2018-12-01
Marlin plugin

G-Pereira:
# Marlin Helper for OctoPrint

This tools helps flashing Marlin to the controler board and notifies the user if a new version is available.

A user friendly configuration tool can be used to set Marlin parameters and thus simplify the process of configuring Marlin. Some tools advanced tools like checking the EEPROM are planned but for now it is not the focus.
Print on!


_plugins/marlin.md
```diff
@@ -0,0 +1,32 @@
...
+layout: plugin
+
+id: marlin
+title: Marlin
+description: Marlin plugin for OctoPrint
+author: Gonçalo Camelo Neves Pereira
+license: AGPLv3
+date: 2018-12-01
+
+homepage: http://marlinfw.org/
```
foosel: Please provide the homepage & source links to *your plugin*, not to the Marlin firmware - the goal is to give people information to find more info and the source code of your plugin after all, not the firmware :)

_plugins/marlin.md
```diff
@@ -0,0 +1,32 @@
...
+source: https://github.com/marlinfirmware/marlin/
+archive: https://github.com/G-Pereira/OctoPrint-Marlin/archive/master.zip
```
foosel: This yields a 404. Please fix this link and also provide a public URL where to review the sources of your plugin.

foosel:

PR #253, opened 2018-12-01
add OctoPrint_Signal-Notifier plugin

aerickson:
Adds an entry for my new plugin for print completion notifications using Signal.

https://github.com/aerickson/OctoPrint_Signal-Notifier


_plugins/signalnotifier.md
```diff
@@ -0,0 +1,40 @@
...
+compatibility:
+  # list of compatible versions, for example 1.2.0. If left empty no specific version requirement will be assumed
+  octoprint:
+  - 1.2.0
+
+---
+
+Receive Signal messages when OctoPrint jobs are complete.
+
+Requires `signal-cli` to be installed and configured. See the plugin's github page for more information.
```
foosel: Might make sense to make "the plugin's github page" to actually link to said page, this is going on a web page after all ;)

foosel:
I'd strongly suggest to protect the sender and recipient number in your plugin's settings via [`SettingsPlugin.get_settings_restricted_path`](http://docs.octoprint.org/en/master/plugins/mixins.html#octoprint.plugin.SettingsPlugin.get_settings_restricted_paths) before publishing this. I'd consider this somewhat sensitive information. I might be mistaken of course.

Also please make sure that the identifier in your update information registration matches your plugin's identifier, right now it doesn't:

https://github.com/aerickson/OctoPrint_Signal-Notifier/blob/d7fba4f3e2b7723bbc94b25b5a74af540ed66681/octoprint_signalnotifier/__init__.py#L73

Otherwise things might not work as expected.

aerickson:
> I'd strongly suggest to protect the sender and recipient number in your plugin's settings via [`SettingsPlugin.get_settings_restricted_path`](http://docs.octoprint.org/en/master/plugins/mixins.html#octoprint.plugin.SettingsPlugin.get_settings_restricted_paths) before publishing this. I'd consider this somewhat sensitive information. I might be mistaken of course.
> 
> Also please make sure that the identifier in your update information registration matches your plugin's identifier, right now it doesn't:
> 
> https://github.com/aerickson/OctoPrint_Signal-Notifier/blob/d7fba4f3e2b7723bbc94b25b5a74af540ed66681/octoprint_signalnotifier/__init__.py#L73
> 
> Otherwise things might not work as expected.

Both done.

Thanks.

aerickson:
My force push seems to have confused the review system...

The requested changes have been made.

foosel:
Markdown and settings restriction look great now! Update config is still not correct, you changed the `displayName`, not the key in the dict, which still says `freemobilenotifier`. Also left a comment on your repo about that.

aerickson:
Thanks. I think I've got it now. :)

I noticed the plugin I copied as a starting point (https://github.com/Pinaute/OctoPrint_FreeMobile-Notifier) is using github_release, but doesn't have any releases published. I guess that would break updating?

foosel:
> Thanks. I think I've got it now. :)

👍 Looks like it :)

> I guess that would break updating?

Sadly yes.

PR #266, opened 2019-01-02
Added plugin BetterHeaterTimeout

tjjfvi:
Added plugin [BetterHeaterTimeout](https://github.com/tjjfvi/OctoPrint-BetterHeaterTimeout).


tjjfvi:
Yes, I use T6 as an alias and would prefer that to be listed, but `tjjfvi` or `T6 (tjjfvi)` or both fine as well if my Github username needs to be included.

kantlivelong:
Merged :+1:

PR #272, opened 2019-01-27
Added WebcamStreamer plugin

adilinden:
I added the WebcamStreamer plugin per instructions.


adilinden:
Hi @foosel,

Is there an example on how this is implemented? I have not looked at or touched the API. Not sure what "path" is and how to find it for the settings field that needs protecting. Is this for API only or do I need to do something for the UI settings panel as well?

I do not use  in my production OctoPi nor in my docker development environment. Will need to see how to enable multi[user to test this.

Adi

foosel:
See the examples in the documentation I linked to, or [here](https://github.com/OctoPrint/OctoPrint-Pushbullet/blob/6fe2fc9fd3c9a6ebf62e3893e262351c679df8e1/octoprint_octobullet/__init__.py#L194-L196) for a real-world example.

In your case something along the lines of this as part of your `WebcamStreamerPlugin` class would probably suffice

``` python
    return dict(admin=[["stream_url"]])
```

The thing is that OctoPrint by default exposes all your settings on the REST API. If you want to limit certain bits of your settings (here only the settings path `stream_url` - "path" because you could have nested settings, you have all of your settings on the first level though) to only logged in users or only admins, you have to tell OctoPrint about that or it won't know. The above snippet protects your `stream_url` setting and makes it admin-only since I understand that the former it is basically the secret URL used to stream to YouTube, Twitch and so on - so you really don't want that to be available to everyone connecting to OctoPrint, you only want to make it available to at the very least users or better yet admins as it's pretty much as sensitive as an API key of some other service would be. You should verify that though, I might have misunderstood something too.

adilinden:
Hi @fossel,

The suggested change has been incorporated. Yes, protecting the stream_url is a good idea.

Thanks,
Adi

foosel:
Apologies, the notification for this got buried under a flood of mails 😱

PR #281, opened 2019-02-26
Added plugin OctoPrint-PrintTrack

ElectricSquid:
Created the file and entry for PrintTrack along with images.


foosel:
Neat idea! 👍

PR #282, opened 2019-02-27
Added plugin rgb_status

EricHigdon:


EricHigdon:
Do I need to only use tags that already exist? Or can I add these new ones?

foosel:
Feel free to use existing and new ones, what ever works best for you.

Taking a quick look at your plugin I noticed that you haven't implemented the update hook. Is that intentional? If you ever want to push an update for your plugin (and rest assured, you will probably have to at some point due to implemented feature requests or reported and fixed bugs, unless you plan to do neither) that will not be possible.

You should implement such a hook. [Here's the documentation](http://docs.octoprint.org/en/master/bundledplugins/softwareupdate.html#octoprint-plugin-softwareupdate-check-config) which also includes an example. Note that you should use your plugin's identifier as the key (`updateplugindemo` in the provided example) to avoid conflicts. If you used the plugin template the hook should also have been generated by default.

Can you look into this and ping me when you've either added the hook or decided on not adding it (as mentioned, the latter is *really* not recommended but I won't force you ;)).

EricHigdon:
@foosel Oops. I implemented the update hook, but I guess I forgot to push it. Should be good now. Thanks!

foosel:
And I just saw that I still need to fix something in the plugin repository files itself, apparently the tags made the build hiccup... I'll look into that ASAP, merging this for now though!

foosel:
Heh, I missed the actual reason for the build error:

```
- ./_site/rgb_status.html
  *  linking to /by_tag/#tag-addressable-led, but tag-addressable-led does not exist (line 236)
     <a class="label" href="/by_tag/#tag-addressable-led">addressable led</a>
  *  linking to /by_tag/#tag-effects, but tag-effects does not exist (line 238)
     <a class="label" href="/by_tag/#tag-effects">effects</a>
  *  linking to /by_tag/#tag-rgb-led, but tag-rgb-led does not exist (line 242)
     <a class="label" href="/by_tag/#tag-rgb-led">rgb led</a>
  *  linking to /by_tag/#tag-ws2811, but tag-ws2811 does not exist (line 244)
     <a class="label" href="/by_tag/#tag-ws2811">ws2811</a>
  *  linking to /by_tag/#tag-ws2812, but tag-ws2812 does not exist (line 246)
     <a class="label" href="/by_tag/#tag-ws2812">ws2812</a>
```

You put the plugin file into the root directory instead of the `/_plugins/` folder. Fixed now through 50d74e517f24523e3efadcf492dd9210938ab227 and now the plugin [is actually live](https://plugins.octoprint.org/plugins/rgb_status/) as well - it wasn't so far ;)

EricHigdon:
Doh. My mistake. Thanks!

PR #283, opened 2019-03-01
Added xMatters integration plugin

svv2014:


foosel:
Hi there! Nice plugin idea! I took a quick look at your plugin and noticed a couple of issues though which you should fix before this gets published on the repository:

  * You have not implemented an [update hook](http://docs.octoprint.org/en/master/bundledplugins/softwareupdate.html#octoprint-plugin-softwareupdate-check-config) so users won't be able to update your plugin to newer versions from within OctoPrint but will have to do that manually. The plugin template should normally generate one for you as well.
  * You should [restrict access to the credentials (API key and so on) on the settings API](http://docs.octoprint.org/en/master/plugins/mixins.html#octoprint.plugin.SettingsPlugin.get_settings_restricted_paths) so that they will only be retrievable if a user is logged in as administrator.

svv2014:
Hi, thank you for pointing this out, totally makes sense! I'll fix it first.

svv2014:
@foosel I believe I fixed these issues, could you please check. thanks !

foosel:
Yep, looks good now! 👍

PR #286, opened 2019-03-07
polarcloud: remove follow-dependency-links

markwal:
The newest pip removed support for follow-dependency-links.  While this will break the plugin install on octopi 0.13, we're working on a workaround for those folks (likely includes asking them to update pip).


foosel:
Wtf, I thought I'd merged this already. Sorry!

PR #289, opened 2019-03-15
New plugin: The Spaghetti Detective (Beta)

kennethjiang:


_plugins/thespaghettidetective_beta.md
```diff
@@ -0,0 +1,170 @@
...
+
+![wizard](/assets/img/plugins/thespaghettidetective_beta/plugin_wizard.png "The Spaghetti Detective Set up Wizard")
+
+If you run into any difficulties, check out the [Setup Guide](https://thespaghettidetective.com/guide.html).
+
+## Does it cost anything?
+
+Yes and No.
+
+* First and foremost, The Spaghetti Detective is an [open source project](https://github.com/TheSpaghettiDetective). You can always grab the code and run the server yourself. Commercial use of the source code is restricted.
```
foosel: If commercial use is restricted, it isn't truly open source. The OSI definition of open source requires the *use* to be completely unrestricted, see point 6 [here](https://opensource.org/docs/osd). As far as I can see your server is licensed under the AGPL however, so I'm not sure what this "commercial use is restricted" comment is even supposed to mean because the AGPL certainly does *not* restrict commercial use.

kennethjiang: Good point @foosel . I meant to say "commercial use needs to adhere to AGPL". Will send a new PR to avoid the confusion.

PR #290, opened 2019-03-17
New plugin: Added Complicated plugin

frenchie4111:


frenchie4111:
It looks like the CI Failure is unrelated to the pull request

foosel:
> It looks like the CI Failure is unrelated to the pull request

I agree.

I've taken a quick look at your plugin (btw, great idea! 👍) and noticed two things:

  * You have not implemented an [update hook](http://docs.octoprint.org/en/master/bundledplugins/softwareupdate.html#octoprint-plugin-softwareupdate-check-config) so users won't be able to update your plugin to newer versions from within OctoPrint but will have to do that manually. The plugin template should normally generate one for you as well.
  * You should [restrict access to the credentials (API key and so on) on the settings API](http://docs.octoprint.org/en/master/plugins/mixins.html#octoprint.plugin.SettingsPlugin.get_settings_restricted_paths) so that they will only be retrievable if a user is logged in as administrator.

You might want to look into this before this gets listed on the repository.

frenchie4111:
Hey foosel. Thanks for the feedback, I will get those two changes implemented this week sometime.

foosel:
@frenchie4111 Hey, just wondering what's the status here? :)

frenchie4111:
Hey @foosel sorry, I updated the code a while ago, but never came back here to notify you. I think it should be good to go now

foosel:
Perfect, date updated and merged then.

PR #292, opened 2019-03-18
Remove comments about commercial use to avoid confusion

kennethjiang:
Also I realized that featured image is important so I'm adding that too. :)


foosel:
Thanks for the quick update!

PR #293, opened 2019-03-18
Fix the wrong archive url

kennethjiang:
@foosel Thank you for your prompt action! And my apology for sending another PR.

Just found out the archive url was actually wrong, which will cause installation to fail. Will appreciate it if you can merge this one at your earliest convenience!


foosel:
No problem ;)

kennethjiang:
Thank you so much @foosel !

PR #295, opened 2019-03-25
New plugin : Marlin Flasher

Renaud11232:


foosel:
Hi @Renaud11232!

I just took a quick look at your plugin's source (which is what I always do) and noticed that your plugin seems to allow flashing (and everything else as well) to every user, not just admins.

That seems a bit risky to me, especially considering that a ton of people sadly run their instances exposed to the internet. So this really should be changed before this plugin is published. [Take a look at how the Plugin Manager restricts access to admins-only](https://github.com/foosel/OctoPrint/blob/master/src/octoprint/plugins/pluginmanager/__init__.py#L201).

Additionally I noticed that you haven't told OctoPrint to override the allowed body size for the upload endpoint ([see the docs here](http://docs.octoprint.org/en/master/plugins/hooks.html#octoprint-server-http-bodysize) and [an example usage here](https://github.com/foosel/OctoPrint/blob/master/src/octoprint/plugins/pluginmanager/__init__.py#L976)) - if someone tries to upload a sketch larger than the default max upload size which IIRC is 2MB they'll be unhappy.

Renaud11232:
Thanks for the advice, I'll fix these points and come back when it's done

Renaud11232:
Hello again,
In addition to some improvements I had already done, I fixed the issues you mentionned.
The plugin does now have a hook to increase the request size for `/sketch_upload` to a configuratble value in MB (defaults to 20).
And all routes now have the `@restricted_access` and `@admin_permission.require(403)` 

In addition to that it now properly uses ko.js

foosel:
Great!

PR #296, opened 2019-03-25
New Plugin "DryRun"

OllisGit:

Thx, in advance
Olli

### DIFF
--- a/_plugins/curalegacy.md
+++ b/_plugins/curalegacy.md
@@ -1,0 +1,100 @@
+---
+layout: plugin
+
+id: curalegacy
+title: Cura Legacy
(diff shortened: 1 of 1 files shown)


foosel:
Hey @OllisGit!

Question, why is the user permission check [here](https://github.com/OllisGit/OctoPrint-DryRun/blob/master/octoprint_DryRun/__init__.py#L84) commented out? I think enabling or disabling dry mode is something that definitely should only be allowed to logged in users. As long as the forcelogin plugin is active it will be protected, but don't depend on that, people are... convenient.

Also: [You forgot to remove something 😜](https://github.com/OllisGit/OctoPrint-DryRun/blob/master/octoprint_DryRun/__init__.py#L4).

OllisGit:
Upps..wrong git-command. I need to open a new PullRequest.

PR #303, opened 2019-04-08
Add plugin FanSpeedMirror

b-morgan:
(second attempt)

### DIFF
--- a/_plugins/FanSpeedMirror.md
+++ b/_plugins/FanSpeedMirror.md
@@ -8,7 +8,7 @@ author: Brad Morgan
 license: AGPLv3
 
 # TODO
-date: 2019-04-08
+date: 2019-04-13
 
 homepage: https://github.com/b-morgan/OctoPrint-FanSpeedMirror
 source: https://github.com/b-morgan/OctoPrint-FanSpeedMirror
--- a/_plugins/FanSpeedMirror.md
+++ b/_plugins/FanSpeedMirror.md
@@ -21,17 +21,17 @@ tags:
 - Controls
 
 # TODO
-screenshots:
-- url: url of a screenshot, /assets/img/...
-  alt: alt-text of a screenshot
-  caption: caption of a screenshot
-- url: url of another screenshot, /assets/img/...
-  alt: alt-text of another screenshot
-  caption: caption of another screenshot
-- ...
+#screenshots:
+#- url: url of a screenshot, /assets/img/...
+#  alt: alt-text of a screenshot
+#  caption: caption of a screenshot
+#- url: url of another screenshot, /assets/img/...
+#  alt: alt-text of another screenshot
+#  caption: caption of another screenshot
+#- ...
 
 # TODO
-featuredimage: url of a featured image for your plugin, /assets/img/...
+#featuredimage: url of a featured image for your plugin, /assets/img/...
 
 ---
 
--- a/_plugins/FanSpeedMirror.md
+++ b/_plugins/FanSpeedMirror.md
@@ -1,0 +1,41 @@
+---
+layout: plugin
+
+id: FanSpeedMirror
+title: OctoPrint-FanSpeedMirror
+description: Mirror the parameter(s) of GCode commands M106, M107 to scripts on the OctoPrint host
+author: Brad Morgan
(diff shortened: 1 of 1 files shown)


b-morgan:
(third attempt)

foosel:
I took at look at your code and there's one thing that also already got mentioned on [the forums]():

I'd *strongly* suggest to not use `os.system` but rather `subprocess`, `sarge` or even OctoPrint's own [`octoprint.util.commandline.CommandlineCaller`](http://docs.octoprint.org/en/master/modules/util.html#octoprint.util.commandline.CommandlineCaller). You don't need shell capabilities for your commands, and they pose a security risk. [Python's own documentation on `os.system` also recommends against using it](https://docs.python.org/2/library/os.html#os.system).

b-morgan:
I'm researching how to replace the os.system call as recommended but in this environment, I'm not sure about the potential security risk. 

In order to install the plugin, one needs the Pi username and password which gives access to the shell. To change the plugins settings, one needs the same access, and to create the scripts that the plugin will execute, one needs the same access. Any "security risk" injected is "well known" by the user. If you already have shell access, what's the point of putting a 3D printer into the middle of any exploit you wish to execute?

foosel:
> In order to install the plugin, one needs the Pi username and password

If you are referring to SSH access, no - that's what the plugin manager in OctoPrint is for. If you are referring to OctoPrint admin credentials, agreed (those don't give immediate shell access however).

The potential issue I see here is that you are calling the script that's configured with user supplied arguments, read, the parsed fan speed from the GCODE command. You currently [restrict that to decimal values only](https://github.com/b-morgan/OctoPrint-FanSpeedMirror/blob/3abcec6a2094059ebc563c2c08987bf4cbec95ae/octoprint_FanSpeedMirror/__init__.py#L89), so there's not much risk of injecting shell code through that.

But now imagine some new firmware feature pops up which allows some form of textual parameter here and you thus adjust that regex down the road without remembering that you provide the parameter 1:1 to your script. Or something else which leads to this validation to become compromised. Now you have a security hole because someone with only *user* access (not admin) to the OctoPrint instance could craft a fan command like `M106 S;rm -rf /` which would be turned into something like `/my/M106/command.sh ;rm -rf/`. If you use something like `subprocess` without `shell=True`, nothing happens. If you use `os.process` or `subprocess` with `shell=True` huge parts of the disk are now gone. And now imagine installation of a reverse shell instead of `rm -rf` and it becomes really funny.

Yes, it is a lot of ifs and slim chances, I can't think of a way to abuse this as it is *right now* either and chances are it will never become a problem. But honestly, when it comes to something like potential abuse of functionality in a way that *can* be very very harmful, you should *always* remove as much risk of this happening as possible, especially if it's easy. You don't need shell capabilities for your callback scripts here. So you lose nothing by switching your `os.system(self.M106command + " " + str(fanPwm))` to `subprocess.call([self.M106command, str(fanPwm)])` but only gain protection against potential issues down the road.

b-morgan:
Gina,
Thank you for the explanation and the example. It wasn't just a one line change but if I did everything correctly, there should be a version 1.1.0 release that uses `subprocess.call` instead of `os.system`. I also adjusted the logging output and fixed my example screenshot and the wiki because ` >>/home/pi/error.txt 2>&1` didn't sit well with subprocess.call.

Please let me know if you don't see the new release or I did something else wrong in git. Is there anything else I need to do?

b-morgan:
You won't like 1.1.0 because it has the __init__.py from Gcode System Commands. Please wait for 1.2.0!

b-morgan:
1.2.0 has been created and awaits your blessing (or curse) :-)

foosel:
Looks good now! :+1:

> (Don't tell anyone but it uses `os.system`)

Look at that. Indeed it does. On a quick look though it doesn't use it with user/GCODE supplied input parameters (as does your plugin), so that's less of a potential attack surface at least, and it could even be argued that shell access is needed here for some specific use cases (which would then no longer work). The difference to your implementation is that the stuff that gets passed to `os.system` is fully defined by an administrator, it doesn't contain components fetched from another source. I'll still talk it over with @kantlivelong though.

PR #307, opened 2019-04-12
adding preprintservice, next try

ChristophSchranz:
solved bug


foosel:
Hi @ChristophSchranz, nice idea!

I took a quick look at your plugin and noticed a couple of issues though which you should fix before this gets published on the repository:

  * You should [restrict access to the API key on the settings API](http://docs.octoprint.org/en/master/plugins/mixins.html#octoprint.plugin.SettingsPlugin.get_settings_restricted_paths) so that it will only be retrievable if a user is logged in as administrator. Additionally you should not put it into your custom template vars. Currently you aren't using the specific template var anywhere, but if you ever make the relevant code line active again you've now leaked the API key in the HTML source.
  * I'm not entirely sure about the full workflow, but it looks like your plugin will require a secondary OctoPrint instance which it uses for slicing? Or should that be the local one? If so, why the detour through HTTP for storing files instead of using filemanager directly? I'd be happy to get some clarification here :D In any case, the URL you use for calling that requires it to be available on port 5000 and at the root of the server. That doesn't necessarily have to be the case, so you should probably allow specifying the full base URL instead of just the host name. But I might also be wrong in how this all comes together.
  * There are a lot of commented out bits and pieces that seem to be left-overs from other plugins.

ChristophSchranz:
Thanks @foosel for your clear feedback!

I'll make the changes and add a workflow diagram (the separation is optional and needed if the octoprint server runs on a slow low-cost machine, e.g. a Raspberry Pi).

Regarding the port: Is there a method that retrieves the port on which octoprint is reachable on? 

Best regards,
Chris

ChristophSchranz:
Hi Gina,
I've made all changes you requested, inclussive a clear description of the suggested workflow.
Best regards from Salzburg.

foosel:
I don't think your code still works now. You've changed to `restricted.admin_only.apikey` yet your code still tries to read `apikey` from the settings at multiple places.

Easy solution:

``` python
	def get_settings_defaults(self):
		return dict(url="http://192.168.48.81:2304/tweak",
					octoprinturl="http://192.168.48.43:5000/",
					get_tweaked_stl=True,
					tweak_action="tweak slice",
					isurlok=False,
					default_profile=os.path.join(os.path.dirname(os.path.realpath(__file__)), "profiles",
												 "default_slic3r_profile.ini"),
					isapikeyok=False,
					apikey="asdf")

		return dict(admin=[["apikey"]])
```

ChristophSchranz:
Danke für deine Antwort!

Der Code funktioniert, sobald ich als Admin eingeloggt bin. Sollte das 
nicht das gewünschte Verhalten sein?
Wenn nicht, kannst du mir Beispielcode zeigen, in dem die 
Programmierkonventionen eingehalten werden? Ich stehe bei der weiteren 
Umsetzung leider an..

Liebe Grüße,
Christoph



Am 03.05.2019 17:32, schrieb Gina Häußge:

foosel:
Das Problem ist, dass du den Pfad in deinen Settings zum API-Key geändert hast, diese Pfadanpassung aber nicht im Rest vom Code - insbesondere in `on_settings_save` - nicht beachtet hast. Sprich, du versuchst dort immer noch `apikey` aus den Settings zu lesen, hast das Teil aber nach `restricted.admin_only.apikey` verschoben, was gar nicht nötig war, denn es reicht, einfach den Pfad `apikey` in `get_settings_restricted_paths` auf `admin` zu limitieren, wie in meinem Beispiel oben gezeigt.

Dieses ganze `restricted.admin_only` Zeugs war nur im Doku-Beispiel, um eine Datenstruktur zu haben, aus der gleich hervorgeht, was wie eingeschränkt ist. Eine solche Hierarchie in deinen Settings anzulegen ist aber nicht nötig, alles, worauf es ankommt, ist, in `get_settings_restricted_paths` die zu schützenden Pfade zurück zu geben, in deinem Fall also einfach nur `apikey`.

---

The problem is that you changed the settings path to your api key but didn't reflect that change in the rest of the code, especially in `on_settings_save`. You still are trying to read `apikey` from the settings there but actually moved that value into `restricted.admin_only.apikey`, which wasn't necessary at all since it would have sufficed to simply define the path `apikey` as restricted to `admin` in `get_settings_restricted_paths`, as shown in my example above.

All this `restricted.admin_only` stuff is only used in the documentation example to have a data structure that immediately shows you what will be restricted how. To create such a hierarchy in your settings isn't necessary though, all that matters is that you define the paths that need to be protected in `get_settings_restricted_paths`, so only `apikey` in your case.

ChristophSchranz:
Thanks Gina for your Solution.
I've committed those and hope it will be published very soon! :)

foosel:
Looking good now! 👍

PR #313, opened 2019-05-18
Updated description

ChristophSchranz:
The functionality is the same.


foosel:
I've rolled back the change to the publication date. Please do not change that on updates - it's the *publication* date after all ;)

ChristophSchranz:
Thanks @foosel 
Just one more question: How is it possible to update the description in https://plugins.octoprint.org/plugins/preprintservice/ ?

I've updated the README.md /extras/preprintserver.md in the Plugin-repo, and the preprintserver.md this repo. However, the description published on plugins.octoprint.org is from the previous release.

foosel:
Hm... could be that the build didn't properly run through, that wouldn't be the first time. I'll try to tickle it into action

foosel:
There we go. Should be fine now.

PR #317, opened 2019-05-25
Adds plugin Upload Anything

rlogiacco:
Plugin identifier `uploadanything`


rlogiacco:
thanks @foosel for your help: I didn't catch it

foosel:
You are welcome, and a nice idea for a plugin 👍 A suggestion though: you are registering the uploaded files as `machinecode`. That means they can be selected for printing. I'd rather recommend to register them as `model`. In that case you can still download them - and if some kind of compatible slicer plugin is installed you could even convert them to `machinecode`. But you can't accidentally print them.

Do you want to change that? If so I'll wait with the merge until you ping me here.

rlogiacco:
Absolutely! I applied the suggested fix and it should now be ready to go. I was totally unaware of the meaning of `machinecode`: I just copy-pasted from other plugins :smile:

rlogiacco:
BTW @foosel I'm releasing version `1.0.0`, I don't know if you have any specific policy regarding plugin versions, I see most of them are in the `0.x` style...

foosel:
> BTW @foosel I'm releasing version `1.0.0`, I don't know if you have any specific policy regarding plugin versions, I see most of them are in the `0.x` style...

No policies at all, all good :)

PR #321, opened 2019-05-27
Added OctoPod plugin

gdombiak:


foosel:
👍

gdombiak:
Thanks Gina!

gdombiak:
Hi Gina,

Not sure why build failed. Let me know if I should look into something to make the build pass.

Thanks,
Gaston

foosel:
@gdombiak seems to have been some weird race condition, ignore it. Unrelated to the page not having updated - that's not triggered by Travis at all but sometimes Github Pages just "forgets" to build stuff on a page push, see also #323

gdombiak:
Gotcha. Thanks @foosel for your help. I have another quick question. In OctoPrint I went to _Settings -> Plugin Manager -> Get More_ and did a _Search_ and the plugin does not appear by its name or text listed on the plugin's page. Is there a reindex that needs to happen? Will run at some moment?

Thanks again,
Gaston

foosel:
This was just reported in #324 as well. Not sure yet what's up, I see the plugin data show up in the generated repository json file.

foosel:
Hm, nope, can't reproduce here, after a refresh of the repository data that is: 

image

PR #325, opened 2019-05-31
Added plugin filamentencore

draagc:


foosel:
Hi there, nice plugin 👍 I took a quick look at the source and noticed that the update hook references the wrong key `octoprint_filament` (instead of `filamentencore`). You should fix this before this is merged, otherwise your plugin's updates might not work.

draagc:
Hi there @foosel ! I have now updated the key to the correct one.

foosel:
And merged, thanks!

PR #327, opened 2019-06-03
Add new plugin "PauseForUserEvent"

zeroflow:
This new plugin adds a new event "paused_for_user" in case 'echo:busy: paused for user' is received on serial
This can be used to forward this event via MQTT to a mobile phone so the user knows that the printer needs manual intervention.

The usecase is the Prusa MMU2.
In case the loading/unloading procedure goes wrong, it then just keeps sending `echo:busy: paused for user` until the user fixes the issue and presses the button on the LCD.

Currently, the user receives no notification here as OctoPrint nicely waits for the printer to signal it's readiness again.

### DIFF
--- a/_plugins/PauseForUserEvent.md
+++ b/_plugins/PauseForUserEvent.md
@@ -16,7 +16,7 @@ tags:
 compatibility:
   octoprint:
   - 1.3.11 
-date: 2019-06-03
+date: 2019-06-11
 
 ---
 
--- a/_plugins/PauseForUserEvent.md
+++ b/_plugins/PauseForUserEvent.md
@@ -2,7 +2,7 @@
 layout: plugin
 id: PauseForUserEvent
 title: PauseForUserEvent
-description: Adds a new event 'paused_for_user' when the printer needs manual intervention (signalled by sending "echo:busy: paused for user")
+description: Adds a new event 'paused_for_user' when the printer needs manual intervention
 author: Thomas Arthofer
 license: AGPLv3
 homepage: https://github.com/zeroflow/OctoPrint-PauseForUserEvent
--- a/_plugins/PauseForUserEvent.md
+++ b/_plugins/PauseForUserEvent.md
@@ -1,7 +1,7 @@
 ---
 layout: plugin
-id: PauseForUserEvent
-title: PauseForUserEvent
+id: pause_for_user_event
+title: PauseForUser Event
 description: Adds a new event 'paused_for_user' when the printer needs manual intervention
 author: Thomas Arthofer
 license: AGPLv3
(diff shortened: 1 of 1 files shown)


foosel:
Hey there, great idea for a plugin!

A couple of points though about the implementation:

  * Your identifier should be all lower case. OctoPrint handles it as case insensitive in most cases anyhow, but I just spotted a location where it doesn't (which I'll fix in 1.3.12 though) and you'll make your life easier for yourself if you make it `pauseforuserevent` or `pause_for_user_event` instead of `PauseForUserEvent` ;)
  * You make use of the new registration feature for custom events that's only available in OctoPrint 1.3.11 and up, so the compatibility should probably be adjusted correspondingly. 
  * You should also actually use the custom event registration and adapt the code accordingly instead of firing off an arbitrary value over the event bus which may or may not clash some day down the road. If you use the event registration the event will be called `plugin_PauseForUserEvent_paused_for_user` instead of just `paused_for_user` (see the [docs](http://docs.octoprint.org/en/master/plugins/hooks.html#octoprint-events-register-custom-events), and instead of hardcoding its name in the `fire` call you should use the `Events.PLUGIN__PAUSE_FOR_USER_EVENT_PAUSED_FOR_USER` constant that gets generated for you by the event registration:

    ```
    2019-06-04 14:25:21,956 - octoprint.server - DEBUG - Registered event paused_for_user of plugin PauseForUserEvent as Events.PLUGIN__PAUSE_FOR_USER_EVENT_PAUSED_FOR_USER = "plugin_PauseForUserEvent_paused_for_user"
    ```

    That event registration is actually the location where I noticed it's not case insensitive and if you turn the identifier into `pauseforuserevent` as suggested above it'll be `Events.PLUGIN_PAUSEFORUSEREVENT_PAUSED_FOR_USER` and `plugin_pauseforuserevent_paused_for_user` instead, or `Events.PLUGIN_PAUSE_FOR_USER_EVENT_PAUSED_FOR_USER` and `plugin_pause_for_user_event_paused_for_user` for the identifier `pause_for_user_event`.

  * I also see that you have the AssetPlugin and TemplatePlugin mixins included even though you don't offer any form of UI - feel free to remove those, no need to register yourself for anything you don't actually do :smile:

zeroflow:
Thank you for your feedback on what is to do.

* Identifier was renamed
* OctoPrint required version updated for the register event feature
* Using defined values for events instead of free-text
* Asset and Template plugin removed

Btw: Your fast answer on patreon on how this can be done did help a lot!

foosel:
> * Identifier was renamed

You [missed one](https://github.com/zeroflow/OctoPrint-PauseForUserEvent/blob/6dea103cd139334275e43f80edb83e370fca2b7d/octoprint_pause_for_user_event/__init__.py#L18) 😅

> Btw: Your fast answer on patreon on how this can be done did help a lot!

Glad to hear! 😊

foosel:
And published! 👍

PR #334, opened 2019-07-06
unregister all plugins by ntoff

ntoff:
removing due to lack of time to ensure future compatibility / bug fixes

Code will likely remain on github, but likely won't be updated.


foosel:
Roger!

PR #339, opened 2019-08-18
Add Ikea Tradfri plugin

ralmn:


foosel:
I took a look at your plugin's source and spotted a few issues that need to be fixed before this can be registered:

  * [This looks like some left over test code](https://github.com/ralmn/OctoPrint-Ikea-tradfri/blob/master/octoprint_ikea_tradfri/__init__.py#L40) with your own security code - you probably should remove that (and reset your code).
  * [Your error handling here](https://github.com/ralmn/OctoPrint-Ikea-tradfri/blob/master/octoprint_ikea_tradfri/__init__.py#L50) will nuke the whole OctoPrint server. Your plugin should *never* just stop the whole process by invoking `sys.exit`. Handle errors gracefully and early. If push comes to shove *return* or raise exceptions, but never just nuke the server.
  * [Use `self._logger` for logging errors please](https://github.com/ralmn/OctoPrint-Ikea-tradfri/blob/master/octoprint_ikea_tradfri/__init__.py#L49)
  * I'd strongly suggest to make use of the higher level `sarge` dependency instead of tinkering around with `popen`, especially if you lack proper error handling like [here](https://github.com/ralmn/OctoPrint-Ikea-tradfri/blob/master/octoprint_ikea_tradfri/__init__.py#L47).
  * You should [restrict access to the credentials (security code) on the settings API](http://docs.octoprint.org/en/master/plugins/mixins.html#octoprint.plugin.SettingsPlugin.get_settings_restricted_paths) so that they will only be retrievable if a user is logged in as administrator.
  * Installation paths can be different on other people's machines. You really should think about making the coap path configurable.

ralmn:
Thanks for your review :) I have starting to fix and I will continue this week/month

foosel:
Any news here? :)

ralmn:
Sorry, I haven't had time to work on it again in the last few months.

foosel:
Ok... I'm closing this then for now, we can revisit this again later :)

PR #340, opened 2019-08-18
Add Octovox Plugin to Repository

johnnyruz:
Request addition of my plugin, OctoVox, to the OctoPrint repository.


foosel:
Hey @johnnyruz, I took a quick look at your plugin's source and spotted some points:

  * You should [restrict access to the credentials (e.g. printer_uid) on the settings API](http://docs.octoprint.org/en/master/plugins/mixins.html#octoprint.plugin.SettingsPlugin.get_settings_restricted_paths) so that they will only be retrievable if a user is logged in as administrator.
  * You [initialized your own logger here](https://github.com/johnnyruz/OctoPrint-OctoVox/blob/master/octoprint_octovox/__init__.py#L27). That's not needed, it will be injected by the system, and having a rogue logger might cause confusion down the road. If you need injected properties for further initialization of components, use the [`initialize` method](http://docs.octoprint.org/en/master/modules/plugin.html#octoprint.plugin.core.Plugin.initialize) which will be called by OctoPrint on your plugin implementation after all injections have been performed.
  * It would be awesome if your plugin's settings page also included a link to the privacy information about your plugin as well. 

Could you fix them and ping me here once that's done? Thanks!

johnnyruz:
@foosel thank you so much for the valuable feedback!!

I believe I have implemented your suggestions and have checked them into my OctoPrint-OctoVox master branch and created a new release. Please take a look whenever you get a chance and let me know if there are any more steps I need to take.

Thanks again!

foosel:
I took a look and spotted one thing that is probably still an issue (unless I'm missing something which might very well be the case).

Your `UpdateStatus` class (which is *not* managed by OctoPrint directly and hence doesn't receive property injections, only your registered plugin implementation `OctovoxPlugin` does) tries to use `self._logger` which isn't defined though, [see e.g. here](https://github.com/johnnyruz/OctoPrint-OctoVox/blob/master/octoprint_octovox/update_status.py#L36).

I'd suggest to inject `self._logger` from your implementation in `initialize` as originally outlined above. So you'll have something like this:

``` python
class OctovoxPlugin(octoprint.plugin.StartupPlugin,
		     octoprint.plugin.SettingsPlugin,
                     octoprint.plugin.AssetPlugin,
                     octoprint.plugin.TemplatePlugin,
		     octoprint.plugin.EventHandlerPlugin):

	def __init__(self):
		super(OctovoxPlugin, self).__init__()
		self._updateStatusTimer = None
		self._update_status = None

	def initialize(self):
		self._update_status = UpdateStatus(self._logger)

	# ...
```

johnnyruz:
@foosel thanks for your patience and you're absolutely right. The "happy path" worked completely fine, but as soon as something happened that triggered a call to the logger in UpdateStatus it generated an exception. Your suggestions have resolved that issue.

One other thing I noticed in actually performing good testing is that by tagging several of the plugin settings as "never" in the restricted paths, they were successfully omitted from the API calls but also became NULL in the Knockout data bindings. I've tagged all of my settings as "admin" in the restricted paths method which has resolved the issue, but I'm interested if there's a better method to store static values that would be exposed to both the python code as well as the Knockout bindings. Although that's likely better discussed in the forum vs. a git pull request comment!

johnnyruz:
Hi @foosel,

I believe I have resolved your code suggestions and everything is pushed to my master branch with release 0.1.5 of the Plugin.

Thanks!

foosel:
Looking good now!

PR #341, opened 2019-08-21
Added plugin right_sidebar

amsbr:


foosel:
Hm... you seem to have mixed two plugin registrations here 🤔 Was that intentional?

amsbr:
> Hm... you seem to have mixed two plugin registrations here 🤔 Was that intentional?

Sorry, it's was not intentional. It's distinct plugins but committed in approximate days :)

foosel:
Do you want both reviewed and published? Don't want to accidentally push something here that's not yet intended for release by you ;)

amsbr:
I reviewed, both published. Thank you ;)

foosel:
Couple of issues:

  * `right_sidebar`:
    * [Broken update configuration](https://github.com/amsbr/OctoPrint-Right-Sidebar/blob/350ae3ab2e3fa924f8095f7edc3a5be3f5a5b383/octoprint_right_sidebar/__init__.py#L43), should be `right_sidebar`, not `systemcommandeditor`.
  * `compressed_upload`:
    * [Broken update configuration](https://github.com/amsbr/OctoPrint-Compressed-Upload/blob/master/octoprint_compressed_upload/__init__.py#L146), should be `compressed_upload`, not `stats`
    * [Leaking file handles and no error handling on unpacking](https://github.com/amsbr/OctoPrint-Compressed-Upload/blob/master/octoprint_compressed_upload/__init__.py#L92) - this really needs a `try...except` wrapper plus a `with open(...)`. Same for the other archive formats.
    * You just happily unpack whatever is in the archive in the upload folder. You really should at the very least limit this to only accepted file types, trigger post processing for all of them and so on. The clean approach would be to unpack to a temporary directory and then add the unpacked files through the file manager. The way you are doing it now can cause hiccups and can also be abused to overwrite metadata.
    * `self._file_manager.path_on_disk(destination, path).replace(path, "")` - this will not work as expected if the file delimiters in `path` and `path_on_disk` don't match. Also - what is the goal of this in the first place? You always want to unpack at the destination root? Why? That feels very unintuitive (also see my previous remark about sanity checking your archives).

Could you please address them and then ping me here again? Cheers!

amsbr:
Thanks for the feedback! I go to fix these.

foosel:
Any news here? :)

jneilliii:
@amsbr were you able to make the requested corrections? Since this PR is so old if we don't hear back from you in the next 14 days this PR will be closed.

foosel:

PR #343, opened 2019-08-25
Added plugin PrusaETAOverride

kanocz:
Lastest firmwares of Prusa printers returns someting like "NORMAL MODE: Percent done: 0; print time remaining in mins: 3" in response on M73 even with priting from SD. This plugin sets ETA to received value


foosel:
👍

PR #348, opened 2019-09-09
A Dashboard tab for Octoprint

StefanCohen:
A dashboard tab for Octoprint that displays the most relevant info regarding the state of the printer and any on-going print jobs.


foosel:
So apparently either the date or the screenshots was the issue O_O

PR #353, opened 2019-09-30
New Plugin "AutoLogout"

OllisGit:

Thx, in advance
Olli

### DIFF
--- a/_plugins/AutoLogout.md
+++ b/_plugins/AutoLogout.md
@@ -1,0 +1,45 @@
+---
+layout: plugin
+
+id: AutoLogout
+title: AutoLogout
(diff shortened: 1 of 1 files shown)


foosel:
Looking good!

A piece of advice though: From an HTTP API perspective it's bad practice to have `GET` requests modify internal state. The SimpeAPI provides a command pattern implementation on the same endpoint that can be used by POST requests for that very reason. You might want to utilize that, it would highly reduce the likelihood of weird issues due to caching or crawlers or whatnot ;)

OllisGit:
Hi @foosel !

Thanks, for the advice: I totally agree, but unfortunately the writing GET-Calls are used by my common component "Reset to Default-Settings".
So, if I change this behaviour, I need to change it in every of my Plugins...maybe a task for X-Mas.

C ya,
Olli

PR #368, opened 2019-11-16
Ultimaker Format Package

jneilliii:
Plugin that adds support for Ultimaker Format Package (.ufp) files and stores thumbnail and extracts gcode.


foosel:
Oh that is a fun one!

PR #370, opened 2019-11-20
Add Filament Revolutions Plugin

RomRider:


foosel:
Hey there, nice idea for a plugin! I just quickly took a look at the source and noticed that the softwareupdate hook uses the wrong identifier here: 

https://github.com/RomRider/Octoprint-Filament-Revolutions/blob/master/octoprint_filamentrevolutions/__init__.py#L263

That should say `filamentrevolutions` instead of `octoprint_filament`.

You might also want to look into a `__plugin_check__` implementation to make sure `RPi.GPIO` is even available (someone might try to install your plugin under an OctoPrint instance not running on a Raspberry Pi - yes, those DO exist ;)).

RomRider:
I think I've fixed the issue you mentioned @foosel and added a `__plugin_check__`

foosel:
Looks good now!

PR #371, opened 2019-11-27
Added Plugin mystromswitch

da4id:


foosel:
Hey there! Nice plugin!

I just spotted a typo in your settings that you might want to fix:

https://github.com/da4id/OctoPrint-MyStromSwitch/blob/master/octoprint_mystromswitch/__init__.py#L199

Should be `powerOnOnStart`, is missing the `p` at the start.

You might also want to secure access to the `ip` settings to admins only, see [the docs](http://docs.octoprint.org/en/master/plugins/mixins.html#octoprint.plugin.SettingsPlugin.get_settings_restricted_paths) on how to achieve that.

da4id:
Hi Gina

Thanks for your Review! 
I fixed both points and released 1.0.1 Version

I couldnt test the restriction of the ip settings. I created an admin user and a test user. The testuser has only the user role. When i log in with this user the Settings button on the top is missing to open the settings.

da4id:
Sorry wrong button

foosel:
Looking good now! 

> When i log in with this user the Settings button on the top is missing to open the settings.

Yeah, the settings however are still getting fetched from the backend. A look into the network tab on the browser dev console would have revealed the IP then.

PR #372, opened 2019-11-30
Adds Resource Monitor plugin

Renaud11232:
This adds a plugin that adds a tab to show various information about resource usage


Renaud11232:
There's an issue with the build but i'm unsure what

foosel:
Nice plugin!

PR #377, opened 2019-12-08
adds prometheus exporter

tg44:


foosel:
Very nice, thank you!

PR #379, opened 2019-12-12
python3 Compability for Plugin GCodeBar

mtowara:
Updated .md file here and setup.py and __init__.py in plugin repo (and tested obviously, what could go wrong ;) ).


foosel:
Merged, but please don't modify the pub date in the future ;)

PR #382, opened 2019-12-13
Added plugin PrusaLevelingGuide

scottrini:


foosel:
Hey, I took a quick look and found some points...

  * it looks your update config is faulty. It's using the identifier "levelingguide" instead of your plugin's identifier "PrusaLevelingGuide":
    
    https://github.com/scottrini/OctoPrint-PrusaLevelingGuide/blob/master/octoprint_PrusaLevelingGuide/__init__.py#L81

  * You also might want to remove the unused plugin command definition which seems to do nothing:

    https://github.com/scottrini/OctoPrint-PrusaLevelingGuide/blob/master/octoprint_PrusaLevelingGuide/__init__.py#L30-L44

  * Your octoprint.comm.protocol.gcode.received hook handler is not optimized, it recompiles a regular expression for every single received time, which will slow down communication way more than it should. Also it will *always*  try to match every single line, even if no leveling is active. That's suboptimal as well.

    https://github.com/scottrini/OctoPrint-PrusaLevelingGuide/blob/master/octoprint_PrusaLevelingGuide/__init__.py#L126-L133

    You should fix this as well, it will cause performance issues when your plugin is installed.

  * Finally please be aware that newly registered plugins from now one should also be tested and made sure to be compatible to Python 3, since Python 2 is EOL. You can find more on that [here](https://plugins.octoprint.org/help/registering/) and [here](https://community.octoprint.org/t/towards-python-3-and-octoprint-1-4-0/12382).

Please take care of these points and ping me here once done.

scottrini:
Thanks for taking a look at this!

You've made some great points and I've updated my plugin.  Fixed the first two callouts.  You are absolutely right about the recompiling of the regex.  That doesn't make sense.  I actually copied the code from another plugin and just didn't think about it again.  So now I'm compiling it on plugin start and re-using it.  Additionally, I tried to make it a little better so it's not checking every line, even though I want it to be able to catch **any** G81 response.  In the future, I want to add a "target bed variance" setting, with the intention that G81 is configured in the end print gcode.  The plugin can compare the new bed variance and display something in the navbar if the user should go through the leveling again.

So I've modified it to listen on gcode sent and set a flag when G81 is actually sent and only check the received gcode for a period of time after G81 is detected.

Additionally, I've started testing the plugin in a VM in python3 and so far have had no issues, but I'd like to test it some more before marking it 100% compatible.

Thanks again for your feedback and everything you've done for the community!

foosel:
Looking good now, thank you!

PR #384, opened 2019-12-19
Added date information to plugin details-page

OllisGit:
Hi,

I missed the date information of each plugin in the detail page. 
Now it looks like this: 
image


foosel:
Merged after moving the date down a bit.

PR #389, opened 2020-01-16
Added plugin progress_title

pwnbus:


foosel:
Hi there, nice plugin! There's one open point however:

  * Your plugin is not Python 3 compatible. Python 2 is EOL. OctoPrint 1.4.0 does support Python 2 and Python 3, plugins will be expected to do so as well (or at least support Python 3) during a transition period of roughly a year after the release of OctoPrint 1.4.0 to allow a migration of the whole ecosystem. Newly registered plugins should be Python 3 compatible out of the box. [See this forum thread for more information](https://community.octoprint.org/t/towards-python-3-and-octoprint-1-4-0/12382?u=foosel).

Can you take a look at that and ping me once that's taken care of? Thank you!

pwnbus:
@foosel Thank you for your kind words! I've updated my plugin to support python3, and tested to verify everything should be all set.

Please let me know if there's anything else that's required!

foosel:
All good now, thanks! 👍

xia0:
nice plugin!

PR #390, opened 2020-01-21
Added Plugin macros

manojdevios:
Added plugin macros


foosel:
Hi, could you please explain what distinguishes your plugin from the [built-in custom controls feature](http://docs.octoprint.org/en/master/features/custom_controls.html)? Right now it seems like it pretty much reimplements a slightly reduced version of that, though with a settings dialog.

manojdevios:
Hi @foosel,
Thanks for taking the time and reviewing my submission.

When i started making this plugin for myself, i wanted to have a simple button where i could copy/edit gcode commands easily. So i used settings page to store the name and gcode.

But like you said is very similar to the custom controls with the advantage of changing the buttons gcode and name from settings page

foosel:
Frankly I'd rather not have a plugin that pretty much copies built-in functionality listed in the repository. I fear it will only confuse users, and end up causing support overhead on both ends from users who are having questions/feature requests/bug reports for your plugin directed at me and vice versa.

manojdevios:
Hi @foosel, I think I completely agree with you. Anyhow thanks for your time on this. 

Thanks again for your awesome work.

PR #391, opened 2020-01-25
Added plugin octohue

entrippy:


foosel:
Hey there, thanks for the plugin registration request! I took a quick look over your plugin and spotted some things:

  * You should [restrict access to the credentials (bridge URL, username, ...) on the settings API](http://docs.octoprint.org/en/master/plugins/mixins.html#octoprint.plugin.SettingsPlugin.get_settings_restricted_paths) so that they will only be retrievable if a user is logged in as administrator.
  * Your plugin is not Python 3 compatible. Python 2 is EOL. OctoPrint 1.4.0 does support Python 2 and Python 3, plugins will be expected to do so as well (or at least support Python 3) during a transition period of roughly a year after the release of OctoPrint 1.4.0 to allow a migration of the whole ecosystem. Newly registered plugins should be Python 3 compatible out of the box. [See this forum thread for more information](https://community.octoprint.org/t/towards-python-3-and-octoprint-1-4-0/12382?u=foosel).

Can you take a look at these and ping me once they are handled? Thanks!

entrippy:
Hi @foosel , no problem Re: restricting settings to the admin user, I'll sort that out ASAP.
Regarding Python3, my mistake. While I haven't been able to test against a Python 3 octoprint, the plugin was linted against 3.7 with no errors. I'll get the compatibility flag corrected.

entrippy:
Hi @foosel , I believe both of these should now be fixed, although I'm not sure how to test the restricted paths.

foosel:
Looking good, but your plugin's `__init__.py` is still missing the `__plugin_pythoncompat__` property to signal Py3 compatibility. There are already example lines in there, you just need to uncomment the right one

E.g.

``` python
__plugin_name__ = "Octohue"

# Starting with OctoPrint 1.4.0 OctoPrint will also support to run under Python 3 in addition to the deprecated
# Python 2. New plugins should make sure to run under both versions for now. Uncomment one of the following
# compatibility flags according to what Python versions your plugin supports!
#__plugin_pythoncompat__ = ">=2.7,<3" # only python 2
#__plugin_pythoncompat__ = ">=3,<4" # only python 3
__plugin_pythoncompat__ = ">=2.7,<4" # python 2 and 3
```

entrippy:
@foosel sorry, face palm moment, I thought I'd only messed up the one in the plugins description, and completely over looked where it is set in the plugin itself. this has now been fixed.

foosel:
Looking good now, merged! 👍

PR #423, opened 2020-04-01
OctoPrint-AutoTerminalInput

jneilliii:
Automatically activate the command input field when switching to the Terminal tab of OctoPrint.


jneilliii:
Sorry about the multiple commits...

PR #433, opened 2020-04-14
Added Filament Sensor Simplified plugin

LuckyX182:


jneilliii:
I've begun the review of your submission. Everything looks good except for a couple of changes will need to be completed prior to merging.

Please change identifier in software update hook return dictionary to match the plugin's identifier [here](https://github.com/LuckyX182/Filament_sensor_simplified/blob/2def47c5dffe6f1a859d0f83b335337805d4b013/octoprint_filamentsensorsimplified/__init__.py#L145). Should be `filamentsensorsimplified`. You may also consider updating the `displayName` just below to match the plugin's name.

It appears this utilizes RPi.GPIO module, which is raspberry pi specific, please update compatibility flags for linux os only.

Would the raised exception on initialize cause OctoPrint to crash in any way or will it still load and operate normally? This is just my inexperience with the OctoPrint lifecycle and @kantlivelong might be able to answer this one for us.

You sure you don't want to include a screenshot of the settings or anything along with the plugin's acceptance into the repository?

kantlivelong:
The RPi.GPIO exception will ultimately cause the plugin to not load which is fine if that is what is intended. Though it would make sense to limit to Linux only if RPi.GPIO is a hard dependency.

foosel:
> The RPi.GPIO exception will ultimately cause the plugin to not load which is fine if that is what is intended.

I would actually use [`__plugin_check__`](https://docs.octoprint.org/en/master/plugins/controlproperties.html) for that instead of failing as late as plugin initialization, that's what that control property is meant for.

kantlivelong:
> > The RPi.GPIO exception will ultimately cause the plugin to not load which is fine if that is what is intended.
> 
> I would actually use [`__plugin_check__`](https://docs.octoprint.org/en/master/plugins/controlproperties.html) for that instead of failing as late as plugin initialization, that's what that control property is meant for.

That's definitely cleaner :+1:

PR #437, opened 2020-04-17
Added Plugin BetterKlipper

lbibass:
This is a revised version of octoklipper that works better with plugins that change themes. The old one was abandoned, unfortunately.


lbibass:
Thank you! I will make those changes tomorrow.

lbibass:
The changes have been made!

jneilliii:
@lbibass it appears you are still referencing the same image as before, it was not updated to your own sub-folder as requested. Not sure if this is just a PR issue or your changes weren't made to the same branch that you were working on before. Also, I just noticed you are missing Python 3 compatibility flag. For new plugins we are requesting that they be both Python 2.7 and 3 compatible. Additional information for this can be found [here](https://docs.octoprint.org/en/master/plugins/python3_migration.html).

jneilliii:
It also appears that none of the requested changes were done. The plugin's identifier in setup.py is still `klipper`. As mentioned this needs to be unique and not the same as the original.

lbibass:
Ack! Sorry about that, I forgot to push my changes.

jneilliii:
Some changes in your plugin directly are still necessary.

- [here](https://github.com/lbibass/OctoprintKlipperPlugin/blob/4689129ae32fa828f6fbedb7b51701176c3f544a/setup.py#L3) needs to be a unique identifier and not re-use `klipper`.

- [here](https://github.com/lbibass/OctoprintKlipperPlugin/blob/4689129ae32fa828f6fbedb7b51701176c3f544a/octoprint_klipper/__init__.py#L310) needs to also be changed to match to whatever you use for above and not re-use `klipper`.
- [here](https://github.com/lbibass/OctoprintKlipperPlugin/blob/4689129ae32fa828f6fbedb7b51701176c3f544a/octoprint_klipper/static/js/klipper.js#L8) might need to be changed as well to not share the same name as original forked repo, but @kantlivelong or @foosel could probably confirm this. If that does need to be changed you will also have to change it [here](https://github.com/lbibass/OctoprintKlipperPlugin/blob/4689129ae32fa828f6fbedb7b51701176c3f544a/octoprint_klipper/static/js/klipper.js#L120).

There may be more, but that's what stands out right now.

lbibass:
So I've done some troubleshooting, and there's an unfortunate problem: changing the ID breaks the whole plugin. Nothing works. Which is unfortunate, to say the least. I have no idea WHY it breaks. It shouldn't. But it does.

jneilliii:
This is one of those situations where using the refactor? tool in PyCharm would be helpful. The plugin is probably broken because all of the knockout bindings throughout the templates are referencing `settings.settings.plugins.klipper...` and all those have to be switched too. I'm not that familiar with the refactor? tool, but vaguely remember foosel using it on one of the recent Live Code & Chat videos on YouTube.

jneilliii:
@lbibass I'd be happy to help you with your refactoring stuff. But looking into the repository and all the forks and merges and software updates of all the different forks pointing to different repositories is confusing me. Is your fork the most recent or is this one?

https://github.com/dushyantahuja/OctoprintKlipperPlugin

and then that one is pointing to this one for updates...I'm totally confused.

https://github.com/ShohninDmitriy/OctoprintKlipperPlugin/

Just trying to figure out who is actually going to be maintaining the official version of the BetterKlipper plugin and then I can help that person get their repo updated.

lbibass:
Sorry, my apologies. I will get to work on this. I think I'll start over from scratch so I can have a clean slate to work from.

jneilliii:
Ok, thanks. Closing this PR for now. Submit a new one when you're ready.

PR #440, opened 2020-04-22
PrusaSlicer Thumbnails Instructions

jneilliii:
Update instructions with proper configuration field name


jneilliii:
@kantlivelong not sure why CI is complaining on this simple update. You mind taking a look?

foosel:
It isn't your change, it's either Travis or Github - the link check failed for all github.com links across the whole repository and that flagged the build. And for some reason Travis also doesn't let me restart it right now, which is really weird. Merging.

PR #444, opened 2020-04-24
Continuous Print plugin

nukeem:
Continuous Print plugin


nukeem:
The save on_after_startup just ensures that the enabled setting is set to false after startup. If its set to true then it can mess with future prints that may not be using the queue.
Sorry didn't realise the comparability was required. Ill add it in

jneilliii:
@nukeem, I'm not sure if that boolean even needs to be a setting. Probably could get away with an internal assignment in `__init__`. I've seen weird things happen when saving at certain time, just can't remember if after startup was one of them. I will defer final review to @kantlivelong on this one and he can merge if he doesn't see any issues.

nukeem:
@jneilliii  Thanks

kantlivelong:
Yeah I don't get the point of saving `cp_enabled` if it's just going to essentially be initialized to `False` on startup. It seems to be being treated as a regular ol' bool in runtime. Not something I'd block this merge for though but you might want to re-eval it to see if it's really needed as-is.

@jneilliii Be sure to mark your approval on the review. :heavy_check_mark:

jneilliii:
I thought I had @kantlivelong

nukeem:
Thanks for the advice i will keep that in mind for the next release, i wasn't sure if using it as a variable would cause it to get garbage collected or something during run-time.

PR #445, opened 2020-04-25
New Plugin TemperatureLegendMover

OllisGit:

Thx, in advance
Olli


jneilliii:
@OllisGit, I think it would be best to separate these two commits because they are about two separate things...one is adding a plugin, the other is updating an existing plugin.

OllisGit:
Upps..that was not what I want.The Phy3 changed shouldn't be included, sorry!
I will create a new PR.

PR #450, opened 2020-04-27
Terminal Response

xia0:
just adding my humble plugin


OllisGit:
Some remarks:
- Currently I have no idea why the Build-Failed. In the past a restart of the build solve the issue.
Someone from the experienced developer should take a look on this.

- No need to push these file/folder to git: ```.pyc``` and ```.egg-info```.

Just create a ```.gitignore```-file that contains something like this:
```
*.pyc
*.swp
.idea
*.iml
build
dist
*.egg*
.DS_Store
*.zip
```
- Everything else looks fine

xia0:
> 
> 
> Some remarks:
> 
>     * Currently I have no idea why the Build-Failed. In the past a restart of the build solve the issue.
>       Someone from the experienced developer should take a look on this.
> 
>     * No need to push these file/folder to git: `.pyc` and `.egg-info`.
> 
> 
> Just create a `.gitignore`-file that contains something like this:
> 
> ```
> *.pyc
> *.swp
> .idea
> *.iml
> build
> dist
> *.egg*
> .DS_Store
> *.zip
> ```
> 
>     * Everything else looks fine:
>       image

Thanks for reviewing. I've added the .gitignore and removed those files.

PR #451, opened 2020-04-28
Mark MultipleUpload as python3 compatible

eyal0:


eyal0:
This fixes https://github.com/eyal0/OctoPrint-MultipleUpload/issues/3#issuecomment-619623478

PR #454, opened 2020-04-29
Added webhooks plugin

2blane:
Added the webhooks plugin to the list.


jneilliii:
Also, it appears that you are trying to load a navbar [here](https://github.com/2blane/OctoPrint-Webhooks/blob/9b81267e2932efdf4681acbeeaa4211089bec525/octoprint_webhooks/__init__.py#L39) but you don't have a jinja2 template for navbar in your templates folder.

jneilliii:
And your js file appears to be completely empty.

https://github.com/2blane/OctoPrint-Webhooks/blob/master/octoprint_webhooks/static/js/webhooks.js

2blane:
@jneilliii thanks for the review! I fixed all of these issues.
Default settings are changed.
There is an updated screenshot of the settings page.
I removed the navbar code - it was old and from a template.
I removed the js file - it was old and from a template.

Let me know if there is anything else to fix or if it's good to go.

2blane:
@jneilliii ok I have moved the image under the assets folder. Should be good to go.

PR #457, opened 2020-05-02
Add smartpreheat plugin

kantlivelong:
:wave:


kantlivelong:
Of course it had to be that. That's what I get for finishing it up at 2AM.

Issue #466, opened 2020-05-04
[Abandoned Plugin] OctoPrint-Display-ETA


foosel:
**Which Plugin? (URL to plugin page on plugin repository):** https://plugins.octoprint.org/plugins/octoprint-eta/

**Ticket on plugin's issue tracker confirming abandonment:** https://github.com/pablogventura/Octoprint-ETA/issues/14

**Is there someone open to adopting it? ("no" or @-mention user):** no

AlexVerrico:
Hi @foosel 
I have copied this plugin to a new repository (https://github.com/AlexVerrico/Octoprint-Display-ETA) and uploaded it to https://pypi.org/project/OctoPrint-Display-Print-ETA/. I have also submitted pull request #467 to the plugins repository.
As well as adding python 3 support, I believe I have fixed the following issues (I can't test any of them myself, but I am sure they were all the result of a colon in a gcode command which I have removed):

- https://github.com/pablogventura/Octoprint-ETA/issues/13
- https://github.com/pablogventura/Octoprint-ETA/issues/10
- https://github.com/pablogventura/Octoprint-ETA/issues/9

Cheers,
Alex.

foosel:
See my comment on #467, there's some open points there.

PR #467, opened 2020-05-07
Added plugin Octoprint-Display-ETA

AlexVerrico:
Please see issue #466


foosel:
There seems to be a bit of a mixup with your plugin identifiers and the plugin name 🤔 

  * `setup.py` states identifier as `octoprint_display_eta`
  * id in update registration is `display-eta`
  * repository is called `OctoPrint-Display-ETA`
  * plugin name is `OctoPrint-Display-Print-ETA`
  * markdown file here is called `octoprint-display-eta.md`
  * id in markdown file is `display-print-eta`

That's a bit too chaotic and will also cause issues 😅 Can you please make sure that the names and IDs and repository locations are a bit streamlined? At least the markdown file name, id in the markdown file and your plugin identifier in `setup.py` and the update info *must* be identical.

I'd also prefer to adapt the existing plugin listing to point to your fork if you are taking over maintenance and it is in fact a fork, instead of creating another listing and causing confusion.

AlexVerrico:
Hi @foosel 
Sorry for the mess it was in, I had just changed what I needed to change for it to install and work. 
I believe I have now fixed all the issues that you mentioned, but if I have missed any let me know.
I created a new entry on the plugin repository rather than updating the existing one since the original author is still active on github, but as I haven't been able to make contact to confirm that they are happy for me to take over the plugin, I have mirrored it to a new repo rather than forking it.
Cheers,
Alex.

AlexVerrico:
@foosel is there something else that needs to be fixed? Or has this just slipped through the cracks?

foosel:
> @foosel is there something else that needs to be fixed? Or has this just slipped through the cracks?

Simply didn't yet get around to do another review. Call it scalability issues.

> I created a new entry on the plugin repository rather than updating the existing one since the original author is still active on github, but as I haven't been able to make contact to confirm that they are happy for me to take over the plugin, I have mirrored it to a new repo rather than forking it.

That's a bit of a tricky situation. Thing is, I'd like to avoid confusing people with forks. There hasn't been any reaction to the ticket enquiring about maintenance status for four weeks now, so even if the author is still active on Github, apparently they are no longer looking into the issue tracker. Which means I'd really like you to take it over on the repository listing, as in, change the registered listing to point at your fork.

I fear we have a lot of de-facto abandoned plugins on the repo right now. If they all get registered a second (or maybe even a third or fourth...) time by someone who takes over maintenance, we'll soon drown in stale registrations AND sooner rather than later run out of meaningful identifiers to use to boot. 

So what I'd propose here is to go all in, replace the registration. I'll then push a notification so users who run the current version (which still will update from the old URL) know they need to update manually to your fork, and then we take it from there. If at some point the author takes an interest again, it might be an option to do joint maintenance or something like that. 

Considering the particular chaos that the original plugin was in though (as you noted as well) and the fact that it didn't persist any settings that would need to get migrated to a new plugin identifier, it might in this case be easier to just remove the old listing completely and go with your's here. That way there at least will no longer be that horrible ID confusion. So that's what I'm going to do here.

PR #473, opened 2020-05-12
OctoPrint_Signal-Notifier is py3 compatible

aerickson:
Confirmed syntax is fine. Tweak metadata entry.

### DIFF
--- a/_plugins/signalnotifier.md
+++ b/_plugins/signalnotifier.md
@@ -29,8 +29,8 @@ featuredimage: /assets/img/plugins/signalnotifier/signalnotifier.png
 
 compatibility:
   # list of compatible versions, for example 1.2.0. If left empty no specific version requirement will be assumed
-  octoprint:
-  - 1.2.0
+  octoprint: 1.2.0   
+  python: ">=2.7,<4" 
 
 ---
 


jneilliii:
@foosel / @kantlivelong, quick question. Does the compatibility > octoprint section of the jinja template need to be a list or is a single value allowed? It appears the build completed without errors, but wanted to verify before merging.

kantlivelong:
@jneilliii the format for the python version requirement submitted is valid. It follows the same format used by setuptools. See https://setuptools.readthedocs.io/en/latest/pkg_resources.html#requirements-parsing

kantlivelong:
@jneilliii Sometimes I think I can't read.

I believe single value or list will work.

https://github.com/OctoPrint/OctoPrint/blob/3fd88b4f1367c16eddeab4e2bab3fe0915a35043/src/octoprint/plugins/pluginmanager/__init__.py#L62
https://github.com/OctoPrint/OctoPrint/blob/3fd88b4f1367c16eddeab4e2bab3fe0915a35043/src/octoprint/util/version.py#L82

jneilliii:
Not sure @kantlivelong, but this is part of the parser for building the plugin repo pages, nothing to do with the plugin loading itself. Not sure if it really matters since the first version that supported plugins was 1.20, but the listing is now showing all versions, while when it is a single list item it would show like this.

**Compatibility Information**
OctoPrint: 1.2.0+

 Just don't know if that is built into the template parser or not to change it to `all versions`.

jneilliii:
Looks like it does parse that way if the line item is a single list with 1.2.0 based on `Auto Terminal Input` with a single list item of `1.2.0`.

kantlivelong:
That's correct. 1.2.0 translates to all versions.

https://github.com/OctoPrint/plugins.octoprint.org/blob/gh-pages/_includes/octoprint_compat_human

PR #477, opened 2020-05-16
Adding Playlist plugin.

jneilliii:
Create's a playlist of gcode files to automatically print back to back.


jneilliii:
Man...I don't know what I keep doing to break Travis...

kantlivelong:
Check your [update hook](https://github.com/jneilliii/OctoPrint-Playlist/blob/e94529afdd7b4a2aec175ab0a685de48f30d4887/octoprint_playlist/__init__.py#L318
).

Also, [overwriting core functions](https://github.com/jneilliii/OctoPrint-Playlist/blob/e94529afdd7b4a2aec175ab0a685de48f30d4887/octoprint_playlist/__init__.py#L185) is generally frowned upon so be careful with that.

jneilliii:
> Check your [update hook](https://github.com/jneilliii/OctoPrint-Playlist/blob/e94529afdd7b4a2aec175ab0a685de48f30d4887/octoprint_playlist/__init__.py#L318).
> 

This has been corrected.

> Also, [overwriting core functions](https://github.com/jneilliii/OctoPrint-Playlist/blob/e94529afdd7b4a2aec175ab0a685de48f30d4887/octoprint_playlist/__init__.py#L185) is generally frowned upon so be careful with that.

That's understandable, can you think of any other way to achieve the gcode stripping without this approach without modifying the files directly like using a pre-processor hook would do?

_plugins/playlist.md
```diff
@@ -0,0 +1,70 @@
...
+title: Playlist
+description: Create's a playlist of gcode files to automatically print back to back.
+author: jneilliii
+license: AGPLv3
+
+date: 2020-05-16
+
+homepage: https://github.com/jneilliii/OctoPrint-Playlist
+source: https://github.com/jneilliii/OctoPrint-Playlist
+archive: https://github.com/jneilliii/OctoPrint-Playlist/archive/master.zip
```
jneilliii, line 14: @kantlivelong can this take jinja variables like in the update hook like `{target_version}`?

foosel:
> Man...I don't know what I keep doing to break Travis...

It's not you. Travis hiccuped on some changes made to the page for the github stat stuff. I have to look into that on my side.

foosel:
Page build should be fine again after pulling from `gh-pages`

jneilliii:
@foosel so the build failed again, but this time it was actually on my markdown file. Seems Travis is getting choked up on the reprap wiki anchored gcode links. Is there some special way those need to be escaped or something because all the links are correct and my local markdown interpreter is able to render the page properly.

_plugins/playlist.md
foosel: I'm not him, but still: it can't.

jneilliii: Thanks, bummer, may have to rethink my github actions then.

foosel:
So the generated HTML looks solid. It seems that linkchecker has issues with fragment URLs maybe. No idea how to fix that right now, but I just pushed a change to the linkcheck call that should make it ignore anything from that wiki page.

kantlivelong:
@jneilliii did you want to merge this or are you making changes?

jneilliii:
I've been trying to wrap my head around this change you suggested, but I know there are existing users that may be impacted. If willing to accept the core comm function wrapping I'm good with you merging now.

jneilliii:
I'm just going to keep this plugin unlisted. I don't really want to put the work into getting rid of the function overwrites and know that's about to change anyway with the new comm layer.

PR #478, opened 2020-05-17
Added compatibility and renewed the description and screenshot

thijsbekke:
Checked for python 3 compatibility and added that to the md file. Also renewed the description and screenshot


thijsbekke:
sorry, didn't know that, pull request is updated.

thijsbekke:
hmm I thought the type of list `-` vs `*` was the issue. If somebody can point out why it is failing I am happy to resolve it.

foosel:
> If somebody can point out why it is failing I am happy to resolve it.

The reason it's failing is outside of you control, it's something I broke.

PR #482, opened 2020-05-20
added Python 3 Compatibility flag

jneilliii:


jneilliii:
Man I give up!!!!!!

foosel:
wtf is up with that test build on Github Actions O_O

foosel:
Let's ignore that for now, I'll remove that build step again, apparently it's broken, also didn't even do what it was supposed to do yesterday during the stream.

PR #483, opened 2020-05-20
Add Author-Link and Release-Infos in Jekyll-Template

OllisGit:
In the plugin-view you can now click on the Authors-Name and then you will be delegated to the ```/by_author/#xxx``` page.

Btw. I like the new stats-sections in the view very much. But I have "challenges" to rebuild the site.
After I executed the ```populate_additional_metadata.py``` script (the execution is done without an error, all pages now modified with the new stat-values) I can not start Jekyll.
I received this error:
```
Liquid Exception: undefined method `new' for BigDecimal:Class in /_layouts/plugin.html
jekyll 3.8.5 | Error:  undefined method `new' for BigDecimal:Class
```
Full execution-log is in the following text-file.
[jekyll-execution.log.txt](https://github.com/OctoPrint/plugins.octoprint.org/files/4659066/jekyll-execution.log.txt)

I can't identify which value raise this issue. Do you have a tipp for my, how I can analyse/debug this issue?

### DIFF
--- a/_build/populate_additional_metadata.py
+++ b/_build/populate_additional_metadata.py
@@ -15,6 +15,8 @@ PLUGINSTATS_7D_URL = "https://data.octoprint.org/export/plugin_stats_7d.json"
 PLUGINSTATS_30D_URL = "https://data.octoprint.org/export/plugin_stats_30d.json"
 
 GITHUBREPO_URL = "https://api.github.com/repos/{user}/{repo}"
+# GitHub limit the response size. You need to do "page-calls" to receive all values. Max-Default: 100
+GITHUB_MAX_PAGESIZE = 100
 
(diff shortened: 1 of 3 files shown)


jneilliii:
@OllisGit does this change work with multiple authors? There are a few plugins that have this situation.

OllisGit:
@jneilliii yes it is working.
E.g. 
```
Author
jneilliii, Gina Häußge
```
Delegates to: ```/by_author/#jneilliii-gina-häußge```

Btw: I found the solution for the BigDecimal-Issue: I need to downgrade the ruby-version from 2.7 to 2.6

OllisGit:
I also added the latest release-name and the release-count (if present). It looks like this:
image

PR #484, opened 2020-05-21
OctoPrint-EasyServo

jneilliii:
New plugin that controls two servos connected to the Pi from the OctPrint control tab.


_plugins/EasyServo.md
```diff
@@ -0,0 +1,56 @@
+---
+layout: plugin
+
+id: EasyServo
+title: Easy Servo
+description: Plugin to control two servos attached to a camera gimbal to control motion.
+author: iFrostizz, jneilliii
```
kantlivelong, line 7: I don't think multiple authors are currently supported in the build.

@foosel ?

jneilliii: @kantlivelong, @OllisGit said it is and the current plugins with multiple authors are linking correctly.

jneilliii: Here's an example.
https://plugins.octoprint.org/plugins/displayprogress/

kantlivelong: Hmmm.. Seems like it will take the value but doesn't break it down individually. Not sure if intentional. Won't hold this up then I guess.

Example:
https://plugins.octoprint.org/by_author/#jneilliii-gina-h%C3%A4u%C3%9Fge

Would think that it should separate them into:
https://plugins.octoprint.org/by_author/#jneilliii
https://plugins.octoprint.org/by_author/#gina-h%C3%A4u%C3%9Fge

jneilliii:
> [Update hook](https://github.com/iFrostizz/OctoPrint-EasyServo/blob/2adfed9648fe9be6f84d5cfd7173f3d12776779b/EasyServo/__init__.py#L100-L115) 🙃
> 
> ...and some info in [setup.py](https://github.com/iFrostizz/OctoPrint-EasyServo/blob/2adfed9648fe9be6f84d5cfd7173f3d12776779b/setup.py#L26-L30)

@kantlivelong, all previous issues corrected.

Issue #486, opened 2020-05-23
[Feature Request] add closed issues to github stats


jneilliii:
**What do you want to see implemented in the plugin repository?**

Now with the updated github stats on the individual plugin pages I think it would be good to include closed issues as well.

**Why, what problems would be solved by this?**

Better understanding to the end user the activeness of the plugin authors possibly.

OllisGit:
Hi @jneilliii ,

this is a little bit tricky, because the value of "closed issues" is not included in the currently used overview response (only: ```open_issues```).
Instead of using the "Issue-API" we can use the search API:
```
GET https://api.github.com/search/issues?q=repo:OllisGit/OctoPrint-DisplayLayerProgress+type:issue+state:closed&per_page=1
```

Or maybe better: Use the GraphQL-API to collect all informations with one call...I never used GraphQL, but I will take a look.

foosel:
This is now live

OllisGit:
Cool, compact visualisation!

PR #489, opened 2020-05-23
Added "closed issue count" to GitHub-Stats

OllisGit:
Based on the feature request from @jneilliii https://github.com/OctoPrint/plugins.octoprint.org/issues/486

I looked into the graphql-api, but also with that approach I can't reduce the API-Calls.
So, I just added a search-call to collect the missing "closed-count" 

Result:
image

### DIFF
--- a/_layouts/plugin.html
+++ b/_layouts/plugin.html
@@ -124,15 +124,15 @@ layout: default
         <dt>Github stats</dt>
         <dd>
           <ul>
-            {% if page.github.releases %}
+            {% if page.github.releases > 0 %}
             <li>Latest Release: <a href="{{ page.github.latest_release_url | strip_html | strip_newlines | escape }}" target="_blank">{{ page.github.latest_release | strip_html | strip_newlines | escape }}</a></li>
-            <li>Release Date: {{ page.github.latest_release_date | date: "%-d %B %Y"}}</li>
+            <li>Release Date: {{ page.github.latest_release_date | date_to_string}}</li>
             <li>Releases: {{ page.github.releases }}</li>
             {% endif %}
             <li>Stars: {{ page.github.stars }}</li>
             <li>Open Issues: {{ page.github.open_issues }}</li>
             <li>Closed Issues: {{ page.github.closed_issues }}</li>
-            <li>Last update: {{ page.github.last_push | date: "%-d %B %Y" }}</li>
+            <li>Last update: {{ page.github.last_push | date_to_string }}</li>
           </ul>
         </dd>
       </dl>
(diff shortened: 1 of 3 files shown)


_layouts/plugin.html
```diff
@@ -130,7 +130,7 @@ <h2>Pictures</h2>
             {% endif %}
             <li>Last update: {{ page.github.last_push | date: "%-d %B %Y" }}</li>
             <li>Stars: {{ page.github.stars }}</li>
-            <li>Open Issues: {{ page.github.issues }}</li>
+            <li>Open/Closed Issues: {{ page.github.open_issues }}/{{ page.github.closed_issues }}</li>
```
foosel, line 133: Since we get closed from a different request, this should really check if we have a value there. Otherwise it might look funny on request errors.

OllisGit: My intention was to give a "compact" overview instead of multiple-lines.
But now, after I added a better response-handler, I can see that I sometimes reach the "search rate-limit" (with auth-user). 
It is different (only 30/minute) from normal APIs-Calls (5000/minute).

Just increasing/adding a delay is probably the easiest solution...but not very smart ;-)
So, maybe I will find a better solution.

OllisGit: I looked into the graphQL-stuff and found a solution to grab all git-informations by one call. Also the rate-limit should not be an issue anymore.

Here is the result:
image

Before I implement at into python, some questions:

- @foosel Currently we are using basic-auth in the python script. Because GraphQL only woks with tokens, and password-auth is deprecated (https://developer.github.com/changes/2020-02-14-deprecating-password-auth/), we need to switch to a personal-access token.
Is this a problem?

- Should we collect other informations (e.g. Wiki-Link, Contributors)?

foosel: I'm actually already using a personal access token, just via Basic Auth. So that shouldn't be an issue.

And great news on the single call! 

For now I'd say let's leave it at what we already grab, we can always add more.

OllisGit: Change to GraphQL is done! 
Now you need to set the "personal access-token" in the env-variable.

One Repository was renamed, because graphQL-Query needs the current name and not the "inital" name -> Error: Repo not found!

If you open the initial url you will be redirected to the new url
- Old: https://github.com/ryanneufeld/OctoPrint-EEprom-Marlin
- Nwe: https://github.com/ryanneufeld/OctoPrint-EEPROM-Marlin-Printrbot

OllisGit: @foosel from my point of view everything is done, so I resolve this conversation.

OllisGit:
@foosel Upps, I didn't expect that github delete the whole conversation. 
Anyway the implementation is done, please let me know if something missing.

Thx, in advance.
Olli

PR #492, opened 2020-05-23
Add Ikea Tradfri plugin

ralmn:
Add Ikea Tradfri plugin

I applied the suggestions of #339


ralmn:
I just made the change in date and compactibility. But I don't understand your remarks about the plugin identifier on `get_update_information`. It's the same of https://github.com/ralmn/OctoPrint-Ikea-tradfri/blob/master/setup.py#L7 for me ...

PR #496, opened 2020-05-27
Added plugin CooldownNotification

gmccauley:
This plugin is designed to execute gcode commands when the print bed cools to a certain temperature.


jneilliii:
Nice @gmccauley very good plugin. The first Python 3 only, although it seems that there's nothing in your python that wouldn't work in Python 2. I'll go ahead and merge, but just wanted you to be aware that it should work in the older Python. Also, would probably be compatible with all versions of OctoPrint that support plugins.

gmccauley:
Thanks @jneilliii, yeah, it will support both versions of python.  I mistakenly thought the goal was python3 only.  Will do another pull request to update to both.

As for OctoPrint versions, I just don't have the background of what was introduced in which versions and I only did my dev/testing on 1.4.0 so that's all I felt comfortable putting as the minimum.

jneilliii:
no problem, I'd probably do the same...helps push people to update...

PR #505, opened 2020-06-01
Added plugin LaMetric

thijsbekke:
Added plugin LaMetric. This plugin will send notifications too your LaMetric device. 

When you look at the source of the plugin you will notice that I had to not verify the SSL when connecting too the [LaMetric ](https://lametric.com/) device. De documentation of LaMetric states the following:
```
We recommend to use secure way of accessing the API via https using port 4343:
--insecure option must be added because of the random IP address LaMetric may have, and it is
 not possible to verify host stored inside the certificate.
```
https://lametric-documentation.readthedocs.io/en/latest/reference-docs/device-discovery.html

The connection will stay inside the network, so that is ok, I guess.


tedder:
@thijsbekke do you mind modifying the description to say "LaMetric smart clock" or something of the sort?

thijsbekke:
@tedder I don't mind, I think. Can I ask why?

tedder:
When the plugin shows up in the list or in the octoprint plugin notification, it says "Display progress of your Octoprint instance on your LaMetric." There's no context. It's certainly possible to find out more, either in the plugin's repo or by googling, but it's importing jargon from a different "universe" than the 3D printing world. I mean, I would kinda expect people to recognize terms like thingiverse, raspberry, slack (maybe), and prusa, but something like LaMetric isn't in-universe and it doesn't define itself either. Does that make sense?

thijsbekke:
@tedder See pull request https://github.com/OctoPrint/plugins.octoprint.org/pull/514, thnx for the feedback.

PR #508, opened 2020-06-02
 OctoPrint-Widescreen

jneilliii:
Plugin that adds a right sidebar to the web interface to utilize the full width on a widescreen monitor.


OllisGit:
Hi @jneilliii ,

I just started looking into that plugin. It looks relay promising...now we can use the free space and everything is scaleable, great!

One small issue:
image
If I move a section from one side to the other I can repeat this step several times.

jneilliii:
Yeah, I'm not quite sure what causes that actually. I think it's related to how I'm binding the arrays maybe.  It seems to only be an issue the first time. If you restart octoprint and reload the page does it still do it?

jneilliii:
@OllisGit I have resolved the binding issue and converted to a draggable UI in the settings for improved usability.

OllisGit:
After I cleaned the plugin-data everything was working.
Drag&Drop feature is a little bit "itchy", but it id working if you hit the right area.

PR #517, opened 2020-06-04
prusa_mini_eta plugin submission

vookimedlo:
Hello,

This is my first attempt to submit my plugin to the OctoPrint.
The plugin is dedicated to the Prusa Mini users and brings more accurate printing ETA.

Thanks,

Michal


jneilliii:
How is this different from the already existing M73 plugins available?

https://plugins.octoprint.org/search/?q=m73

jneilliii:
Outside of the above question, you also need to update your submission so the id in the template matches your plugin_identifier.

vookimedlo:
> How is this different from the already existing M73 plugins available?
> 
> https://plugins.octoprint.org/search/?q=m73

At least, what I have seen, the none of them provides the approximate estimate before the printing starts.

If there is a problem with duplicity, please close this PR. I can live without the official plugin repository.

jneilliii:
Thanks for the fix. After further review it appears that you are right. This is an improvement over the other ETA variants as it also has analysis factory. I'll give it some testing to verify everything is working and merge once completed. Thanks for the quick response.

jneilliii:
I'm assuming this would work with any printer profile that enables M73 in PrusaSlicer wouldn't it, not just the Mini?

PR #518, opened 2020-06-05
Signalclirestapi plugin

bbernhard:
Another Signal Messenger Plugin.


bbernhard:
> @bbernhard I think it would be a good idea to include some language requiring there are pre-requisites for this plugin to work and either include the docker instructions or link back to your github instructions to your submission. I don't see this would be a problem merging once that's included.
> 
> Mainly trying to keep the issue/requests to a minimum for those that don't look at the original repo and just install the plugin and expect it to work.

Good idea! 

I've updated the PR - please let me know in case there's missing something :)

PR #523, opened 2020-06-11
Added SMuFF plugin V1.1.4 to repository

technik-gegg:


jneilliii:
@technik-gegg nice plugin. I've merged, but would recommend updating the names to be consistent between the plugin's name in setup.py to the override in init.py and the update check callback. Not really an issue.

jneilliii:
@technik-gegg just noticed when the page built that your second screenshot didn't show up. File names including extensions are case sensitive. Please correct that.

technik-gegg:
@jneilliii 
Thanks for the hint. Did correct both.

jneilliii:
@technik-gegg you'll need to submit a new PR for the case change here.

technik-gegg:
Silly me. It was the wrong pic anyways. Now it's the right one.

jneilliii:
thanks

PR #528, opened 2020-06-14
Added plugin DuCalibrator

Fabi0San:
An [OctoPrint](https://octoprint.org/) Plugin for performing micro calibration in linear delta machines.

[README](https://github.com/Fabi0San/DuCalibrator/blob/master/README.md)

## Key features:
* Automatic probing
* Calibrates up to 18 geometry parameters 
* 3D visualization of probed points and estimated points after proposed calibration


Fabi0San:
I don't quite understand the integration error, can someone please shed some light here?

jneilliii:
The build error is specific to a new labeler action, not sure why that is failing. Took a quick look and it appears your including three.js in your plugin which is part of the core octoprint and used in the gcode viewer. Have you experienced any issues related to that redefining of the included core library? May need to get @foosel to check in on this one.

foosel:
Turns out the labeler action is broken by design due to limitations in Github Actions (they even mentioned that in the README but I missed it). I just removed it again, sorry for the confusion.

Otherwise seconding the `three.js` issue, please don't include that, it's already included in OctoPrint and including it again could lead to hard to debug issues in the field.

I also just spotted a `.vscode` folder checked into your plugin, that also doesn't belong and should be removed.

Fabi0San:
Thank you Gina, will do.

`Otherwise seconding the three.js issue, please don't include that, it's already included in OctoPrint and including it again could lead to hard to debug issues in the field.`
I'm stumbling on this a bit, I tried simply removing it but then couldn't reference classes in THREE, is it nested on some viewmodel?

`I also just spotted a .vscode folder checked into your plugin, that also doesn't belong and should be removed.` 
Whoopsy.

jneilliii:
Let me know @Fabi0San when those changes are complete and I can run through the final review.

Fabi0San:
JS is not really my jam but I looked a bit today and it seems GCodeViewer uses Three.js [revision 52](https://github.com/mrdoob/three.js/releases/tag/r52) which is from 2013 and defines THREE classes in the local scope, therefore not accessible from other plugins.

My plugin uses [revision 90](https://github.com/mrdoob/three.js/releases/tag/r90) from 2018 which defines its classes in the global scope, also not great as plugins would have to use the same version and upgrade together.

Given THREE doesn't try too hard being [backwards compatible](https://github.com/mrdoob/three.js/wiki/Migration-Guide), shipping a single version for all plugins may bring a few headaches. 

So I suggest we allow plugins to bring their own Three version as long as they keep it in local scope.

@jneilliii @foosel , what do you think? Is there a better alternative I'm missing?
Thanks!
Fabio.

foosel:
> So I suggest we allow plugins to bring their own Three version as long as they keep it in local scope.

I can live with that.

jneilliii:
The PrettyGocde plugin also includes a three.js in its repo. Since I too am no js wizard, how would one keep it in local scope if the library is a global define?

Fabi0San:
I spend the whole day yesterday trying to find an elegant solution to this and it seems the best is referencing THREE as a module but for that we need the plugin viewmodel to be a module and for that we need some extensive changes in octoprint and how it packs and publishes the viewmodel javascript for plugins. 

I assume that is too much work for now, but we should think about how we can better handle multiple libraries/dependencies/etc in octoprint. Ideally the plugin would manifest the libraries/versions it needs some packager fetches them to octoprint host and octoprint loads and exports them on the client's browser so the plugin viewmodel can elegantly import them.

As a compromise, we can either change the THREE.js code so the THREE class is defined in the block scope the way it was back in r54 (single line change) or we can after importing it as is copy globalThis.THREE to a block level THREE and set globalThis.THREE back to undefined. Which IMO is uglier.

Do you guys have different ideas?

In any case I've updated my plugin to use the latest three.min.js with the one line change that makes THREE a block scope class.
https://github.com/Fabi0San/DuCalibrator/commit/b882bb8ae69ae06b028844a3724711840885d07f

jneilliii:
I will have to defer to @foosel on this one. I'm curious if this internal library would benefit from an upgrade anyway or if it would break the gcode viewer. As mentioned, there is the Pretty GCode plugin and I just remembered the GcodeThumbs plugin that had also included three.js. Of course none of these plugins have been submitted to the Plugin Repo here so it hasn't come up yet.

Fabi0San:
@foosel can you please take a look and offer some guidance?

foosel:
Internal lib would prolly benefit from an update, however not for 1.4.1. If the lib question is the only thing holding this up, here's my go ahead.

And sorry, I'm trying to finalize 1.4.1 and hence had my head elsewhere and dropped the ball on this one.

Fabi0San:
Thank you very much guys!

PR #529, opened 2020-06-15
Added plugin homeassistant

cmroche:


jneilliii:
Just curious, how is this any different than the HA integration?

foosel:
For the record, lease ignore the Pull Request Labeler check fail.

cmroche:
@jneilliii The octoprint integration uses the web api, but requires manual setup of integration directly through the configuration file. It's also a local polling sensor, os it constantly tries to hit that web-api, and IIRC would dump errors into the log if the service was offline and slow update of sensor values.

MQTT is more robust in this case in that it becomes a local push integration, by simply installing this plugin (and having MQTT configured) it will auto-register the devices and make them available in the device and entity registry, and it better handles the availability of the service marking the corresponding sensors as "unavailable" when the service goes offline without requiring any special integrations or setup on HomeAssistant.

There are a number of integration on Home Assistant that allow the use of a direct integration, or MQTT as a large amount of the HA community uses and prefers the MQTT approach. Off the top of my head Z-Wave, and Zigbee devices, Tuya devices, Tasmota device, ESPHome devices, etc.

It could be interesting to provide an auto-configure for the OctoPrint directly to HA in the future, but even then without a change to a local-push model this integration will be a bit more robust.

jneilliii:
ok thanks @cmroche I'll test the plugin and see how it goes. I just so happen to have all the pre-requisites.

jneilliii:
First stumbling block. Your URL is misspelled [here](https://github.com/cmroche/OctoPrint-HomeAssistant/blob/3298d0719e33804987a08c6d5d485115cbf441c5/setup.py#L30), and therefore the homepage link in plugin manager is wrong.

cmroche:
@jneilliii Ooooups. Fixed :)

jneilliii:
Thanks. I think that appears to be the only issue. Seemed to show right up in HA, nice addition.

foosel:
I took the liberty to add the missing python compatibility string. It's py3 compatible after all, but without that it wasn't detected as such. 

Might have to look into populating that automatically, it's a common thing to forget...

PR #533, opened 2020-06-22
Name change due to the merge of OctoPrint Anywhere and TSD

kennethjiang:
Also the changes to reflect the new functions and features.

### DIFF
--- a/_plugins/thespaghettidetective.md
+++ b/_plugins/thespaghettidetective.md
@@ -2,9 +2,9 @@
 layout: plugin
 
 id: thespaghettidetective
-title: The Spaghetti Detective
-description: An essential plugin if you print remotely. Monitor and control your printer over the internet. No port-forwarding or VPN is needed. Best part? AI-based failure detection!
-author: The Spaghetti Detective Team
+title: Access Anywhere - The Spaghetti Detective
+description: Monitor and control your printer anywhere over the internet, on your phone! No port-forwarding or VPN is needed. Best part? AI-based failure detection!
+author: TSD Team
 license: AGPLv3
 
 # TODO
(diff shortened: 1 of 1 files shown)


jneilliii:
@foosel and @kennethjiang based on this merging of plugins should we remove the one that is no longer going to be actively developed?

kennethjiang:
Thank you @jneilliii for your prompt action!

I'd love to retire OctoPrint Anywhere as soon as I can.

In all places I have control over, I have made it clear that OctoPrint Anywhere has been merged into TSD. However, I'm myself surprised by how many references to OctoPrint Anywhere are still around. Try [this search](https://www.google.com/search?q=octoprint+remote+monitoring&oq=octoprint+&aqs=chrome.0.69i59l3j69i60l5.3599j0j7&sourceid=chrome&ie=UTF-8).

I'm open to any suggestion that can safely retire OctoPrint Anywhere without leaving a lot of broken links on the internet that can result in confused users.

foosel:
We could remove the old entry for OctoPrint Anywhere and add a redirect to TSD. That would leave backup restores containing a reference to OctoPrint Anywhere incomplete however (no way to redirect those atm, and it would also feel wrong to install a different plugin, even if it's the official successor).

kennethjiang:
Our goal is definitely to deprecate and remove the old plugin in time. The redirect is a really good idea. We only started trying to ease users over to the new plugin a few months ago and so maybe it would make sense to give them a bit more time before submitting the hard redirect?

In the mean time we're working with other bloggers to update their content and links, and to make sure the language is clear on what's happening. Right now the OA plugin page has a pretty clear "soft redirect" to the TSD plugin page and we're hoping this rename may help get people straight to the new plugin out of the gates. Maybe revisit the hard redirect in a few months?

jneilliii:
I'm going to go ahead and merge these changes @foosel. I'll let you decide on the redirect scenario later. Couldn't @kennethjiang handle the migration from the software update hook callback within the two plugins later down the road? For example, push an update to the deprecated plugin whose software update hook then checks against the successor plugin? I realize it could be tricky getting the versioning in line, but seems like that might be an option that he could control himself?

kennethjiang:
Thank you @jneilliii @foosel for being responsive and helpful, as always. I just love how OctoPrint dev community works closely together to make 3D printing more accessible for everyone!

PR #537, opened 2020-06-24
Add RewriteM600

wgcv:
Add to the plugins RewriteM600, rewrite M600 to use without support.

https://github.com/wgcv/RewriteM600


jneilliii:
For your pull request you need to set the featuredimage property to also go to the same address as your screenshot. Your plugin is flagged for Python 2 only. Python 2 is end of support and we are no longer accepting plugins that don't support both versions. Your plugin's `__init__.py` file indicates the plugin is Python 3 compatible, so you probably just need to update the flag in your PR here.

On initial review of the plugin itself I have a couple of issues that will need to be addressed. 

1. Why do you have the SettingsPlugin mixin if you aren't actually utilizing it? 
2. In your [onDataUpdaterPluginMessage](https://github.com/wgcv/RewriteM600/blob/67e45f4e28f1b18d32d7e3c4f01edc6346d98361/octoprint_RewriteM600/static/js/RewriteM600.js#L16) callback it would be good to have a check at top for your plugin rather than process on all commands. If any other plugin sends a message with `data.type == "popup"` it will trigger your plugin. Please add a line at the beginning of that function like below.
```
			if (plugin !== "RewriteM600") {
				return;
			}
```

_plugins/RewriteM600.md
```diff
@@ -0,0 +1,92 @@
...
+- color change
+- hook
+- M600
+
+screenshots:
+- url: /assets/img/plugins/RewriteM600/M600-in-action.png
+  alt: Askin to change the filament
+  caption: Ask to change filament - M600
+
+featuredimage: /assets/img/plugins/RewriteM600/M600-in-action.png
```
wgcv, line 34: Solved sorry I forget these.

_plugins/RewriteM600.md
```diff
@@ -0,0 +1,92 @@
...
+  # Compatible Python version
+  #
+  # Plugins should aim for compatibility for Python 2 and 3 for now, in which case the value should be ">=2.7,<4".
+  #
+  # Plugins that only wish to support Python 3 should set it to ">=3,<4". 
+  #
+  # If your plugin only supports Python 2 (worst case, not recommended for newly developed plugins since Python 2
+  # is EOL), leave at ">=2.7,<3"
+      
+  python: ">=2.7,<4"
```
wgcv, line 87: I was thinking that it was for python 3 😶 but was set for python2. Work well in Python3  👀 🎉

PR #539, opened 2020-06-25
Add python compatibility (and change date)

b-morgan:
Update OctoPrint-DisplayZ with python compatibility


jneilliii:
@b-morgan the date should never change on these from the original submission date. I'm going to change that back and once @foosel merges your PR we can merge.

PR #540, opened 2020-06-27
Update descripton and screenshot

razerraz:


jneilliii:
For future PRa please try to limit the number of commits to 1.

PR #547, opened 2020-07-03
Adding M150 plugin

horfee:


OllisGit:
Hi @horfee,

I looked into your plugin and found the following points:

- Is your plugin working with the latest OP Version? Or only with 1.2.0?
- Is it working with Python3?
- The description of the plugin is missing in M150Contol.md
- In __init__.py at the beginning is a "marked todo"

Please look at these topics

BR
Olli

jneilliii:
@OllisGit if there is just a single version listed in the octoprint compatibility list that is considered as a minimum version requirement.

horfee:
Hello @OllisGit 
I reviewed the points you mentionned.
- As defined in M150Control.md, it is compatible with python 2 and 3.
- I removed the TODO tags in __init__.py
- Added the plugin description in M150Control.md
- 1.2.0 is the minimum requirement, but working on 1.4.0 (I did not tested on 1.4.1RC)

PR #553, opened 2020-07-15
Added plugin TimeToFilament

eyal0:


jneilliii:
I'll defer to @foosel on reviewing this plugin as it appears you are overwriting internal core methods/functions (link below) and that is typically frowned upon.

https://github.com/eyal0/OctoPrint-TimeToFilament/blob/8067c047e0abc1a4b2badccc89caba0641f70b8f/octoprint_TimeToFilament/__init__.py#L112-L113

eyal0:
I was trying to think of a way to not need to do that.  Is there some modification to OctoPrint that would make it possible to do this stuff without all the [monkey patching](https://stackoverflow.com/questions/5626193/what-is-monkey-patching)?

One way would be to add hooks to functions.  The problem is, there are a lot of functions and you never know how many hooks you'd need.  It could get out of control keeping track of all the different hooks and who is overriding.  Especially if you want to be able to add and remove hooks on the fly.  It's not so simple.  The advantage is that it's easy for plugins to hook and unhook.

Another way would be to allow plugins to hook into the creation of major classes, like self._printer .  Perhaps it could be possible for plugins to register a function that lets them override?  Like, plugins would have a function `def override_printer(_printer):` that needs to return a printer object.  The default would just be to return _printer.  However, classes would likely take _printer, monkey patch it, and then return the new, monkey-patched version.  The advantage would be that it would be clear to see which plugins did monkey patching and which didn't.  And the docs for the function could explain how it's important for the monkey-patch to not assume that is it the first nor last to monkey-patch, and that it should maintain the contract of the underlying function so that other monkey-patches work together correctly.  This isn't really different from what I'm already doing, it's just more formalized and structured so people will be less likely to get it wrong.  And maybe OctoPrint might have a way to maintain the un-monkey-patched version so that it could override the wishes of the user/plugin and undo the monkey-patch.

Another way is using reflection.  Each object, like the Printer object, could maintain a map from function_name to plugin name to replacement function.  Plugins would write into this map, like if plugin Foo wants to replace method bar with method baz, it would be `the_map['bar']['Foo'] = baz`  And then, every function in the class is modified to check the map before doing the work.  [Modifying every function](https://stackoverflow.com/questions/6307761/how-to-decorate-all-functions-of-a-class-without-typing-it-over-and-over-for-eac) can be done as a decorator that uses reflection to loop through all the functions in the class.

So there are three possibilities.

Of course, you can just ignore it and say "Caveat Emptor".  You download a plugin, you have to trust the author.

Other ideas?

jneilliii:
Without deep diving into what's being monkey patched by your plugin, there are a couple of hooks that could be related to your suggested approach. For example, [octoprint.printer.factory](https://docs.octoprint.org/en/master/plugins/hooks.html#octoprint-printer-factory) and [octoprint.printer.estimation.factory](https://docs.octoprint.org/en/master/plugins/hooks.html#octoprint-printer-estimation-factory) are a couple. Pretty sure any of the "factory" hooks were a step in achieving manageable monkey patching.

foosel:
Monkey patching should really only ever be the last resort and only be done in extreme circumstances. Fixing a crippling bug via a plugin? Sure, monkey patch. Adding functionality? Let's talk. If you need to monkey patch a part of OctoPrint in order to achieve something with your plugin, then that means there's an extension hook missing in OctoPrint somewhere and should be added. 

The problem with monkey patching is, it might break your plugin on the next update. It might cause horribly hard to debug and diagnose issues. And in the end, it's just a dirty hack. Yes, I do it in OctoPrint core here and there. I've done it in places where I had to modify third party dependency behaviour to work around bugs or limitations. And it's cost me DEARLY over the years and I try to get rid of monkey patches as soon as I can. They are just... stop gap solutions. Hacky band aid patches. They should not be the final solution to a problem. I'd be open to something like "if OctoPrint version < X, monkey patch, otherwise use hook/mixin", but then we also need to get going on adding the hook or mixin in question.

So, as it is now - I'm not too happy with seeing this published as-is. You are not only accessing a non public member of the printer here, you are even overwriting a non public method on the non public member. And then your replacement also doesn't give a damn about accessing non public stuff either. That's just asking for trouble, unstable AF and hacky. Let's talk how to achieve what you need to do here. Parsing a gcode file from within the state monitor certainly doesn't sound like the right way.

eyal0:
Okay sure, let's figure it out.

Currently, the printer object sends updates to the web UI regularly with how long until the print is done.  I would like to add stuff like, "how long until I need to change filament".  So I want to enhance the JSON message that OctoPrint is sending to include that.

I can try to use the octoprint.printer.factory but the problem with it is that it is not composable.  If the user installs two plugins that hook this same octoprint.printer.factory, only one plugin will work and the other will be ignored because as soon as one succeeds, [the loop ends](https://github.com/OctoPrint/OctoPrint/blob/master/src/octoprint/server/__init__.py#L450).  The solution that I did might look like a hack but it *is* composable.

If there were a hook in the sending of regular data to web UI, that would work.  The hook could be [here](https://github.com/OctoPrint/OctoPrint/blob/6232a04e52fbf09a8b912590718764fccc3598ba/src/octoprint/printer/standard.py#L885).  I bet a lot of plugins might find that useful!

foosel:
> If there were a hook in the sending of regular data to web UI, that would work. The hook could be [here](https://github.com/OctoPrint/OctoPrint/blob/6232a04e52fbf09a8b912590718764fccc3598ba/src/octoprint/printer/standard.py#L885). I bet a lot of plugins might find that useful!

See, and that's why I prefer talking about stuff like this instead of just going "eh, whatever" about monkey patching ;) I agree, that would be a neat addition! However, considering that the location you marked is inherently bound to the progress in the file, but plugins might want to send regular status updates along side the regular stuff not related to the progress, my suggestion would rather be to extend it a couple of levels further up, add a plugin specific name space to the current (and historic) data pushs and allow that to get populated on each push. Let me think a bit about where to best locate that (such that it also doesn't interfere with anything).

eyal0:
Okay.  So perhaps update the [StateMonitor](https://github.com/OctoPrint/OctoPrint/blob/6232a04e52fbf09a8b912590718764fccc3598ba/src/octoprint/printer/standard.py#L1399).  You could modify it to be generic so that any number of hooks could hook in to make state updates instead of just having one for temperature and one for progress.  And the Printer class which uses the StateMonitor would hook into it the same way that the plugins do instead of passing in functions for the temperature and progress change.

So that's a more general solution.

jneilliii:
> plugins might want to send regular status updates along side the regular stuff not related to the progress

This sounds similar to the temperature changes that were merged that allowed for plugin's to add additional data to the callback.

eyal0:
> > plugins might want to send regular status updates along side the regular stuff not related to the progress
> 
> This sounds similar to the temperature changes that were merged that allowed for plugin's to add additional data to the callback.

Ah, where is that in the code?  Maybe it would suit me?

jneilliii:
It's coming in the next 1.4.1 release of OctoPrint. An example of it's usage is linked below, but the current temperature graph implementation doesn't make use of it yet. You'd have to install the plugin that the example linked below is in to graph it.

https://github.com/jneilliii/OctoPrint-PlotlyTempGraph/blob/master/test_plotly_graph_fan_speed.py

OllisGit:
Sorry, for interrupting the discussion about "Monkey-Patching".

But because my DisplayLayerProgress-Plugin already include this functionality (ChangeFilamentTimeLeft, EstimatedChangeFilamentTime and ChangefilamentCount), 
I am very interested how your implementation works.

You already mentioned your plugin in this [issue](https://github.com/OllisGit/OctoPrint-DisplayLayerProgress/issues/115), but to be honest I couldn't figure it out how your estimation algorithm works.

So it would be nice, if you could provide a short description how your plugin works.

Btw. you should remove the TODOs in your readme before publishing the plugin.

Thx, in advance
Olli

eyal0:
You're asking how the estimation works or how the monkeypatching was done?

The estimation is done by PrintTimeGenius and TimeToFilament requires PrintTimeGenius.

PrintTimeGenius includes an analyzer which a compiled version of Marlin source code that simulates printing, but much faster.  It is able to compute how long every line of gcode will take, include acceleration, junction deviation, max jerk and all the others setttings.  It runs actual Marlin code that is lightly modified.  That produces a mapping from file position to total printing time.  That mapping is stored for each file.

PTG also installs an PrintTimeEstimator, which is a hook that OctoPrint provides to let you override the built-in one.  It uses the pre-computed mapping to provide accurate print time estimates.

---

So all that is how PTG works.  Then some users asked to know how long until the filament change is needed and they wanted it in PTG.  I didn't want to muddy PTG with that so I modified PTG's estimator a little so that it can estimate not only the time until the print is done but the time until any line in the gcode.

Then I made TimeToFilament.  TimeToFilament reads through the currently printing gcode file looking for the next occurrence of a filament change.  It gives that to the currently installed estimator (PTG) and gets the answer.  That answer is added to the twice-per-second info that OctoPrint sends to the web UI about progress.  TTF on the web side parses that and inserts it into the webpage.

---

Hope all that makes sense.  Let me know if something is unclear.

foosel:
@eyal0 take a look at https://docs.octoprint.org/en/maintenance/plugins/hooks.html#octoprint-printer-additional-state-data

eyal0:
> We need to talk about the purpose for the monkey patching and how to achieve what you want to achieve in a less hacky way.

Done.  I'm using the new hook now.

Please review again.  Thanks.

eyal0:
> We need to talk about the purpose for the monkey patching and how to achieve what you want to achieve in a less hacky way.

@foosel I fixed this already.  How do I get this reviewed again?

foosel:
Sorry, that got buried. @OctoPrint/plugin-repository-maintainers can one of you take a look?

jneilliii:
Yeah, I'll take a look @foosel.

PR #562, opened 2020-07-25
OctoPrint-Consolidated Tabs

jneilliii:
Plugin that allows combining tabs into a single tab as draggable and resizable panels.


jneilliii:
@foosel confused here about the check failure...are my old eyes that bad?

OllisGit:
...and an other piece of software from the op plugin-factory @jneilliii ;-)

Looks pretty nice, so I installed the plugin in my dev-environment:
OP 1.4.0
Py: 2.7.9 and 3.6.8

but I received the following error in the browser-console:
image

jneilliii:
That first error is strange. Probably need to add in something to handle nulls in that onAllBound viewmodel callback, although it should only be processing viewmodels with onTabChange or onAfterTabChange callbacks themselves. What all plugins do you have installed?

I have seen the second error, as it's related to the temperature graph initialization. Doesn't seem to effect functionality though, so not really sure if it's an issue. My PlotlyTemGraph plugin doesn't have the same error.

foosel:
There's a bug in the image path validation of the PR check, see #565. Looking into it and fixing ASAP.

foosel:
Check failure has been cleared and #565 should be fixed. There was a warning due to externally hosted images: https://github.com/OctoPrint/plugins.octoprint.org/actions/runs/182462956

jneilliii:
@OllisGit I just pushed updates to the plugin that resolves the temperature error issue. Can you please re-review and approve/merge?

OllisGit:
Hi @jneilliii ,

I analysed the first issue. The reason is that you expect an string for the binding, but my plugins using jquery-elements for the viewmodel binding.
```
		self.onAllBound = function(allViewModels) {
			// get all available tab change callbacks
			ko.utils.arrayForEach(allViewModels,function(item){
				if((item.onTabChange || item.onAfterTabChange) && item !== self){
					ko.utils.arrayForEach(item._bindings,function(binding){
							self.tab_callbacks()[binding.replace('#','')] = item;
						});
					}
				});
```
image


My plugin:
```
    OCTOPRINT_VIEWMODELS.push({
        construct: PrintjobhistoryViewModel,
        // ViewModels your plugin depends on, e.g. loginStateViewModel, settingsViewModel, ...
        dependencies: [
            "loginStateViewModel",
            "settingsViewModel"
        ],
        // Elements to bind to, e.g. #settings_plugin_PrintJobHistory, #tab_plugin_PrintJobHistory, ...
        elements: [
            document.getElementById("tab_printJobHistory"),
            document.getElementById("settings_printJobHistory"),
            document.getElementById("modal-dialogs-printJobHistory")
        ]
    });
```

In the [OctoPrint-Documentation](https://docs.octoprint.org/en/master/plugins/viewmodels.html#registering-custom-viewmodels) you could find this sentence:

> elements
> A list of UI elements to bind to. Each binding target can be either a string which will then be passed to jQuery’s $(...) method to resolve the target, or alternatively directly a jQuery element

jneilliii:
Ok, thanks printJobHistory is a plugin I can use to test against. This never came up for me because your implementation is rare in the plugins I've used/seen/reviewed. I guess I'll have to figure out how that is being passed over in the onAllBound callback.

jneilliii:
New version pushed that no longer errors with your plugin @OllisGit. Please let me know if it's working for you now.

OllisGit:
I tested the new version 0.0.6.

The first issue is now gone, but the second issue is still there. But as you mentioned, it works!
```Error calling onAfterTabChange on view model ConsolidatedtabsViewModel : Error: Invalid dimensions for plot, width = 0, height = 350```
This error is only gone if I add all tabs to the "Combined Tabs-Section".
Do you need more information to reproduce the error?

Sometimes I "lost" a tab after I reduced (or removed) the width, but after pressing the "Clear positioning" the tab is visible again.  

Also, there is a lot of noise in the browser console. Is that on purpose?

jneilliii:
0.0.7 released without the debugging noise. Hopefully that's good enough for you to approve and merge this plugin now.

OllisGit:
Okidoki, nice work!

jneilliii:
Thanks @OllisGit, If you have any ideas of how to get rid of all the whitespace that is generated at the bottom of the page I'm all ears. I haven't been able to find a solution for that.

OllisGit:
I saw that, but I am not a UI expert....my knowledge is "googling" for a solution ;-) Maybe you can find someone in the forum or on discord.

PR #563, opened 2020-07-26
Added plugin marlinbft

CharlesWillis3:


jneilliii:
The check failure I think might be related to an action issue, had the same thing happen to one of my recent submissions. Nice plugin, looking forward to testing it out. This is using the newest generation of binary transfer, because I know that was getting some updates?

CharlesWillis3:
@jneilliii This is using the version of binary transfer currently in Marlin bugfix-2.0.x. An update is planned for that feature, which will require updates to the implementation when it comes out.

foosel:
There's a bug in the image path validation of the PR check, see #565. Looking into it and fixing ASAP.

foosel:
Check failure has been cleared and #565 should be fixed. There were a bunch of warnings due to externally hosted images: https://github.com/OctoPrint/plugins.octoprint.org/actions/runs/185359331

CharlesWillis3:
@foosel thanks for the fix! I don't understand the externally hosted image warnings. Those image files are in the expected path and included as part of the PR

foosel:
Hm... then apparently something is still broken. I'll take a look in a bit. Sorry for the hiccups, validation is new and still needs some kinks ironed out, but will safe everyone involved a lot of hassle once that's happened.

foosel:
That's fixed now as well.

PR #569, opened 2020-07-31
Python 3 compatibility to OctoPrint-MyStromSwitch and OctoPrint-PrinterAlerts

cp2004:
2 more down, 80ish left to go?

Linked issues/PR:
OctoPrint-MyStromSwitch 
https://github.com/da4id/OctoPrint-MyStromSwitch/issues/25

OctoPrint-PrinterAlerts
https://github.com/Patronics/OctoPrint-PrinterAlerts/issues/6
https://github.com/Patronics/OctoPrint-PrinterAlerts/pull/7

PR #570, opened 2020-07-31
Added plugin VirtualPrinterSettings

cp2004:
I'm enjoying life on the plugin repo too much... 

**Please don't merge until 1.4.1 stable is out,** but I had some time so I finished up this plugin and wanted to get ahead, and move on.

@jneilliii Thanks so much for your help on this one!

Similarly to the Py3 compat PR (#569), this has the same ghost commit from the rebase, as they are from two different branches. But hopefully it will be gone after that....


foosel:
Any reason why I shouldn't merge that yet? You have set the OctoPrint compatibility so this will only be shown as compatible to 1.4.1 anyhow, so I don't see any harm in merging it now and maybe given some of the RCs users a chance to play around with it...

cp2004:
Ok lets go for it... The OctoPrint version check should stop most installing it and coming my way to say it doesn't work. And there's a note in the listing.

foosel:
Haven't yet been able to do a proper review, but two things I noticed:

  * featured image's URL is broken (no idea why that wasn't detected by the build *sigh*), plugin id in path is wrong
  * the plugin id is a bit... confusing ^^ It's not a new version of the virtual printer, it's a settings page for the existing one. Might be a good idea to have the id reflect that too (as the name does) so we don't confuse people here.

cp2004:
@foosel I agree on the I'd part - initially I had it set as `virtual_printer_settings` but that caused conflicts with the original settings page, so @jneilliii renamed it to virtual_printer2. Will look at something that is more clear. Will let you know once I've changed it.

cp2004:
@foosel Is `vp_settings` ok for an ID? That's what I've changed it too, as I think that closely mirrors the name, while not interfering with the existing virtual printer.

kantlivelong:
Just poking my head in. I'd love to see this merged into the core plugin
over a separate one.

cp2004:
I'm going to close this, and open a new PR to avoid commit clutter..... I've messed up the IDs again

@kantlivelong Gina said she didn't want it in core for now, as she didn't want to maintain it, and it was too late for 1.4.1. It's a plugin for now, but we'll have to see if we can sneak it in at some point :)

PR #571, opened 2020-07-31
Add plugin Virtual Printer Settings

cp2004:
It's there under id `vp_settings` now. Created new PR as the other one would have meant commit clutter, and I decided this was easier. 👍


cp2004:
As mentioned on discord, there is an issue with the replaces binding on the plugin. Seems it's hard to get the ID to not conflict and render the page twice

foosel:
Sorry, I was a bit busy with the release stuff... is that supposed to get merged now or not yet?

cp2004:
Needs to be tested, as jneilliii reported the replaces binding was broken again. On it at the moment, will get back to you shortly

cp2004:
@foosel Confirmed fixed, ready for the plugin repo!

foosel:
Did you intentionally not rename the plugin package?

cp2004:
Not intentionally, does it need to be renamed? I suppose it brings everything in line

foosel:
Would certainly be cleaner and avoid conflicts should someone down the line go with the same package name, fitting the now not claimed id. 

I'm sorry that this has been such a hassle, I'll look into the bug you mentioned on Discord ASAP so that future efforts like this are less of an id juggle nightmare ;)

cp2004:
Ok, have renamed it. In the process I think I pinned down the bug - If there is a template with name the same as the ID of the plugin, it gets rendered in the main web interface too. And maybe if you collide with another ID. Not got time to look into it further, but there's varying degrees of examples in the commit history of my plugin now!

PR #589, opened 2020-08-20
Procastinator plugin

juergenpabel:
Ever wanted to start a printjob sometime later? Instead of scheduliing a printjob by other means
(cron & curl via API), this plugin allows for a printjob to be started and immediately put on hold
until a given time (or immediately, if so desired). 

This might be useful if you want to start a printjob in the middle of the night (while sleeping)
or even just a little later on (without the risk of forgetting to actually start the printjob later on).

The plugin is configurable as to:
1. a time window during which a printjob is put on hold (like only in the evening)
2. 5 options as for times when to actually start the printjob

If a printjob is started during the configured time window then it is immediately put on hold and the
user is shown a dialog for selecting the time at which the job is to be continued (the first option is
always "NOW", just in case).


_plugins/procastinator.md
```diff
@@ -0,0 +1,65 @@
...
+layout: plugin
+
+id: procastinator
+title: Procastinator
+description: Puts printjobs on hold until a certain time
+author: Juergen Pabel
+license: AGPLv2
+
+# today's date in format YYYY-MM-DD, e.g.
+date: 2015-06-22
```
jneilliii, line 11: date needs to be corrected.

PR #601, opened 2020-09-02
Update listing for WS281x LED Status

cp2004:
Updated my plugins' listings to reflect recent updates.

Also corrected the name on Virtual Printer Settings, since it was pointing to `Charlie` instead of `Charlie Powell`


cp2004:
That was quick! Thank you 🙂

PR #603, opened 2020-09-05
Added plugin auth_ldap

battis:


jneilliii:
Awaiting clarification to linked PR to original repo versus fork to find out which one should be registered in the official plugin repo.

battis:
Question that may render which fork is posted moot: this plugin depends on python-ldap, which in turn [requires its own set of non-Python dependencies](https://github.com/battis/OctoPrint-LDAP#prerequisites)... which functionally means, I now realize, that simply installing from the repository is a non-starter, since those dependencies would need to be previously installed to successfully install the plugin.

Thoughts? Perhaps my big push to make my life as a lab manager easier by having this plugin in the repo was always doomed...

jneilliii:
That's definitely a sticking point. It's typically frowned upon to do install scripts during initialization that install external app dependencies. Typically, it's enough to clarify pre-installation steps on the plugin listing to ensure those are done ahead of time, but that doesn't help your situation one bit. @foosel do you have any ideas that would be acceptable for the plugin repo rules?

cp2004:
How about a setup wizard on install?

The only acceptable way to run system commands is if the user presses something that says 'run this'. I run it in my plugin WS281X LED Status, as it needs to change some OS level config to work. Especially if you need the password to run apt.

foosel:
> @foosel do you have any ideas that would be acceptable for the plugin repo rules?

Setup wizard as @cp2004 pointed out. Can be as simple as detecting that the dependencies are not met and then telling them user what to do about that manually (`apt install ...`)

battis:
@cp2004 @foosel Cool. I'll get a first pass done. Probably this weekend. Good idea. K.I.S.S.

foosel:
What's the status on this?

Issue #599, opened 2020-08-29
[Abandoned Plugin] OctoPrint Klipper Plugin


AliceGrey:
**Which Plugin? (URL to plugin page on plugin repository)**
https://github.com/mmone/OctoprintKlipperPlugin

**Ticket on plugin's issue tracker confirming abandonment:**
Last release was 2 years ago
https://github.com/mmone/OctoprintKlipperPlugin/releases/tag/0.2.5

**Is there someone open to adopting it? ("no" or @-mention user):**
I am open to adopting it

jneilliii:
@AliceGrey you linked to the releases, which is not necessarily an indication of abandonment, although based on the open issue count and pull requests it does in fact appear to be abandoned.

@mmone it appears there are several pull requests against your repo that will fix several issues, etc. Are you no longer actively developing this plugin? If you are not or we don't hear back from you within 14 days we will transition this plugin to @AliceGrey's fork and update the official plugin repo accordingly.

@AliceGrey relative to that transition, I'm curious if you merged in all those Pull Requests that fix other issues in your fork?

AliceGrey:
@jneilliii 
I pulled in the PR's except for https://github.com/mmone/OctoprintKlipperPlugin/pull/64
I have yet to review it fully. The other two PR's resolved the issues mentioned in the original PR's and have been merged. I also have confirmed in a local dev environment that the plugin runs on python 3 after some minor tweaks.

jneilliii:
Perfect, if we don't hear back from @mmone we'll make the switch.

jneilliii:
Just one final follow-up @mmone, if we don't hear back from you in the next 7 days we will transition this official listing to @AliceGrey's maintained fork.

jneilliii:
@AliceGrey ok, all I need from you now that we've waited for the proper number of days is a PR for the plugin repo md page with corrected URLs to your repo, etc. Once I've gotten that and reviewed we'll submit the announcement for people to manually reinstall to get them back in line for future software update checks.

cowboy3d:
Just saw this activity... after having created a pull request against my repository. @AliceGrey: Have a look at https://github.com/cowboy3d/OctoprintKlipper3Plugin if you're interested to see what I have changed to make it work with python 3.
Thank you for taking this over and good luck!

GitIssueBot:
This issue has been mentioned on **OctoPrint Community Forum**. There might be relevant details there:

[...]

PR #613, opened 2020-09-12
Added plugin klipper_python3.

cowboy3d:
I did some changes to allow the existing klipper plugin to also work with pyton 3.
The original author didn't react to some requests to have it ported to python 3 therefore I decided to fork it.


jneilliii:
@cowboy3d someone has already adopted this plugin and it will be merged soon. See abandonment ticket below. 

https://github.com/OctoPrint/plugins.octoprint.org/issues/599

PR #620, opened 2020-09-19
OctoPrint-BackupScheduler

jneilliii:
Adds a rudimentary scheduler to run backups at configured intervals.


jneilliii:
apparently I still haven't figure out how to re-base properly...sorry for the extra commits.

PR #629, opened 2020-09-22
Added plugin terminalmessaging. 

jeffeb3:
Added terminalmessaging.

I didn't test with Jekyll. Sorry, I just shouldn't install a bunch of gems on this machine. I can try on another machine later if you want.


_plugins/terminalmessaging.md
```diff
@@ -0,0 +1,105 @@
...
+  # Compatible Python version
+  #
+  # Plugins should aim for compatibility for Python 2 and 3 for now, in which case the value should be ">=2.7,<4".
+  #
+  # Plugins that only wish to support Python 3 should set it to ">=3,<4". 
+  #
+  # If your plugin only supports Python 2 (worst case, not recommended for newly developed plugins since Python 2
+  # is EOL), leave at ">=2.7,<3"
+
+  python: ">=2.7,<3"
```
jneilliii, line 76: we are no longer accepting plugins that don't run in python 3. Looking at the plugin source it appears to be python3 compatible, so this just needs to be updated to the following.

```
python: ">=2.7,<4"
```

_plugins/terminalmessaging.md
```diff
@@ -0,0 +1,105 @@
...
+This is similar to how a text messaging app would differentiate between sender and receiver in the
+UI.
+
+![Screenshot](/assets/img/plugins/terminalmessaging/screenshot.png)
+
+This plugin is very small, and just edits the css of the elements in the terminal for readability.
+
+Enjoy!
+
+## Setup
```
jneilliii, line 96: I would probably remove this bit for the plugin repo...

jneilliii (changes requested):
@jeffeb3 thanks for the shout out on this plugin. A couple of minor thing need to be corrected before approval and merge, please see the comments.

In addition to those changes, please create an initial [Release](https://github.com/jeffeb3/OctoPrint-TerminalMessaging/releases) on the teminalmessaging plugin for 0.0.1, so it is there as reference for future updates.

_plugins/terminalmessaging.md
jneilliii: I meant specifically the whole Setup section....

_plugins/terminalmessaging.md
jeffeb3: Can I just remove the whole compatability section? I saw conflicting advice (the template says I don't need it and the docs say I need to be python3 compatible).

jneilliii: Yeah, that's an option too and it will get extracted from the init file. I still tend to manually add it anyway.

jneilliii (approved):

PR #647, opened 2020-10-02
Adds plugin MakerOS

makeros3d:
Hi there, MakerOS has developed an official plugin and would like to add it to the plugin repository.

Please let me know if there's anything that needs to change so I can get that done asap.

Thanks so much and best always!


jneilliii:
@makeros3d please update the description of your plugin to notify the end users that this is a commercial product that it connects to and that there is not free tier available (that I could find). Once that is done we'll have no issue in merging in this PR.

cp2004:
Be warned as well that you will get lots of Python 2 users installing this and asking questions about why it doesn't work. The migration is underway but still only in early stages 🙂

sofubi:
Hey there, thanks so much for the quick response, it's greatly appreciated.
Please find that the description has been updated to include a link and information about the fact that this requires a subscription.
Thank you for the heads up on Python 2. That is something that we plan to build into our coming releases.

makeros3d:
Hi there, thanks for the quick response, it's greatly appreciated!
Please find that the description has been updated to include the necessary information and a link.
Thank you for letting me know about Python 2, that is something that will be built into our future releases.

jneilliii (approved):

PR #658, opened 2020-10-10
Added plugin raisecloud

Raise3D:
Added plugin raisecloud


jneilliii:
@Raise3D I noticed the header in RaiseCloud says `waive printer count and membership limits`. Your submission here in the official plugin repo should include information related to the standard limits and if there are free tiers available to the community, etc.

jneilliii (approved):

PR #671, opened 2020-10-28
Multiple authors for one plugin

j7126:
Adds a new authors property that can be used in any plugin .md file. 

This allows multiple authors for one plugin to be linked on the plugin page and displayed separately in the 'by author' section

image

image

This PR additionally adds this property for the dashboard plugin.

### DIFF
--- a/.github/scripts/validate_front_matter.py
+++ b/.github/scripts/validate_front_matter.py
@@ -58,6 +58,7 @@ SCHEMA = Schema({
 	Required("title"): NonEmptyString,
 	Required("description"): NonEmptyString,
 	Required("author"): NonEmptyString,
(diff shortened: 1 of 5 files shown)


j7126:
This PR will fail build checks because it adds a new property. 7f106f6 (in this PR) updates the build checks to allow this property

foosel:
Merged, after some changes... It's enough now to supply `authors`, I've also adjusted the help (and the cookiecutter template) accordingly. The templates will make sure that multiple authors get comma-separated in the generated exports that only support one string with all authors. 

Also extended the frontmatter validator accordingly.

*(this commented got added way later than planned due to a short internet outage just when I wanted to post it)*

PR #673, opened 2020-10-30
Continuous printing change repositories

Zinc-OS:


jneilliii (changes requested):
Couple of things @Zinc-OS . Don't you want to update the author on the repo in this PR? Also one small problem on your repo [here](https://github.com/Zinc-OS/continuousprint/blob/ab4b22a1c3ff05a3de53e2bdb0dcde0628c244ae/continuousprint/__init__.py#L238). The user needs to be updated there too.

Zinc-OS:
Ok. All fixed. What is the date for though? It threw an error when I changed that to today.

cp2004:
Registration date should not be changed from initial registration of the plugin - That date is used inside the plugin manager, announcements of new plugins, and helps to order the home page of https://plugins.octoprint.org

jneilliii (approved):

PR #676, opened 2020-11-06
More Python 3 compatibility

cp2004:
More Python 3 compatibility PRs in the last couple days - up to 76% now. Only one technically needed to be done since I think the rest would be automatically picked up, but they're here just in case that fails 🙂 

* OctoPrint Macro - https://github.com/mike1pol/octoprint_macro/pull/14
* OctoPrint TouchTest - https://github.com/Peaches491/OctoPrint-Touchtest/pull/26
* OctoPrint IFTTT - https://github.com/tjjfvi/OctoPrint-IFTTT/pull/35

### DIFF
--- a/_plugins/IFTTT.md
+++ b/_plugins/IFTTT.md
@@ -25,6 +25,9 @@ screenshots:
   alt: OctoPrint-IFTTT Settings
   caption: OctoPrint-IFTTT Settings
 
+compatibility:
+  python: ">=2.7,<4"
+  
 ---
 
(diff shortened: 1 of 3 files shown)


jneilliii:
ifttt and octoprint_macro need to make official version releases on their repo. but I will go ahead and merge.

cp2004:
I think that should be OK if they don't, since irrc the plugin manager installs using the zip of the branch, rather than the latest release - so these changes *should* be included should anyone press install *from now*

jneilliii:
yep, initial install does come from the main target branch zip in most cases.

PR #688, opened 2020-11-18
Abandoned Filament Sensor Reloaded

ssorgatem:


foosel (approved):

nickmitchko:
@ssorgatem I'd be willing to take over the plugin, I am a daily driver of it and it provides a rather simple interface which I like

PR #690, opened 2020-11-19
support python 3 using octoprint >= 1.4 

ChristophSchranz:


_plugins/preprintservice.md
```diff
@@ -31,12 +31,15 @@ compatibility:
 
   octoprint:
   - 1.3.0
+  - 1.4.0
```
jneilliii, line 34: @ChristophSchranz I don't this this is going to do what you expect and I don't believe is necessary. If they are running older versions of OctoPrint then they will be utilizing the python 2.7 side of things, which you are tagging as compatible. When there is only 1 version here in this list, then it is any version newer than that listed version will be supported. The python compatibility flag below should be enough. If there are more than one versions listed here than only those versions will be able to install the plugin and run.

I think you should just remove the `- 1.4.0` part.

jneilliii (changes requested):
@ChristophSchranz please see comments.

_plugins/preprintservice.md
ChristophSchranz: Did it, thanks!

jneilliii (approved):

PR #693, opened 2020-11-29
Added Plugin Gcode Leveling

Willmac16:


jneilliii:
Your dictionary returned [here](https://github.com/Willmac16/OctoPrint-GcodeLeveling/blob/9cddb5047522475d39e23cccc621997d0fe3c2b7/octoprint_GcodeLeveling/__init__.py#L264) should match your plugin's identifier, case-sensitive.

jneilliii:
Also, we encourage you to include screenshots on the plugin repository to give the user's an idea of what the plugin might contain, but that is not mandatory.

jneilliii:
One other thing, I know from experience with the Bed Level Visualizer plugin you might want to add a warning on the plugin repo in regard to the length of time it takes numpy to compile in python 2.7 environment, might save you some headaches in the long run.

Willmac16:
Should I add an enable button in the settings in case someone installs the plugin and doesn't configure it? or can I assume that people will only install the plugin if they will properly configure it?

jneilliii:
> Should I add an enable button in the settings in case someone installs the plugin and doesn't configure it? or can I assume that people will only install the plugin if they will properly configure it?

That's really up to you.  I would think that only people that are going to use it/configure it will install it. Plugin Manager also has the standard enable/disable functionality in it in case the plugin just needs to be disabled.

jneilliii (approved):

PR #701, opened 2020-12-05
Added plugin webcamtab_g

gruvin:
Forked original author's (incomunicado) version and added support for OctoPrint versiosn 1.4 and 1.5, as well as python3.

### DIFF
--- a/_plugins/webcamtab_g.md
+++ b/_plugins/webcamtab_g.md
@@ -1,0 +1,27 @@
+---
+layout: plugin
+
+id: webcamtab_g
+title: WebcamTab
+description: Moves the webcam stream from the controls into its own tab
+authors: 
+- Sven Lohrmann
+- Bryan J. Rentoul
+license: AGPLv3
+date: 2020-12-06
+homepage: https://github.com/gruvin/OctoPrint-WebcamTab
+source: https://github.com/gruvin/OctoPrint-WebcamTab
+archive: https://github.com/gruvin/OctoPrint-WebcamTab/archive/master.zip
+compatibility:
+  octoprint:
+  - 1.3.5
+  - 1.4.0
+  - 1.5.0
+tags:
+- webcam
+- ui
+---
+
+This OctoPrint plugin moves the webcam stream from the controls into its own tab. Keyboard control is still possible.
+Now supports python3 as well as OctoPrint versions 1.3.5, 1.4.x and 1.5.x.
+


jneilliii:
If the author has become "incomunicado" as you say, I think this would be better to do as a plugin adoption rather than a whole new plugin with overlapping functionality just for compatibility reasons.

jneilliii:
A little more detail can be found in the [Abandoned plugin](https://github.com/OctoPrint/plugins.octoprint.org/issues/new?labels%5B%5D=abandoned+plugin&template=abandoned_plugin.md&title=%5BAbandoned+Plugin%5D+%3Cinsert+name+of+plugin+here%3E) issue template.

gruvin:
The template for _plugins/plugin_name.md at https://plugins.octoprint.org/help/registering/ appears to be incongruent with reality. 

```
# Compatible Python version
  #
  # Plugins should aim for compatibility for Python 2 and 3 for now, in which case the value should be ">=2.7,<4".
  #
  # Plugins that only wish to support Python 3 should set it to ">=3,<4".
  #
  # If your plugin only supports Python 2 (worst case, not recommended for newly developed plugins since Python 2
  # is EOL), leave at ">=2.7,<3"

  python: ">=2.7,<3"
```

 Specifically ...

```
python: ">=2.7,<3"
```

... is rejected as a disallowed extra field.

gruvin:
@jneilliii understood. Didn't know that wa an option. Will address when I find some more time. Thanks.

jneilliii:
> The template for `_plugins/plugin_name.md` at https://plugins.octoprint.org/help/registering/ appears to be incongruent with reality.
> 
> ```
> # Compatible Python version
>   #
>   # Plugins should aim for compatibility for Python 2 and 3 for now, in which case the value should be ">=2.7,<4".
>   #
>   # Plugins that only wish to support Python 3 should set it to ">=3,<4".
>   #
>   # If your plugin only supports Python 2 (worst case, not recommended for newly developed plugins since Python 2
>   # is EOL), leave at ">=2.7,<3"
> 
>   python: ">=2.7,<3"
> ```
> 
> Specifically ...
> 
> ```
> python: ">=2.7,<3"
> ```
> 
> ... is rejected as a disallowed extra field.

I just noticed this comment, and actually it's even more out of date then that. If the plugin resides on a github repo the github actions will actually attempt to extract the python compatibility information from the plugin's repo if the value is not there at all, but since that only works with github hosted plugins (I think) the documentation is still valid. Were you getting an error because your python line wasn't indented under compatibility maybe? If you remove the comments it's a little easier to see.

```
compatibility:
  octoprint:
  - 1.3.0
  os:
  - linux
  - windows
  - macos
  - freebsd
  python: ">=2.7,<3"
```

gruvin:
> ... because your python line wasn't indented under compatibility maybe?
Almost certainly. :-)

I guess it's time to close this PR and re-do it, once whatever has to happen with the plugin adoption process has happened ... and whatever else I then need to do.

PR #707, opened 2020-12-11
Add OctoPrint-CustomCSS plugin

crankeye:
A simple plugin for adding your own custom CSS snippets easily.

### DIFF
--- a/_plugins/customcss.md
+++ b/_plugins/customcss.md
@@ -1,0 +1,45 @@
+---
+layout: plugin
+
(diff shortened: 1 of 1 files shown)


jneilliii (approved):

jneilliii:
Interesting starting point, since that is my plugin (Thumbnails). There are settings in the plugin to adjust the size, but that would do it for all interfaces and not just TouchUI. Good plugin submission.

PR #714, opened 2020-12-17
Added plugin display_panel

sethvoltz:
Simple control and status paired with a physical button panel and OLED display


_plugins/display_panel.md
```diff
@@ -0,0 +1,75 @@
...
+
+  python: ">=3,<4"
+
+---
+
+This plugin implements the software control side of an OctoPrint Control Panel for Octopi. The hardware half is a series of 4 buttons, an OLED screen and a 3D printed case that mounts on the printer next to the Raspnerry Pi so it can be plugged in to the header pins.
+
+**Watch a demo of the panel in action!** More photos are at the end of this readme.
+
+[![Screenshot of demo video](https://img.youtube.com/vi/78emT1ollu4/0.jpg)](https://youtu.be/78emT1ollu4 "Click here to watch a demo on YouTube")
```
jneilliii, line 51: needs to be replaced with youtube embed token. See the `Do not directly embed third party screenshots/widgets/iframes!` section on the registering instructions page [here](https://plugins.octoprint.org/help/registering/) at the end of the 3rd numbered item.

_plugins/display_panel.md
```diff
@@ -0,0 +1,75 @@
...
+
+Please see the [plugin repository](https://github.com/sethvoltz/OctoPrint-DisplayPanel) for the latest hardware and wiring instructions. The wiring instructions at the time of publishing are below.
+
+![Wiring diagram for the control panel](/assets/img/display_panel/wiring-diagram.png)
+
+## Setup
+
+**NOTE:** This plugin required OctoPrint to be updated to run on Python 3. Please follow [these instructions](https://community.octoprint.org/t/upgrade-your-octoprint-install-to-python-3/23973) if you are not already on Python 3.
+
+Before installation, you will need to `pip` install a few dependencies. Please ensure you are using the `virtualenv` bundled version:
```
jneilliii, line 65: Rather than having the user manually install these module dependencies I would recommend you add the required module references [here](https://github.com/sethvoltz/OctoPrint-DisplayPanel/blob/3c7b0555548603b894d7e0751a8f8b483230005b/setup.py#L36) in your plugin. You'll need to make sure to define all required modules, ie `adafruit-circuitpython-ssd1306` and `pillow` etc. That section is an array of strings similar to what you would put in requirements.txt and will allow the plugin manager in OctoPrint to automatically install those for you. Might help reduce your support overhead with questions about why it won't load.

jneilliii (changes requested):
Please see my comments after review. Other than those couple of items I see no reason this can't be merged. Thanks.

_plugins/display_panel.md
sethvoltz: Whoops, I definitely saw the part about embedding in the plugin and moved to the next bullet, missing the part about the description. Updating...

sethvoltz: I opted to bring the cover photo in as a local asset to avoid any remote asset or YouTube iFrame issues.

jneilliii (approved):

PR #726, opened 2021-01-03
Added plugin LongPathSDList

Petttkous:


jneilliii (commented):
@Petttkous I see nothing wrong with this plugin submission, but have one recommendation. Although the plugin is extremely simple, I think you might still want a software update hook in case you do need to release a new version to fix bugs in the future. This will prevent your users from having to uninstall and reinstall the plugin. 

https://docs.octoprint.org/en/master/bundledplugins/softwareupdate.html#sec-bundledplugins-softwareupdate-hooks-check-config

Petttkous:
Okay, I have added the update hook.

jneilliii:
> Okay, I have added the update hook.

Thanks, one minor issue. The dictionary returned needs to be the same as the plugin_identifier, so the line [here](https://github.com/Petttkous/OctoPrint-Longpathsdlist/blob/396dae7edf1d0709c9588c015c4ad0872e5d6e1c/octoprint_LongPathSDList/__init__.py#L20) needs to be 

```
LongSdPaths=dict(
```

foosel:
I would also strongly suggest to place a big warning into the description that this plugin could lock up the communication indeterminately depending on the contents of the SD card in the printer.

jneilliii:
Oh, I didn't no that @foosel. Is that a known bug with the M33 command? I don't use SD cards, so was unaware.

cp2004:
It's not a bug, but since it is requesting the name for *every* file, if you have 100 files it will promptly pause serial comm while it goes back and forth to query each one. If you have more than that...

jneilliii:
ahh, that makes sense.

Petttkous:
I have corrected the identifier and also added warnings to the plugin description.

jneilliii (approved):

PR #728, opened 2021-01-09
Added MeatPack plugin

scottmudge:
Adding my new MeatPack g-code compression plugin to the official repository.


jneilliii (changes requested):
I think the software update check stuff is case sensitive, so you need to make sure your username [here](https://github.com/scottmudge/OctoPrint-MeatPack/blob/ff0c7bd642f05ed859566fb3606d569f39076ddf/OctoPrint_MeatPack/__init__.py#L142) and the link [here](https://github.com/scottmudge/OctoPrint-MeatPack/blob/ff0c7bd642f05ed859566fb3606d569f39076ddf/OctoPrint_MeatPack/__init__.py#L147) is all lower, based on your github info.

scottmudge:
Thanks for spotting that, it's been corrected.

jneilliii (approved):
Out of an abundance of caution I'm going to have to request @foosel check out this plugin overall since it's using a serial factory. This is something that I've never worked with and she'll be able to verify that this couldn't possibly cause problems or not. Everything else looks ok to me.

scottmudge:
@foosel - Given that Marlin has now integrated this feature firmware-side into their official repo, how far along is the review process?

It would be nice for people not to have to install/update the plugin by logging into their OctoPrint installations via SSH/terminal and using pip.

foosel:
I gotta admit that the request for my review got completely lost (which means I'll need to look into some kind of automatic reminder). I'll try to get it done ASAP.

People don't need to fool around with SSH however to install a plugin, they can just copy-paste the install URL in their plugin manager right from the web interface.

scottmudge:
Thanks! And good to know! I guess I never attempted to try manual installation that route.

foosel (approved):

foosel:
That was straight forward enough, should be live once the page build has run through in a few minutes.

foosel:
Also, let me just say, this is exciting! I've been thinking here and there about something like this but never found the time to do some actual brainstorming and experimenting, so it's absolutely wonderful to see this implemented :D

scottmudge:
Great, thanks! And for sure, hopefully it will help some people as it did myself. At the very least it should free up some resources/interrupt time for other processes.

scottmudge:
Quick question -- do I need to do a new PR if I change the plugin description in the plugins library fork?

jneilliii:
You only need to submit a PR if you need to change anything that shows up on plugins.octoprint.org for your plugin. That's what this template is used for, to build those web page.

scottmudge:
Thanks, I'll have to submit a new PR to fix the description to reflect that Marlin now supports it.

PR #732, opened 2021-01-14
OctoPrint-Slic3r supports python3

eyal0:


eyal0:
Seems like python >2.7,<4 is the default anyway so I don't even know if it matters if this is merged.\

Is that true?

cp2004:
No, default is still Py2, but there's a clever script that checks the compatibility string in your plugin's code, so if that changes then the plugin repository has done as well.

cp2004 (approved):

eyal0:
> No, default is still Py2, but there's a clever script that checks the compatibility string in your plugin's code, so if that changes then the plugin repository has done as well.

That's pretty cool.  So the script that uploads this repo to the serve also visits each plugin and gets the compatibility info from the setup.py or whatever?  Neat.

foosel:
Yep.. I also wrote a blog post about the automation on the plugin repo that takes care of stuff like that: 

https://octoprint.org/blog/2020/12/15/turning-a-static-website-more-dynamic/

PR #736, opened 2021-01-18
Added plugin OctoBuddy

mlo821:
Adding OctoBuddy plugin to repo


jneilliii (changes requested):
Couple of things. First please remove your vs code folder from the PR. That type of stuff should not be carried over to the plugin repo. Second, it seems that you are missing a software update hook in your plugin. This will prevent users from being able to benefit from the software update checks. We highly recommend that all plugin registered with the official plugin have that so managing bug fixes from your repo directly are possible.  You can find additional information about this hook [here](https://docs.octoprint.org/en/master/bundledplugins/softwareupdate.html#sec-bundledplugins-softwareupdate-hooks-check-config).

mlo821:
@jneilliii requested updates have been made to the PR.  Thanks for the heads up on the software update hook, that will come in useful :)

jneilliii (changes requested):
@mlo821 I apologize I missed this before, but one more change on your repo side. The line [here](https://github.com/mlo821/OctoBuddy/blob/44e3a93cf0b5ea5ddccfe00323189a28d6620292/setup.py#L36) needs to include your dependencies (Rpi.GPIO). So the line should be something like this...

```
plugin_requires = ["RPi.GPIO"]
```

mlo821:
@jneilliii  no worries, repo has been updated with that change

jneilliii (approved):

PR #738, opened 2021-01-19
Take over Slicer development

eyal0:


_plugins/slicer.md
```diff
@@ -4,15 +4,15 @@ layout: plugin
 id: slicer
 title: Full-featured Slicer
 description: A full-blown GUI-based slicer. Rotate and scale model; slice multiple STL files at a time; set layer height and other slicing settings.
-author: Kenneth Jiang
+author: eyal0
```
jneilliii, line 7: rather than author, shouldn't we make this 
```
authors:
- eyal0
- Kenneth Jiang

jneilliii (commented):
@eyal0 I have reached out to @kennethjiang on Discord to see if he would be willing to transfer this repo to your ownership rather than changing over to your fork. This is a little cleaner when it comes to updates and such as we don't have to use the overlays. If I haven't heard back in a couple of days we'll go the overlay route.

Aside from that, I think you should give credit to the original author the way that we have with other plugins, per my code comment.

kennethjiang:
@eyal0 I'm glad that you offered to maintain and further developer the slicer plugin. I have been too busy on TSD to spare any time on the attention this plugin deserves.

I'll be fine with listing @eyal0 as the sole author, or both of us.

_plugins/slicer.md
eyal0: Fine by me.  I didn't know that a list was possible there.

kennethjiang:
@eyal0 I have also added you to my original repo. You think it's a good idea to still have the plugin pointing to my original repo, since the stars accumulated in that repo would give the plugin some extra boost in user's confidence. ;)

eyal0:
> @eyal0 I have also added you to my original repo as a collaborator. You think it's a good idea to still have the plugin pointing to my original repo, since the stars accumulated in that repo would give the plugin some extra boost in user's confidence. ;)

That seems good to me, too.  @jneilliii What do you think?  I'll continue the work there and we'll drop this PR?

eyal0:
Also it means that I won't have to copy over all the bug reports.  That's an even better reason not to switch.

eyal0:
I modified this PR.  Please take a look again.

kennethjiang:
lgtm. The only thing I'd say is to put you as the 1st author as you will be actively working on it. :) @eyal0

jneilliii:
Yeah, the order is up to you guys. I could see it both ways, easier to add the new maintainer to the end of the list later down the road, or have current first...your choice. I think for the other plugins that have been adopted it has been new maintainer first if you want to just keep consistency.

jneilliii (approved):

PR #742, opened 2021-01-22
MarlinGcodeDocumentation: Klipper support

costas-basdekis:


jneilliii (changes requested):
All of your screenshot URLs will be broken. You need to make sure to include the pull relative path, so put the `/assets/img/plugins/marlingcodedocumentation/` back in.

costas-basdekis:
Thanks good catch, fixed @jneilliii

jneilliii (approved):

PR #750, opened 2021-01-26
Plugin meatpack - update description w/ less dynamic information.

scottmudge:
Updated description with more general description and a link to an external list of supported firmware. Also updated compression ratio with latest numbers.

### DIFF
--- a/_plugins/meatpack.md
+++ b/_plugins/meatpack.md
@@ -47,7 +47,6 @@ It compresses/packs outgoing g-code by a ratio of 0.55 (or 0.61 without whitespa
 
 ### Firmware with MeatPack Support:
 
-* Marlin (merged Jan. 25, 2021, so check that your release was built after this date)
-* Prusa (official build reviewing support, unofficial build with support available at https://github.com/scottmudge/Prusa-Firmware-MeatPack)
+To check if your firmware has MeatPack support, please check here: https://github.com/scottmudge/OctoPrint-MeatPack#firmware-with-meatpack-support
 
 Adding support is fairly straight-forward, so if you'd like to add support into other firmware, please contact me and I can provide the latest firmware-side source code (also available in Prusa fork above).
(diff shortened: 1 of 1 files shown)


jneilliii (approved):

scottmudge:
@jneilliii Thanks, one last question -- this line in the plugin's descriptor/md file:

`python: ">=2.7,<4"`

This plugin is only compatible with python >= 3.7. Should I change it to:

`python: ">=3.7,<4"` 

?

scottmudge:
Nevermind, I'll just make it more compatible. Makes the most sense.

cp2004:
They should match, so that people know before installing. It can be quite a lot of effort to go back to Python 2 - I wouldn't recommend it. OctoPi 0.18 is out with Python 3, over 15% of installs (and rising quicker now, watch https://data.octoprint.org/).

jneilliii:
I agree, force it to python 3 only and just submit a PR to update here.

foosel:
Just to chime in: yep, focus on py3. I wish I didn't have to still support 2, be glad if you don't ;)

scottmudge:
Alright, so I made multiple notes in the release notes and README that it's highly recommended to update to Python 3, but given it only required minor syntax changes, it should in theory be compatible back to 2.7.

Not that I *want* to support it moving forward, but I figured most of the older systems which might see the most benefit from g-code compression are still running 2.7, and a lot of people (unfortunately) aren't going to bother manually upgrading to Python 3.

scottmudge:
One last question, how often does the plugin repository pull update information from the associated plugins' repos? If I publish a new release, will it take a few hours to be reflected in the Plugin Repository?

cp2004:
Every 12 hours, at midnight/midday UTC. Or the build also runs when a PR is merged.

scottmudge:
Great, good to know. Thanks!

PR #759, opened 2021-02-02
Update from 0.2.1 to 0.2.2

iFrostizz:


cp2004 (changes requested):
You still need all the metadata at the top of the page, for the plugin listing to function, please put this back.

The paths to the images are not valid, most of them don't exist and they must be in `/assets/img/plugins/EasyServo/....`, they should be uploaded here (not externally linked) and then the links changed to reflect where the images actually exist.

You can test the build of the plugin repository page using the instructions [on the plugin repository help link](https://plugins.octoprint.org/help/registering/), though Jekyll can be a massive pain to install on windows so I recommend using WSL or an actual Linux box if you need to build the page.

jneilliii:
I think I got this PR squared away now @cp2004 could you put one last set of eyes on it to verify please?

cp2004 (approved):

PR #769, opened 2021-02-08
Add "a" in the template for abandoned plugins

awenelo:
Minor change to the issue template for the abandoned plugins, adding an "a":
```
If you do not get a response on this within reasonable time
```
```
If you do not get a response on this within a reasonable time
```


foosel:
👍

PR #774, opened 2021-02-11
Add my robot control plugin to the repository

Zinc-OS:


FormerLurker:
Will take a look after work, thanks for the submission!

_plugins/robotcontrol.md
```diff
@@ -0,0 +1,56 @@
...
+
+tags:
+- i2c
+- automation
+- robot
+- robotics
+- GPIO
+
+
+featuredimage: /assets/img/plugins/robotcontrol/screnshat.png
```
FormerLurker, line 24: Is that a typo?  Not exactly an error though :)

FormerLurker (commented):
Just want you to have an opportunity to fix a typo first.  See my comment about your featured image.

_plugins/robotcontrol.md
Zinc-OS: will fix :)

FormerLurker: LOL!  I'm the king of typos, so no biggie.  I'm still reviewing the plugin code.  Will take a bit of time to complete.

FormerLurker:
The following is related to the actual plugin code in your repository.  First off, this is my first plugin review, so I am sort of in training :)  @jneilliii will verify that what I say here is accurate.

#  ```setup.py```
In [this line](https://github.com/Zinc-OS/octoprint-robot-plugin/blob/f0416f3be5eb154ba31ce08e263262738b4132ca/setup.py#L10) the instructions suggest the Plugin Package should be in the following form:  ```octoprint_<plugin identifier>```

However, your plugin did install, and I've found others that do not follow this form.  Probably fine, but I believe this is to prevent potential package name conflicts.  If you want to change it, you'd have to also change your root directory name to match the new plugin package.  I just wanted to mention it, but learned that i need to dig into the module import code, and don't understand it enough yet to say if this is good practice or not.

# ```__init__.py```

This isn't exactly a problem getting your plugin into the repo, but on [this line](https://github.com/Zinc-OS/octoprint-robot-plugin/blob/691828e6868be279be161999d0d5b083f36d5bbd/robotcontrol/__init__.py#L41) I don't think you meant to save your settings here.  Any reason you did this?
I could be mistaken.

I believe this [info logging](https://github.com/Zinc-OS/octoprint-robot-plugin/blob/691828e6868be279be161999d0d5b083f36d5bbd/robotcontrol/__init__.py#L75) should be altered, because it will log every queued gcode (say, while printing), which is probably going to impact performance.  I would recommend removing it completely, though you could also change it to debug if you feel it is necessary.

This is just an FYI, but [here](https://github.com/Zinc-OS/octoprint-robot-plugin/blob/691828e6868be279be161999d0d5b083f36d5bbd/robotcontrol/__init__.py#L100) you could just do self._logger.exception("message here") to get a stack trace and log the exception.  This is a very convenient way to log exceptions in general, and you can avoid having to fetch the info yourself.

I believe [this](https://github.com/Zinc-OS/octoprint-robot-plugin/blob/691828e6868be279be161999d0d5b083f36d5bbd/robotcontrol/__init__.py#L102) is probably unintended behavior.  I think you are trying to suppress commands that start with 'servo', no?  If so, 
```
return None, 
```
is probably what you want.  That will prevent the gcode from being sent.

Also, I see you have added ```@restricted_access``` to your api calls.  Did you intend for all signed in users, including read-only users, to be able to send api requests?  Take a look at @jneilliii's implementation [here](https://github.com/jneilliii/OctoPrint-TPLinkSmartplug/blob/dd3568665a182452c456a694a5ba2f00dc542d19/octoprint_tplinksmartplug/__init__.py#L1096-L1106) if you want to further restrict permissions to admins.  Personally, I have used ```with permissions.Permissions.ADMIN.require(http_exception=403): ``` at the very top of the blueprint methods, but there are other ways to accomplish this.  If allowing all logged in users access to your API is acceptable, changes aren't necessary.

# ```robotcontrol.js```

I don't believe you need to push the UI_API_KEY into the header [here](https://github.com/Zinc-OS/octoprint-robot-plugin/blob/f0416f3be5eb154ba31ce08e263262738b4132ca/robotcontrol/static/js/robotcontrol.js#L56-L70).  Try removing it from your ajax requests and see if they work.  I believe this is deprecated.  I ran a test through my browser to confirm, but you should give it a go in your source to verify.  No need to add stuff you don't need to the header :)

# summary
Looks pretty good to me!  No major changes are needed, just a few tweaks and some things I wanted to make you aware of, especially in the queuing hook.

_plugins/robotcontrol.md
```diff
@@ -0,0 +1,56 @@
...
+# Robot Control Plugin
+
+Plugin for controling my robot via the i2c bus.
+
+
+## Configuration
+
+Currently For a four servo robot(I plan to add support for more).
+
+You will also need to enable i2c bus via ```sudo raspi-cofig``` throught the terminal, accessible through ```ssh pi@octopi.local:22```
```
jneilliii, line 45: should be `through` the terminal?

jneilliii (changes requested):
Good review @FormerLurker I do agree with everything you pointed out, specifically the logging of the gcode hook. There is one typo also in my review here.

Zinc-OS:
The logging was for testing purposes

_plugins/robotcontrol.md
Zinc-OS: it should

Zinc-OS:
Thanks for the help @FormerLurker

FormerLurker:
Everything looks good now I think.  @jneilliii, can you verify, then I think this can be merged.

jneilliii (approved):

PR #777, opened 2021-02-14
Added plugin inlineconfirm

j7126:


_plugins/inlineconfirm.md
```diff
@@ -0,0 +1,35 @@
+---
+layout: plugin
+
+id: inlineconfirm
+title: OctoPrint-InlineConfirm
+description: Removes the confirm cancel print dialog and replaces it with confirmation by clicking the cancel buton twice.
+authors:
```
Zinc-OS, line 7: should be author, not authors

foosel: `authors` is valid too actually, though it requires a list

jneilliii (approved):

PR #779, opened 2021-02-16
Adding TpLinkAutoShutdown as a plugin

jamesmccannon02:
I am adding a plugin to allow the auto-shutdown of the printer when a print has completed using Kasa Smart-Plugs


jneilliii:
We highly recommend that plugins have the software update hooks, so that you can deliver bug fixes automatically. Please check the docs [here](https://docs.octoprint.org/en/master/bundledplugins/softwareupdate.html#octoprint-plugin-softwareupdate-check-config).

jamesmccannon02:
Ok, will have a look at implementing this.

jamesmccannon02:
@jneilliii The update hooks have now been added and i can confirm that it works when a new update has been released.

jneilliii (approved):

PR #787, opened 2021-02-19
Added plugin IgnorePausedForUser

DavideM84:
Added plugin IgnorePausedForUser


_plugins/ignorepausedforuser.md
```diff
@@ -0,0 +1,99 @@
...
+  # Compatible Python version
+  #
+  # Plugins should aim for compatibility for Python 2 and 3 for now, in which case the value should be ">=2.7,<4".
+  #
+  # Plugins that only wish to support Python 3 should set it to ">=3,<4".
+  #
+  # If your plugin only supports Python 2 (worst case, not recommended for newly developed plugins since Python 2
+  # is EOL), leave at ">=2.7,<3"
+
+  python: ">=2.7,<3"
```
jneilliii, line 87: we no longer accept plugins in the repo that are not python 3 compatible. Looking at your plugin it appears to be specifying python 3 compatible, so this should be `">=2.7,<4"`.

_plugins/ignorepausedforuser.md
```diff
@@ -0,0 +1,99 @@
...
+- Davide Mencarelli <d.mencarelli@outlook.com>
+#- second autor name
+license: GNU v3.0
+
+# today's date in format YYYY-MM-DD, e.g.
+date: 2021-02-19
+
+homepage: https://github.com/DavideM84/OctoPrint-IgnorePausedForUser
+source: https://github.com/DavideM84/OctoPrint-IgnorePausedForUser
+archive: https://github.com/DavideM84/OctoPrint-IgnorePausedForUser/archive/v1.2.2.zip
```
jneilliii, line 17: this shouldn't point to a specific release version, but the asset download url for a branch or latest version tag.

cp2004:
We aren't accepting new plugins that aren't Python 3 compatible - please either make it Py 3 compatible or correct the listing if it is in fact Python 3 compatible.

jneilliii (commented):
Concerned a little about the writing of events to the log with info rather than debug [here](https://github.com/DavideM84/OctoPrint-IgnorePausedForUser/blob/533376f130b561232007e8c6da34eb19f4c05564/octoprint_IgnorePausedForUser/__init__.py#L28-L29) and [here](https://github.com/DavideM84/OctoPrint-IgnorePausedForUser/blob/533376f130b561232007e8c6da34eb19f4c05564/octoprint_IgnorePausedForUser/__init__.py#L31) because file write operations should be avoided in the communication layer to avoid blocking/stuttering, so either change that to debug or thread the writes using another method please.

_plugins/ignorepausedforuser.md
DavideM84: Fixed

_plugins/ignorepausedforuser.md
DavideM84: Fixed

_plugins/ignorepausedforuser.md
jneilliii: It's still the same @DavideM84. This line should read as follows.
```
  python: ">=2.7,<4"
```

_plugins/ignorepausedforuser.md
```diff
@@ -0,0 +1,99 @@
+---
+layout: plugin
+
+id: ignorepausedforuser
```
jneilliii, line 4: this id should match the name of the submitted plugin.md file (which it does) but it also should match [here](https://github.com/DavideM84/OctoPrint-IgnorePausedForUser/blob/f4b762611b0c28c641aa4547080a353341b1ce78/setup.py#L7) and [here](https://github.com/DavideM84/OctoPrint-IgnorePausedForUser/blob/f4b762611b0c28c641aa4547080a353341b1ce78/octoprint_IgnorePausedForUser/__init__.py#L94) in your plugin's repo. 

So the options are to adjust the `plugin_identifier` in setup.py and the dictionary name returned in the init.py file to be all lowercase or change the `id` and rename this md file to match the camel case identifier you have in your plugin.

jneilliii (changes requested):
It appears the python version didn't get updated. The link looks good. I noticed one more thing that needs to be addressed in the comments below, sorry I missed this before.

DavideM84:
All issues fixed. Thanks for supports.
Let's go ;)

jneilliii:
@FormerLurker/@cp2004 you think you guys could take a second look?

cp2004:
Seems good to me, but I'll leave it to the plugin reviewers for the final decision 😄.

FormerLurker:
Sure.  Give me a couple hours.  In the middle of a difficult build.

DavideM84:
@FormerLurker  any news ?

FormerLurker:
@DavideM84, on it, sorry for the delay.

FormerLurker:
In your ignorepausedforuser_settings.jinja2.jinja file [here](https://github.com/DavideM84/OctoPrint-IgnorePausedForUser/blob/a8348a5b55e0045301ca6ca7ae703b994a6dc18e/octoprint_ignorepausedforuser/templates/ignorepausedforuser_settings.jinja2#L5) you are retrieving an image from Github.  This isn't an error, obviously, but consider adding the image to your project and retrieving it from there lest you get issues submitted for offline users.

This is just a question about the functionality, but [in your event handler](https://github.com/DavideM84/OctoPrint-IgnorePausedForUser/blob/a8348a5b55e0045301ca6ca7ae703b994a6dc18e/octoprint_ignorepausedforuser/__init__.py#L67-L74), do you intend to create history even when the plugin is disabled?

Everything looks good though.  Just decide if you want to make any changes and let us know!

jneilliii:
I'm still concerned about the blocking communication layer that initially asked about. That's still a problem isn't it @FormerLurker? I realize this is only on paused from printer, but seems like it's doing a fair amount of IO [here](https://github.com/DavideM84/OctoPrint-IgnorePausedForUser/blob/a8348a5b55e0045301ca6ca7ae703b994a6dc18e/octoprint_ignorepausedforuser/__init__.py#L28-L31).

cp2004:
I would go the opposite way, and say that it would be quite helpful to know what has 'un paused' the printer from a log - it would not log that frequently at all, since the first time the printer is paused it just starts again. It would be interesting to know if starting a thread has more overhead - I suspect it does, but there is zero evidence there. Maybe I'll test it...

FormerLurker:
@jneilliii, that code only gets hit when [these conditions are true](https://github.com/DavideM84/OctoPrint-IgnorePausedForUser/blob/a8348a5b55e0045301ca6ca7ae703b994a6dc18e/octoprint_ignorepausedforuser/__init__.py#L18-L19).  In which case the printer is paused anyway, no?

Honestly I was considering suggesting that 'enabled' is checked before the string comparison, but I'm not sure if the dict lookup is faster or slower than the string compare.  I suspect the lookup is faster, but maybe not since there is some overhead.

cp2004:
Some comparisons here (since this got me interested) - Each test done 100000 times:

* Logging a single line `A message` to a file took 0.000067secs
* Creating a thread to do the logging in took 0.00030secs
* Simple Dict lookup - just a simple dict, rather than the overhead of OctoPrint's settings (which is a thing) - 0.00000024secs
* OctoPrint settings lookup - single key from the config - 0.000017secs
* Checking for the string in the line - 0.00000042secs

In order, from smallest to largest:
* Simple dict
* Checking string
* OctoPrint settings
* Logging
* Thread start

I don't think it is worth any difference

jneilliii:
> that code only gets hit when these conditions are true. In which case the printer is paused anyway, no?

yeah, that's why I thought it might not be such an issue.

Interesting results @cp2004 so in reality, logging a line is quicker than creating a thread. But if you follow the logic in this plugin you are logging 3 lines, and opening and writing to a json file if you follow the command from [here](https://github.com/DavideM84/OctoPrint-IgnorePausedForUser/blob/a8348a5b55e0045301ca6ca7ae703b994a6dc18e/octoprint_ignorepausedforuser/__init__.py#L31)

FormerLurker:
> Logging a single line A message to a file took 0.000067secs

If the file is already open, writing should be extremely quick, to a point.  The data is buffered, and written in the background.  Now if you overrun the buffer.....

FormerLurker:
> But if you follow the logic in this plugin you are logging 3 lines, and opening and writing to a json file if you follow the command from here

Yes, that is highly inefficient.  It is basically loading the entire file into memory, adding a record, then writing the entire file back.  Here is where we need an embedded database of some sort.

One could do something like CSV, but then 'updates' won't be efficient.  At least inserts would be constant time.

At least the history size is small (defaults to 10), so it won't be reading too much.  If it were unbounded, however....

cp2004:
I would still vote for this to be listed, there's no 'set things on fire' potential that I can see - with a look to the future improvements.

DavideM84:
All clear.
I will work on the reports tonight.

Thanks to all

jneilliii:
I agree @cp2004 no real concern to not merge, just as long as @DavideM84 is aware of the potential of issues.

FormerLurker (approved):
Going to approve this, but it sounds like there could be more code coming.  I may change the status depending on what I see.

DavideM84:
Pushed code of v1.2.7 (not relased yet)

jneilliii:
@DavideM84 can you merge in my PR here, or could you make those changes to your submission, now that the date is wrong again.

https://github.com/DavideM84/plugins.octoprint.org/pull/1

DavideM84:
@jneilliii fixed the date
Sorry but i didn't understand what else I have to do :D 
(What is missing for publishing in the repository ?)

_plugins/ignorepausedforuser.md
```diff
@@ -0,0 +1,99 @@
...
+title: IgnorePausedForUser
+description: Plugin to ignore 'Paused for User' message from printer.
+authors:
+- Davide Mencarelli <d.mencarelli@outlook.com>
```
jneilliii, line 8: I think you should take the email address part out of this line @DavideM84

jneilliii (changes requested):
Just take out the email part here from the authors list and I can merge right away.

_plugins/ignorepausedforuser.md
DavideM84: Done (I hope :D)

FormerLurker:
FYI, I looked at the recent commits, and everything looks good.

jneilliii (approved):

FormerLurker:
Hooray!  I will give this a go.

DavideM84:
Thanks to all

PR #790, opened 2021-02-20
gcodestats Estimator for Cura

NilsRo:


cp2004:
New plugins should be Python 3 compatible - it looks like this is, but it is just not mentioned in the data. Refer to https://plugins.octoprint.org/help/registering/ for details of what the format should be to add it here.

jneilliii:
actually, I think if it's left out of the submission @cp2004 it will pull that info from the github repo's init file.

cp2004:
Good point... I don't know if it is better to have it here explicitly or not.

jneilliii:
My bigger question is that this appears to be a fork of https://plugins.octoprint.org/plugins/gcodestatEstimator/ with the addition of one of the many [M73 plugins](https://plugins.octoprint.org/search/?q=m73)?

foosel:
> I don't know if it is better to have it here explicitly or not.

Just to clarify that: better have it here, one less API call during build that could fail for several possible reasons.

NilsRo:
> 
> 
> My bigger question is that this appears to be a fork of https://plugins.octoprint.org/plugins/gcodestatEstimator/ with the addition of one of the many [M73 plugins](https://plugins.octoprint.org/search/?q=m73)?

It is not related to M73 as this will add commands to the printer if I am correct. This plugin reads slicers information to update print time estimation in OctoPrint out of the information added by the slicer. 

Implemented now a generic implementation for handling M117 commands with print time estimation to cover all slicers. As I implemented everything new with additional configuration page I will close the pull to change the name of the plugin and wrote a small guide to set it up.

Issue #797, opened 2021-02-24
[Abandoned Plugin] OctoPrint-Slack


awenelo:
**Which Plugin?**

https://plugins.octoprint.org/plugins/slack/

**Link to ticket on plugin's issue tracker confirming abandonment**

richjoyce/OctoPrint-Slack#17
No response in 16 days,
No response to bug reports:
richjoyce/OctoPrint-Slack#16
No response to Python3 compatibility requests:
richjoyce/OctoPrint-Slack#15
richjoyce/OctoPrint-Slack#13

**Is there someone open to adopting it?**

Don't think so

mkevenaar:
I am willing to try to maintain this. How do I get started?

jneilliii:
The first step would be creating a fork of the plugin and update all its software update hooks and such to point to your repo on github. Then make whatever necessary changes are there for python 3 compatibility and release a new version updating the setup.py file with the new version. Once that is done, submit a PR against this repo changing the plugin's listing file to point to your github repo, add the authors: list to include yourself. We can help you through these last steps as part of our review of your PR.

mkevenaar:
I can do that, will be working on that tomorrow! Thanks

PR #803, opened 2021-02-27
Change maintainer of slack (#797)

mkevenaar:
See #797 
New release can be found here: https://github.com/mkevenaar/OctoPrint-Slack/releases/tag/0.2.4

### DIFF
--- a/_data/notices.yaml
+++ b/_data/notices.yaml
@@ -135,3 +135,11 @@
     repository of the old maintainer. Please uninstall the old plugin from the plugin manager, then reinstall from the plugin
     repository to update it. The release notes of the version by the new maintainer can be found at the "Read more..." link below.
   link: https://github.com/tkoecker/OctoPrint-Mmu2filamentselect/releases/tag/1.0.3
+- plugin: slack
+  pluginversions: ["<0.2.4"]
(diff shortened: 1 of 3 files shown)


jneilliii (changes requested):
one small fix @mkevenaar on your plugin's repo [here](https://github.com/mkevenaar/OctoPrint-Slack/blob/8c038a16f448bd5b3df565bac577a9aeca3fb5ed/setup.py#L17). This needs to match the version number in your release, unless you're doing some other versioneer trickery on release.zip via CI automation? If you are doing that approach, you may also want to consider changing your software update hook [here]() to use a different format for the URL. Something like this maybe?

```
pip="https://github.com/mkevenaar/OctoPrint-Slack/releases/download/{target_version}/release.zip"
```

mkevenaar:
@jneilliii You are totally right, I forgot to fix that. Let me fix that real quick!

mkevenaar:
@jneilliii I fixed the release, I forgot to bump the version manually. 
I did not see anyone doing things with those jinja style versioning in the plugins repo. Latest always points to the latest GH release in this case.

jneilliii (approved):

PR #806, opened 2021-03-02
Drop python2 for TranslateModel

Willmac16:


cp2004:
Add `'python_requires': ">=3,<4"` to the additional setup parameters in your setup.py, to avoid the people who will ignore the compatibility info, think they know better about it not being installable in the plugin manager and install manually

jneilliii:
You also need to update your init.py file [here](https://github.com/Willmac16/OctoPrint-TranslateModel/blob/d3cdd0ac4845638c66f6e3e1dad6fab0b7f0645c/octoprint_translatemodel/__init__.py#L157).

jneilliii:
I'll go ahead an merge assuming you're going to make those changes @Willmac16.

PR #820, opened 2021-03-12
Add Plugin filament_sensor_orangepipc

NandoGommez:


jneilliii:
rather than a fork of @deadly667 repository, wouldn't it be better to submit an adandoned issue template and adopt it? We've already got way too many filament sensor plugins out here, and I think the adoption process would be better in this specific situation or does your implementation only work on orangepi-pc?

NandoGommez:
> rather than a fork of @deadly667 repository, wouldn't it be better to submit an abandoned issue template and adopt it? We've already got way too many filament sensor plugins out here, and I think the adoption process would be better in this specific situation or does your implementation only work on orangepi-pc?

my application is specific to OrangePiPc, Although, it can work on other OrangePi, but there is no way for me to test it, since each OrangePi has its own different OS.

and since I don't have all the OrangePi cards, I only have the most used card which is the OrangePiPc.

NandoGommez:
> rather than a fork of @deadly667 repository, wouldn't it be better to submit an abandoned issue template and adopt it? We've already got way too many filament sensor plugins out here, and I think the adoption process would be better in this specific situation or does your implementation only work on orangepi-pc?

Yes, we have several filament sensors, but none work with OrangePi, most work only with Raspberry Pi

jneilliii (changes requested):
Ok, no problem since this is specific to OrangePiPC I suppose. There is one thing that needs to be addressed on your plugin's repo. [Here](https://github.com/NandoGommez/OctoPrint-Filament-Sensor-OrangePi-PC/blob/edef0fbd344dd094920b7cd334a9414d28d80c5f/octoprint_filament_sensor_ng_orangepi/__init__.py#L146) the `octoprint_filament_sensor_ng_orangepi` needs to be changed to `filament_sensor_orangepipc` to match your plugins identifier.

cp2004:
Also, enable issues so that people can get in contact should they have any problems. They're not enabled for forks by default but you can still enable it under the repository settings.

NandoGommez:
> Ok, no problem since this is specific to OrangePiPC I suppose. There is one thing that needs to be addressed on your plugin's repo. [Here](https://github.com/NandoGommez/OctoPrint-Filament-Sensor-OrangePi-PC/blob/edef0fbd344dd094920b7cd334a9414d28d80c5f/octoprint_filament_sensor_ng_orangepi/__init__.py#L146) the `octoprint_filament_sensor_ng_orangepi` needs to be changed to `filament_sensor_orangepipc` to match your plugins identifier.

Oh yes, thanks for letting me know, in fact I was already aware and editing this.
I'm working on it little by little, in my spare time.

We just needed to send it to the repository.
In fact, I have finished editing, just resubmitted to the repository and already made an initial release.

Could you check?

NandoGommez:
> Also, enable issues so that people can get in contact should they have any problems. They're not enabled for forks by default but you can still enable it under the repository settings.

Done, thanks for letting me know.

jneilliii:
I still believe this plugin submission should be handled as an adoption. Literally the only code change within the plugin itself was adding the python 3 compatibility flag, refactoring the name, and updating the documentation based on [git comparison](https://github.com/deadly667/Octoprint-Filament-Sensor-ng-OrangePi/compare/master...NandoGommez:master). 

I'd like for some other @OctoPrint/plugin-repository-maintainers to chime in here as well.

NandoGommez:
> I still believe this plugin submission should be handled as an adoption. Literally the only code change within the plugin itself was adding the python 3 compatibility flag, refactoring the name, and updating the documentation based on [git comparison](https://github.com/deadly667/Octoprint-Filament-Sensor-ng-OrangePi/compare/master...NandoGommez:master).
> 
> I'd like for some other @OctoPrint/plugin-repository-maintainers to chime in here as well.

I am working on a beta version, in which I use the base of this plugin, I am also testing an adaptation of the IFTTT to send SMS notification and / or webhooks warning about the lack of filament.

I have not yet uploaded it here on github because I am using another platform.

What made me host this adataption here on Github was just some friends that using Octoprint on OrangePiPc (like me), however they were unable to install it because of the Python version compatibility.

I can adopt this plugin without any problems and work with it in my free time, how to proceed?

cp2004:
> I still believe this plugin submission should be handled as an adoption. Literally the only code change within the plugin itself was adding the python 3 compatibility flag, refactoring the name, and updating the documentation based on [git comparison](https://github.com/deadly667/Octoprint-Filament-Sensor-ng-OrangePi/compare/master...NandoGommez:master).

I agree, looking at that diff there is no functional difference between the two plugins. It's pretty much just a bunch of name changes, adding 'PC' on the end of everything. I agree that adopting and just changing the Python compatibility string would make more sense here - we already have **8** filament sensor plugins, and 4 of them are forks of the others.

NandoGommez:
> > I still believe this plugin submission should be handled as an adoption. Literally the only code change within the plugin itself was adding the python 3 compatibility flag, refactoring the name, and updating the documentation based on [git comparison](https://github.com/deadly667/Octoprint-Filament-Sensor-ng-OrangePi/compare/master...NandoGommez:master).
> 
> I agree, looking at that diff there is no functional difference between the two plugins. It's pretty much just a bunch of name changes, adding 'PC' on the end of everything. I agree that adopting and just changing the Python compatibility string would make more sense here - we already have **8** filament sensor plugins, and 4 of them are forks of the others.

my idea is just to keep the plugin active, and as i mentioned in the previous comment, i'm making some implementations to the plugin, like for example support to IFTTT with sending SMS and gmail warning of events.

I am testing a beta version on one of my SD cards, and the one I am developing is specific for use with OrangePi PC (so I initially changed the descriptions and includes the PC specifically)

I'm thinking of creating a new repository for the plugin I'm developing based on this, and redo the Pull Request.

Thank you for your support.

PR #825, opened 2021-03-19
Add Octolabel to plugin directory

LowieGoossens:


LowieGoossens:
required key not provided @ data['layout']
Is this about layout: plugin on the second line?

foosel:
Your markdown file only consists of frontmatter, since you removed the `---` delimiter before the actual content (starting at the headline). 

Add a `---` line right before "# Octoprint-OctoLabel 1.0". And you really should not have the version number in there or add a headline either, the layout will do that for you. It is really recommended to test the page build locally yourself, the guide tells you how.

jneilliii:
Interesting project, just curious of the use case.

As foosel already suggested, you removed a line in the template that needs to be there. Also I notice you give credit to a couple of other plugins that were "forked" for your plugin, is this plugin just the label printer merged in with octorant, etc. into a single plugin? If so, why wouldn't you just do the label printing parts in your own plugin and install the others for that tweet functionality. We're just trying to be very careful about duplication of plugins now that the repo has grown so big.

LowieGoossens:
I am sorry that i mist this line, thank you a lot to help me out.

I changed the code from the octotweet plugin, in stead of sending the data to twitter the plugin now sends the data to a seperate raspberry pi with an static ip adres. On this seperate raspberry pi runs the code from pklaus to create the label.
So i want to give most credit to the other developers, I have only combined their code and made some changes to make it run smooth.

jneilliii (changes requested):
In your plugin repository you need to update the software update hook line to match your plugin's identifier. The line [here](https://github.com/LowieGoossens/octolabel/blob/efb870aedf7a34ad7512f0f72c7d97149ca706d7/octoprint_octolabel/__init__.py#L143) should be 
```
octolabel=dict(
```

jneilliii (changes requested):
@LowieGoossens I fixed up your PR to get the content back with screenshots, etc. I think there was just a little confusion about was required. There was one thing that I need you to change on your software update hook though. The line [here](https://github.com/LowieGoossens/octolabel/blob/0af3ace2b250279986fe5976b7ad6d3bc2e2d13b/octoprint_octolabel/__init__.py#L143) needs to be like this (all lowercase for your plugin_identifier).

```
octolabel=dict(

LowieGoossens:
Thank you for helping me! I did change your suggestions.

jneilliii (approved):

PR #832, opened 2021-03-30
adding OctoText plugin

berrystephenw:
Adding the OctoText plugin that gives notifications via text or email on common printer events (start, stop, file upload, error, fail, progress). It uses the email library and smtp lib for communication with a email server, and stores the password securely. Optionally it will capture a webcam snapshot and include that along with the printer status. All events are configurable by the user on the settings page. Please let me know if there are any issues and I'll get them taken care of. 

Thanks!


jneilliii:
changes/clarification requested via Discord.

cp2004:
@berrystephenw it makes more sense to include a link that will always be your latest release, such as `https://github.com/berrystephenw/OctoText/archive/refs/heads/main.zip`, do your work on another branch then merge to main and tag for a release. That way you don't have to keep updating here every time you change something in the plugin.

_plugins/OctoText.md
```diff
@@ -0,0 +1,145 @@
...
+authors:
+- Stephen Berry
+license: AGPLv3
+
+# TODO
+date: 2021-03-26
+
+homepage: https://github.com/berrystephenw/Octotext
+source: https://github.com/berrystephenw/Octotext
+archive: "https://github.com/berrystephenw/OctoText/archive/refs/tags/0.1.4.zip"
```
jneilliii, line 16: agree with @cp2004 comment, this shouldn't be to your release version, but either a branch or latest tag release.

```
archive: https://github.com/berrystephenw/OctoText/archive/refs/heads/main.zip
```

jneilliii (changes requested):

_plugins/OctoText.md
berrystephenw: ok, I like that actually since right now I'm changing three files. I'm confused as to how main.zip gets created though. When I create a release with a tag, the zip file that github makes for me is the tag.zip filename. How are they connected (tag release and main)?

cp2004: main.zip is a zip of the current state of the main branch when it is downloaded. You can find it at the big green download button on the top right of the repository.

PR #837, opened 2021-04-05
Added plugin energeniecontrol

arlohb:


jneilliii:
Link to product page seems broken? The line [here](https://github.com/arlohb/OctoPrint-EnergenieControl/blob/c3edd6ae2661d90c6b651bda8f53cfb483a940f3/octoprint_energeniecontrol/__init__.py#L164) in your plugin needs to be 
```
            energeniecontrol=dict(
```

arlohb:
I've now fixed this. Is everything now correct?

jneilliii:
Thanks, you don't have any other product page to link to? The one currently being used is broken and it appears this product may be discontinued.

arlohb:
The link was working this morning, not sure why it no longer is. I've replaced it with working links.

jneilliii (approved):

PR #850, opened 2021-04-20
Retire OctoPrint Anywhere from the plugin manager

kennethjiang:
*Note: The service is still live for existing users.*


kennethjiang:
Haha! Good thing is it's being retired because TSD is now 10x better. :)

Thanks again @jneilliii !

PR #858, opened 2021-04-26
Add NetworkHealth plugin

jonfairbanks:
Hello,

I've had some issues with the network connection on my Pi going dead while printing. Although the print finishes, it is inaccessible by SSH unless networking interfaces are restarted.

I've put together a small plugin to help with this scenario. It's helping me and hoping others find it useful as well. 

https://github.com/jonfairbanks/OctoPrint-NetworkHealth

Thanks!


_plugins/networkhealth.md
```diff
@@ -0,0 +1,47 @@
...
+id: networkhealth
+title: Network Health
+description: Monitors the health of the Network connection and restarts it if necessary
+archive: https://github.com/jonfairbanks/OctoPrint-NetworkHealth/archive/master.zip
+homepage: https://github.com/jonfairbanks/OctoPrint-NetworkHealth
+source: https://github.com/jonfairbanks/OctoPrint-NetworkHealth
+authors:
+- Jon Fairbanks
+license: AGPLv3
+featuredimage: https://raw.githubusercontent.com/jonfairbanks/OctoPrint-NetworkHealth/master/logo.png
```
jneilliii, line 12: images should be included in your PR to go on the site for GDPR related concerns. place them in `assets/img/plugins/networkhealth` and reference them with a relative url `/assets/img/plugins/networkhealth/logo.png`.

jneilliii (changes requested):
@jonfairbanks after initial review the only thing I see is the lack of a software update check. We recommend this in case you have to distribute updates related to bugs, etc. You can find more information on how to do that [here](https://docs.octoprint.org/en/master/bundledplugins/softwareupdate.html#octoprint-plugin-softwareupdate-check-config). The other thing is commented in code relative to the featured image. The other option is to not include a featured image at all, it's not required. That specific one is what gets distributed with new plugin announcements on twitter, discord, within octoprint.

_plugins/networkhealth.md
jonfairbanks: Updated PR to reflect local assets

jonfairbanks:
Thanks for the quick review, @jneilliii. I have update the PR and the plugin with fixes for the feedback above.

foosel:
I just looked at the code and that plugin seems to have some issues. I'm on mobile right now, so please hold on merging, will provide details as soon as I'm back on a PC.

_plugins/networkhealth.md
```diff
@@ -0,0 +1,47 @@
...
+date: 2021-04-20
+
+tags:
+- network
+- wifi
+- ethernet
+
+compatibility:
+
+  os:
```
foosel, line 22: Your plugin hard codes linux paths and commands, so it is definitely not windows or macos compatible, and I have my doubts about freebsd too. This should be 

``` yaml
compatibility:
  os:
    - linux
```

only.

foosel (changes requested):
Ok, so, the issue is as follows. The plugin itself is defined [here](https://github.com/jonfairbanks/OctoPrint-NetworkHealth/blob/953ed81f7fd61bb7332592a53921668f48d50605/octoprint_NetworkHealth/__init__.py#L6-L8).

The lines that actually *do* something (which are [here](https://github.com/jonfairbanks/OctoPrint-NetworkHealth/blob/953ed81f7fd61bb7332592a53921668f48d50605/octoprint_NetworkHealth/__init__.py#L17-L48)) never get called. And they would in fact also fail because [this line](https://github.com/jonfairbanks/OctoPrint-NetworkHealth/blob/953ed81f7fd61bb7332592a53921668f48d50605/octoprint_NetworkHealth/__init__.py#L42) refers to `self` which doesn't even exist in this standalone function. The same goes for your software update hook handler, which is currently broken not only due to that but also because it uses the wrong identifier.

Suggestion, turn all of this into this:

``` python
import os
import time

from octoprint.util import RepeatedTimer

class NetworkHealthPlugin(octoprint.plugin.StartupPlugin):
	def __init__(self):
		self._timer = None

    def on_after_startup(self):
        self._logger.info("Beginning to monitor Network health...")
		self._timer = RepeatedTimer(180, self._check_network)
		self._timer.start()

	def _check_network(self):
		try:
	          if not check_ping():
	              self._logger.error("No Network Connection - Resetting Adapter(s)...")
	              reset_wlan0 = 'ip link set wlan0 down; ip link set wlan0 up'
	              reset_eth0  = 'ip link set eth0 down; ip link set eth0 up'
	              os.system(reset_wlan0)
	              os.system(reset_eth0)
		except Exception:
			self._logger.exception("Could not run network health check")

	def get_update_information(self, *args, **kwargs):
    	return dict(
        	networkhealth=dict(
            	displayName=self._plugin_name,
	            displayVersion=self._plugin_version,

    	        type="github_release",
	            current=self._plugin_version,
	            user="jonfairbanks",
	            repo="OctoPrint-NetworkHealth",

	            pip="https://github.com/jonfairbanks/OctoPrint-NetworkHealth/archive/{target}.zip"
	        )
	    )


def default_gateway():
    import socket
	import struct
    with open("/proc/net/route") as fh:
        for line in fh:
            fields = line.strip().split()
            if fields[1] != '00000000' or not int(fields[3], 16) & 2:
                continue
            return socket.inet_ntoa(struct.pack("<L", int(fields[2], 16)))

def check_ping():
    hostname = default_gateway()
    response = os.system("ping -c 4 " + hostname)
    if response == 0:
        return True
    else:
        return False


__plugin_name__ = "Network Health Monitor"
__plugin_version__ = "1.0.0"
__plugin_description__ = "Monitors the health of the Network connection and restarts it if necessary"
__plugin_pythoncompat__ = ">=2.7,<4"
__plugin_implementation__ = NetworkHealthPlugin()
__plugin_hooks__ = {
	"octoprint.plugin.softwareupdate.check_config": __plugin_implementation__.get_update_information
}
```

That way it will a) run the check in its own thread with a proven to work framework utility class b) only after server startup, c) the software update also works and d) there's some error handling. 

You'll also need to mark your plugin as Linux only because it has some Linux-only stuff hardcoded. Might also be compatible to BSD, I don't know, but if you've only tested it under Linux, it should be marked as such. It would also be good idea to add an os check to the plugin itself, replacing the control properties like this:

``` python
__plugin_name__ = "Network Health Monitor"
__plugin_version__ = "1.0.0"
__plugin_description__ = "Monitors the health of the Network connection and restarts it if necessary"
__plugin_pythoncompat__ = ">=2.7,<4"

def __plugin_check__():
	import sys

	if not sys.platform.startswith("linux"):
		import logging
		logging.getLogger("octoprint.plugins.networkhealth").error("Cannot run this plugin under anything but Linux")
		return False

	return True

def __plugin_load__():
	global __plugin_implementation__
	global __plugin_hooks__

	__plugin_implementation__ = NetworkHealthPlugin()
	__plugin_hooks__ = {
		"octoprint.plugin.softwareupdate.check_config": 		__plugin_implementation__.get_update_information
	}
```

jneilliii:
Thanks Gina. I was going to do a full plugin review once those other changes were done, and see that you already provided necessary changes.

jneilliii (changes requested):
@jonfairbanks the PR looks good now, just need the changes suggested by foosel on your plugin directly though. Let me know when that's done and we can move forward. I'm looking forward to this one because we have a mega-thread on the forum about wifi disconnects.

jonfairbanks:
Thanks for the comments. Working on them now and will update here once complete.

jonfairbanks:
@foosel @jneilliii The PR's above should resolve the issues above. Thanks again for reviewing.

jneilliii:
one more small change I think would be beneficial @jonfairbanks. The line [here](https://github.com/jonfairbanks/OctoPrint-NetworkHealth/blob/7cdfc16e38d4b6f819cef76e9d607ffb2aa1917d/octoprint_NetworkHealth/__init__.py#L7) I think should also include RestartNeeded mixin to trigger an OctoPrint restart after install. Otherwise the plugin doesn't get initialized after install I think.

```
class NetworkHealthPlugin(octoprint.plugin.StartupPlugin, octoprint.plugin.RestartNeedingPlugin):
```

jonfairbanks:
Added the restart capabilities: https://github.com/jonfairbanks/OctoPrint-NetworkHealth/pull/7

foosel (approved):

jneilliii (approved):

jonfairbanks:
Nice, thank you both! 😄

PR #859, opened 2021-04-26
update screen shots, and title

lsellens:


jneilliii (changes requested):
date field should never be modified from original submission date @lsellens.

lsellens:
@jneilliii fixed. Thanks

PR #861, opened 2021-04-26
[Simple Emergency Stop] Reduce donation size button

Sebclem:
The Buy me a coffee was to big and aggressive, ducing it.

### DIFF
--- a/_plugins/simpleemergencystop.md
+++ b/_plugins/simpleemergencystop.md
@@ -50,7 +50,5 @@ If you need more information, you can find the project's [Github Page here](http
 
 
 
-[![Buy me a coffee][buymeacoffee-shield]][buymeacoffee]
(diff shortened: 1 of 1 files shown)


jneilliii (changes requested):
img sources need to be part of the PR for GDPR concerns and stored in the folder `assets/img/plugins/simpleemergencystop/`, but looking at this it seems it's a dynamic badge image with a heart count or something? Might need to get @foosel 's opinion on this one.

foosel:
Would be better to keep it local, and I'm having some worries here, but since it's an external service with its own proper [Privacy Policy](https://www.buymeacoffee.com/privacy-policy) which at least claims there's no PII storage happening without an active login, it should be fine.

jneilliii (approved):

PR #864, opened 2021-04-29
Add Filament Scale Enhanced

techman83:
This is a continuation of Victor's work, the plugin appeared to be abandoned and opened an issure dieki-n/Octoprint-Filament-Scale#14

Changes:
- Updated to Python 3
- Filtering of noisy sensor data
- Code formatting

It wasn't previously listed, so I kept the identifier for easier transition.


jneilliii:
@techman83 one major problem here. Your fork is using the same plugin_identifier as the original plugin that is not registered with the plugin repo and this will cause a conflict for the original plugin author and any users with that plugin already installed. I'm a little confused while dieki-n didn't register their plugin in the first place, but even more confused why they wouldn't consider collaboration via a PR. 

Another thing is I see your repo doesn't have issues enabled, we like to see those enabled so there is somewhere that users having problems can be pointed to for help.

avr12345:
hi, Is there a chance to connect more then one sensor? I have 4 sensors/spools in line with switchable RJ45 plugs. but its not that accurate because of the calibration. would be awesome to have 4 independent sensors connected

techman83:
> @techman83 one major problem here. Your fork is using the same plugin_identifier as the original plugin that is not registered with the plugin repo and this will cause a conflict for the original plugin author and any users with that plugin already installed. I'm a little confused while dieki-n didn't register their plugin in the first place, but even more confused why they wouldn't consider collaboration via a PR.

Likely life, there hasn't been a change in 2 years. It seemed abandoned and it's a seamless upgrade from the original, to this version. I only changed the name as technically Victor would have the copyright over that, even though the source code is open and I don't have permission.

Changing the identifier broke a lot of stuff that I haven't had time to figure out (been nearly 6 or so years since I played with knockout), and technically it was never claimed, so, I claimed it :)

> 
> Another thing is I see your repo doesn't have issues enabled, we like to see those enabled so there is somewhere that users having problems can be pointed to for help.

Oh! Yes, my mistake. Fixed!

> hi, Is there a chance to connect more then one sensor? I have 4 sensors/spools in line with switchable RJ45 plugs. but its not that accurate because of the calibration. would be awesome to have 4 independent sensors connected

@avr12345 - open a feature request over at my github. A discussion here wouldn't be appropriate (though that's on me for not turning on issues!)

jneilliii (approved):

PR #865, opened 2021-04-29
Added plugin camerasettings

The-EG:
Camera Settings allows a user to interactively change camera settings.

image

OS requirement set to Linux because this requires `v4l2-ctl`. Technically, 'posix' or 'nix' may also be appropriate but I'm unsure if this would work on MacOSX or FreeBSD. If I find a user that wishes to use this plugin on one of those systems and is willing to test, I will update the requirement after verifying it.

edit: Python 3 only! There are some features used that are not Python 2 compatible, notably `os.scandir` and `subprocess.check_output`


jneilliii (changes requested):
small issue on your plugin's side. The line [here](https://github.com/The-EG/OctoPrint-CameraSettings/blob/3d4b172d4ce4ffac647a61e6f3260da9be83e13f/octoprint_CameraSettings/__init__.py#L192) needs to return a dictionary object named the same as your plugin_identifier (ase sensitive). 

```
camerasettings=dict(
```

jneilliii:
you may also consider adding plugin_check control property to not load if not on linux, in case someone attempts to install manually. Also maybe `additional_setup_parameters = {"python_requires": ">=3, <4"}` to prevent non python 3 installs in setup.py.

The-EG:
> small issue on your plugin's side. The line [here](https://github.com/The-EG/OctoPrint-CameraSettings/blob/3d4b172d4ce4ffac647a61e6f3260da9be83e13f/octoprint_CameraSettings/__init__.py#L192) needs to return a dictionary object named the same as your plugin_identifier (ase sensitive).
> 
> ```
> camerasettings=dict(
> ```

Arggh, I knew having it lower case in one spot and not the other would bite me somewhere....thanks! That's been fixed.

> you may also consider adding __plugin_check__ control property to not load if not on linux, in case someone attempts to install manually. Also maybe additional_setup_parameters = {"python_requires": ">=3, <4"} to prevent non python 3 installs in setup.py.

Implemented both of these too, and tested them. Interestingly enough, in both cases the 'success' popup still comes up, but says  'Installed: unknown'.

cp2004:
> Implemented both of these too, and tested them. Interestingly enough, in both cases the 'success' popup still comes up, but says 'Installed: unknown'.

Yeah it's sometimes strange with that popup, because the notification is reused for all plugin installs you can get into a situation with a green popup, cross icon and 'Installed *Unknown*. At least you should also get a screen like this:

image

It seems to have been quite effective in communicating that people can't install the plugin, at least from general experience.

In other news, Python 3 usage is at 41.7%, which is awesome.

The-EG:
> In other news, Python 3 usage is at 41.7%, which is awesome.

Maybe I can help convince a few more to convert 😄

jneilliii (approved):

PR #870, opened 2021-05-03
Added plugin print_eta

lewisbennett:


_plugins/print_eta.md
```diff
@@ -0,0 +1,52 @@
...
+source: https://github.com/lewisbennett/OctoPrint-Print-ETA
+archive: https://github.com/lewisbennett/OctoPrint-Print-ETA/archive/master.zip
+
+tags:
+- time
+- eta
+- finish time
+
+screenshots:
+- url: url of a screenshot, /assets/img/plugins/print_eta/screenshot.png
```
jneilliii, line 23: should only be the url to the image, no comments...

jneilliii (changes requested):
How is this different from [Display ETA'(https://plugins.octoprint.org/plugins/display_eta/) plugin which you apparently used as a starting point, since it's showing up in your update hook [here](https://github.com/lewisbennett/OctoPrint-Print-ETA/blob/410e5ffb8d2cc0e2e8adb5c419beaade86158b67/octoprint_print_eta/__init__.py#L64)? That needs to be `print_eta` instead of `display_eta` BTW. Also, see the failed actions above, looks like you need to fix your screenshot URL, should only be the URL.

_plugins/print_eta.md
lewisbennett: Hi, yes, Display ETA was used as a starting point. I and some friends have been using that plugin for a while but were experiencing issues with it, and I've been wanting to look into OctoPrint plugin development for a while, so saw it as an opportunity to learn.

Print ETA, currently, does the same job however, I have fixed the issues (mainly unreliability of ETA readings showing within OctoPrint) and feel like I've improved the settings too.

Thanks for pointing out the issues, have fixed and committed.

cp2004: Was there a reason you chose not to collaborate with the original author of DisplayETA to get these issues fixed and submit a Pull Request?

lewisbennett: Not really. As I said, I wanted to look into learning about building OctoPrint plugins, so this just seemed like a good place to start.

jneilliii: I agree, collaborating with the original plugin would have been a better approach here IMO. We are trying to limit the number of duplicated plugins on the repo, especially after the plague of filament sensor plugins.

jneilliii: That being said, we are not trying to discourage anyone from learning plugin development and contributing to the community. If the other plugin author is not open to collaboration then of course we would accept this PR, and may anyway as it does have additional functionality (navbar) that the original plugin does not.

lewisbennett: Alright, fair enough.

I've learnt something either way regardless if it gets published or not, so I'm happy. I may reach out to the Display ETA author like you suggested.

If nothing else, I guess it was worth a shot 😄 Thanks guys.

lewisbennett: I'll still be using this version personally, so will likely end up adding and changing things further down the line.

I may revisit trying to publish it in the future depending on a) the outcome of reaching out to the author of Display ETA or b) adding enough additional functionality that differentiates the plugins further.

jneilliii: I'm pretty sure that plugin was adopted by @AlexVerrico and he'll more than likely be open to collaboration.

jneilliii:
Ok, checks are passing now, github actions are happy. I was curious though, is it required that your plugin react to ever event that OctoPrint fires? It seems that's what your on_event override method is doing.

@lewisbennett if you are unable to get any response from @AlexVerrico please reply back here. I will leave this PR open in the interim, and may request @foosel to chime in on adoption of this plugin.

lewisbennett:
Not necessary, I don't think. Complete oversight actually. I did originally have it only reacting to `PrintStarted` and `PrintResumed` (I think those are right), but needed it to be more reactive, so disabled the filter. I will review the OctoPrint docs and try to filter it down some more. The plugin doesn't actually notify OctoPrint or the printer unless the string is different to the one calculated previously, but I guess querying the actual event will allow me to prevent unnecessary ETA calculations too.

Thanks again.

_plugins/print_eta.md
AlexVerrico: G'Day all. 
I'm defintely interested in collaborating, feel free to open a PR or issue over on my plugin repo if you want to discuss futher. 
Cheers, Alex.

bell47 (approved):

bell47 (changes requested):
Kalau bisa ada perubahan lagi

bell47 (approved):

foosel:
@bell47 don't abuse the review function.

PR #887, opened 2021-05-18
Added plugin draggable_files

SanderRonde:
Add plugin `draggable_files`. It adds the ability to drag files in the File Manager to folders. So dragging files/folders **between** the file manager folders (see gif).

dragging

It includes the additional feature of making the file manager resizable. See gif below.

resizing


jneilliii (approved):
Nice plugin @SanderRonde

SanderRonde:
@jneilliii I've got a question about how the plugin is shown in the repository that you might know the answer to. My plugin is listed with the display name "OctoPrint-Draggable_files" even though I'm pretty sure I set the title to "Draggable Files" in the relevant fields. Do you know why this is the case? It makes finding it a bit harder since "draggable files" no longer matches. It even appears that a while ago it was listed under the correct display name. This is now causing some issues with updating. See for example this image where it's listed as having no remote:

image

This also appears to be an issue for other users: https://github.com/SanderRonde/Octoprint-Draggable-Files/issues/6#issuecomment-858503372

Do you have an idea as to what might be causing this? Thanks in advance!

kantlivelong:
@SanderRonde If I'm understanding correctly you would want to change the `title` key in
https://raw.githubusercontent.com/OctoPrint/plugins.octoprint.org/gh-pages/_plugins/draggable_files.md

cp2004:
This is two different issues in one. First, you have the 'title' of the plugin on the repository, this is set by the title field mentioned above. 
Second, is your issue with updating. OctoPrint does not use the repository to check for updates, it goes to whatever configuration you set up under the `get_update_information` hook. Here you have [the `repo` field misconfigured](https://github.com/SanderRonde/Octoprint-Draggable-Files/blob/377868130ba251fb33973f8092fbbcb5f93d9585/octoprint_draggable_files/__init__.py#L42), it needs to match the name of the repository, https://github.com/SanderRonde/OctoPrint-Draggable_files sends to a 404

PR #894, opened 2021-06-02
Add Navbar Clock plugin

ManuelMcLure:
Plugin to add a clock to the navbar.


jneilliii (approved):
Nothing critical @ManuelMcLure but your plugin has "empty" css and less files that are still included in get_assets, etc. No show stopper, assuming you're leaving those there as placeholders in case you need them in the future otherwise, you may want to just remove them, I know those come along in the cookie cutter template.

PR #896, opened 2021-06-08
Added plugin ArducamCameraControl

ArduCAM:


jneilliii:
So this plugin is a duplicate of ArducamFocus plugin with support for PTZ? Wouldn't it be better to contribute to the ArducamFocus plugin to add this support rather than having a duplicate plugin? I'm pretty sure @moof-src would be open to collaboration.

moof-src:
Arducam sent me camera samples and a link to the new plugin.  I would prefer they took over development.  In the meantime, I plan on adding their support for the PTZ to my plugin in the near future.

jneilliii:
Ok, thanks @moof-src for verifying. Only one problem I see with the current plugin @ArduCAM. The latest release version doesn't match the current version specified in setup.py, which will lead to update loops where it will keep trying to update to version 0.2 because what it thinks is installed is 0.1. Please update setup.py plugin_version and re-release 0.2 and we should be good to adopt this plugin.

_plugins/ArducamCameraControl.md
```diff
@@ -0,0 +1,149 @@
...
+  # A single version number will be interpretated as a minimum version requirement,
+  # e.g. "1.3.1" will show the plugin as compatible to OctoPrint versions 1.3.1 and up.
+  # More sophisticated version requirements can be modelled too by using PEP440
+  # compatible version specifiers.
+  #
+  # You can also remove the whole "octoprint" block. Removing it will default to all
+  # OctoPrint versions being supported.
+
+  octoprint:
+  - 1.2.0
```
cp2004, line 60: The plugin makes use of `octoprint.access.Permissons`, which is good, but this is only available since OctoPrint 1.4.0, so this compatibility needs to be 1.4.0 at least.

cp2004 (changes requested):

AllaboutJay:
Hello,

Thanks for contacting Arducam.

Thank you very much for your suggestion, the matching release version is still under test, we will release it the first time after the test is complete.


If you have any questions, kindly contact us.
Thanks!

jneilliii (approved):

PR #901, opened 2021-07-05
added GPIO Fan Controller plugin

z4gunn:


cp2004:
One thing I noticed - your plugin imports `gpiozero` but this package is not in the plugin requirements. This will cause it to fail to load. Add dependencies you want to install alongside the plugin in `plugin_requires` [here](https://github.com/z4gunn/OctoPrint-GpioFanController/blob/7df042cdf2e2ac4d374e57210d6afc3090c38bcf/setup.py#L36).

z4gunn:
> One thing I noticed - your plugin imports `gpiozero` but this package is not in the plugin requirements. This will cause it to fail to load for users without this package (which would be coincidence if they do have it). Add dependencies you want to install alongside the plugin in `plugin_requires` [here](https://github.com/z4gunn/OctoPrint-GpioFanController/blob/7df042cdf2e2ac4d374e57210d6afc3090c38bcf/setup.py#L36).

Good catch i will add gpiozero to the requirements

jneilliii:
@z4gunn in addition to that plugin_requires, you may want to consider adding python 3 to `additional_setup_parameters` like in the example [here](https://github.com/jneilliii/OctoPrint-GoogleDriveBackup/blob/b9b7d4d661bba92d05f99f26b28fd5948f916db8/setup.py#L62) to avoid people trying to side-install outside of the plugin repository. Let us know when you've at least completed the addition of gpiozero and we can continue with the verification process.

z4gunn:
> @z4gunn in addition to that plugin_requires, you may want to consider adding python 3 to `additional_setup_parameters` like in the example [here](https://github.com/jneilliii/OctoPrint-GoogleDriveBackup/blob/b9b7d4d661bba92d05f99f26b28fd5948f916db8/setup.py#L62) to avoid people trying to side-install outside of the plugin repository. Let us know when you've at least completed the addition of gpiozero and we can continue with the verification process.
> 
> EDIT: I don't see any issues accepting this PR once those changes are completed. Looks pretty straight-forward and nothing stands out as breaking.

@jneilliii thanks for the feedback.  I updated requirements in latest [release](https://github.com/z4gunn/OctoPrint-GpioFanController/releases/tag/0.1.1).

jneilliii (approved):

PR #903, opened 2021-07-09
Added plugin LightControls

RoboMagus:


_plugins/LightControls.md
```diff
@@ -0,0 +1,85 @@
...
+authors:
+- RoboMagus
+license: AGPLv3
+
+# TODO
+date: 2021-07-09
+
+homepage: https://github.com/RoboMagus/OctoPrint-LightControls
+source: https://github.com/RoboMagus/OctoPrint-LightControls
+archive: https://github.com/RoboMagus/OctoPrint-LightControls/archive/master.zip
```
jneilliii, line 16: incorrect URL, your default branch is main on github, adjust accordingly.

jneilliii (commented):
need to adjust the archive URL as noted. Also would recommend updating the `additional_setup_parameters` to include python 3 compatibility to prevent side-installing outside of the plugin manager like [here](https://github.com/jneilliii/OctoPrint-GoogleDriveBackup/blob/b9b7d4d661bba92d05f99f26b28fd5948f916db8/setup.py#L62).

RoboMagus:
@jneilliii Ah thanks, good catch. I've updated all references to master.
I still can't get used to the fact that the default branches of GH have now been changed to main.

Do you know if the Cookie Cutter templates have been updated to reflect the changed defaults? If so, I might need to update something.

cp2004:
Cookiecutter template is still using `master` - git itself still creates repositories with `master` as the default, but this is configurable. It's Git*Hub* that is using main - either way we are going to get it wrong for some people.

jneilliii (approved):
Looks good. Quick question though for my own personal curiosity with other plugins I've seen. Does the valueUpdate='input' on the slider binding make it not post an API call on every move of the slider, but when released?

RoboMagus:
@jneilliii The way I wanted the sliders to operate is to have it update as soon as possible when interacting with it rather than on release or click. With the bindings used this works out great. To avoid flooding the network though, KO's rate limit feature comes in handy: `this.extend({ rateLimit: 100 });` used [here](https://github.com/RoboMagus/OctoPrint-LightControls/blob/main/octoprint_LightControls/static/js/LightControls.js#L31).

jneilliii:
Thanks, was curious because I noticed a recent plugin I installed floods the api when sliding.

PR #905, opened 2021-07-11
Added plugin mote_lighting_control

nigelm:


jneilliii (changes requested):
We request that you include your screenshots as part of the PR. put them in `/assets/img/plugins/mote_lighting_control` and adjust URLs accordingly. 

I'd also recommend updating `additional_setup_parameters` to include the python 3 requirement to prevent users from side-loading using the URL directly, an example can be found [here](https://github.com/jneilliii/OctoPrint-GoogleDriveBackup/blob/b9b7d4d661bba92d05f99f26b28fd5948f916db8/setup.py#L62).

> Unfortunately the plugin documentation is insufficiently complete to make sole
use of - so inspecting other plugins is the only option.

Lastly, I'm curious what you found lacking in the documentation? It's not uncommon to use other plugins to figure things out, so that's of course no issue.

_plugins/mote_lighting_control.md
```diff
@@ -0,0 +1,101 @@
...
+This plugin draws very heavily on
+[`OctoPrintMote`](https://github.com/topshed/OctoPrintMote) by
+[Richard Hayler](https://github.com/topshed)
+
+Additionally some code was used from
+[`OctoLight`](https://plugins.octoprint.org/plugins/octolight/)
+(the API handling code and js), and
+[`TP-Link Smartplug`](https://plugins.octoprint.org/plugins/tplinksmartplug/).
+
+Unfortunately the plugin documentation is insufficiently complete to make sole
```
cp2004, line 100: I don't think this is right to put on the plugin listing - especially not on the official plugin repository.

As with all things, feedback is critical. Tell us what you found hard, reach out in the community and usually when people point out documentation issues, then we work *with* you to make it clearer for everyone else. But please, don't just passively complain without saying what is *actually* wrong.

nigelm:
Changes made.

Regarding the documentation comments, I think the main issue is that saying use existing plugins as examples is a bit similar to saying you should find example code on stackoverflow - there is lots there, but it takes rather more work to sift out good examples from mediocre, and best practices are even harder to properly identify.

For example error handling.  The original code I worked from appeared to be working on the basis that if there was a problem talking to the hardware that this code was for, then it should hard error taking down the entire system.  But there is presumably a set of best practices somewhere between basically detecting and silencing issues with only minimal logging, spamming the log with error messages, raising exceptions etc...   Some of these ex[ectations of plugins should be made more explicit.

An ideal would be to have a complex (api, config adn event handling) plugin that is "blessed" as a best practice example.

cp2004:
I have been wondering about adding to the plugin docs a 'best practices' guide for certain things. It has to be a balance between teaching someone how to code (experienced developers will come by complaining that there's different better ways etc. etc.) and providing some tips of how to do it. I do want to make a best practices for frontend styling, so we can try and help out those making themes & things to try and keep it consistent.

If you want to get involved in collecting ideas for a 'best practices' guide, feel free to open an issue for discussion on the OctoPrint repo or have a chat on the [OctoPrint discord](https://discord.octoprint.org) as well.

jneilliii (approved):

PR #909, opened 2021-07-17
Updated URLs with correct username & fixed version compatibility

farfalleflickan:


jneilliii (approved):
@farfalleflickan I would recommend updating your additional_setup_parameters like [this](https://github.com/jneilliii/OctoPrint-GoogleDriveBackup/blob/b9b7d4d661bba92d05f99f26b28fd5948f916db8/setup.py#L62) now that you are switching to Python 3 only. It will prevent people from installing directly using the URL in python 2 environments.

PR #912, opened 2021-07-21
Add matrix notifier

Cadair:
Hello :wave: 

I wrote a simple plugin which sends messages to matrix, hopefully I have done everything I need to do to get it included.


jneilliii (approved):
we recommend setting `additional_setup_parameters` in setup.py like [this](https://github.com/jneilliii/OctoPrint-GoogleDriveBackup/blob/b9b7d4d661bba92d05f99f26b28fd5948f916db8/setup.py#L62) for python 3 only plugins to prevent installing from URL directly vs plugin manager. 

Also, if you don't mind @Cadair, could you update the date field? It's a couple of days back.

Cadair:
Thanks for the pointer, I have pushed a new version with that in it.

jneilliii (approved):

jneilliii:
@cp2004 do you mind taking a quick look? I don't see any issues.

cp2004 (approved):
*Almost* perfect, but the software update hook needs adjusting here:

https://github.com/Cadair/octoprint-matrix-notifier-plugin/blob/9b69f61baf9a3cfeb0e2669ccdf6b1a889ee94da/octoprint_matrix_notifier/plugin.py#L123

```diff
- "repo": "OctoPrint-Matrix_Notifier",
+ "repo": "octoprint-matrix-notifier-plugin", 
```

If people come around asking about why it won't install on Python 2, then the link you need is https://community.octoprint.org/t/python-version-mismatches-in-plugins-and-how-to-solve-them/35160 - but this situation is a lot better than it used to be so I wouldn't worry about that!

Otherwise, nice plugin - once the update hook is adjusted we should be good to go.

jneilliii:
Good catch, thanks Charlie. 

@Cadair let us know when that is complete and we can merge.

Cadair:
Thanks, I have pushed a new version with the correct repo name :)

cp2004:
I was about to merge, but I can see you created a new release, for version 0.2.0 on GitHub. However, the version number has not been changed in `setup.py`. This will result in an update loop as the plugin sees 0.2.0 as available but when downloaded it is identifying as 0.1.0.

Before every release/tag, you'll have to bump the version number. You'll probably need to go to 0.2.1 in setup.py and then release 0.2.1 on GitHub after this is pushed.

Cadair:
I am out of the habit of needing to manually change the version number in setup.py, all my main projects use setuptools_scm now. I have fixed thanks.

cp2004:
There are ways to do that, I use Python-versioneer in most of mine to avoid having to remember to bump the versions manually. Anyway, I think it should be good to go 🚀

PR #913, opened 2021-07-22
Fix image url in matrix notifier

Cadair:
I got the URL wrong and didn't notice till I saw the update in the octoprint UI :frowning_face:


cp2004:
I just noticed the same thing too, need to work on getting preview deployments going to make catching things like this easier.

cp2004 (approved):

PR #916, opened 2021-07-27
 🧟 PrintHistory plugin has been resurrected. Adding back

guysoft:
New 1.3 version adds python3 support and handles folder payload correctly.

### DIFF
--- a/_plugins/printhistory.md
+++ b/_plugins/printhistory.md
@@ -1,0 +1,16 @@
+---
+layout: plugin
+id: printhistory
+title: Print History
+description: Saves filename, print time and filament usage for each print
+archive: https://github.com/imrahil/OctoPrint-PrintHistory/archive/master.zip
+homepage: https://github.com/imrahil/OctoPrint-PrintHistory
+source: https://github.com/imrahil/OctoPrint-PrintHistory
+author: Guy Sheffer
+license: AGPLv3
+featuredimage: https://github.com/imrahil/OctoPrint-PrintHistory/raw/master/printhistory.png?raw=true
+date: 2020-07-27
+
+---
+
+![PrintHistory](https://github.com/imrahil/OctoPrint-PrintHistory/raw/master/printhistory.png?raw=true)


jneilliii:
good to hear @guysoft, one thing that needs to be fixed however is the missing `__plugin_pythoncompat__ = ">=2.7,<4"` flag in your init.py file. Also we have started requesting all images be bundled into the assets/plugin/printhistory/ folder and linked from there instead of to the github repo and links updated accordingly.

jneilliii:
we also like to give credit for the original authors, so if you don't mind changing `author` to an `authors` list with the original owner that would be great.

guysoft:
Hey, not sure what is marked as not compatible
The error is: ``not flagged as Python 3 compatible @ data``

guysoft:
@jneilliii Thank for answering all my questions and even the ones I didn't even think of in time.
New to plugin contribution :)
Will fix

guysoft:
Seems like python3 is already in place:
https://github.com/imrahil/OctoPrint-PrintHistory/blob/master/octoprint_printhistory/__init__.py#L12

guysoft:
Ok, seems like you had to add compatibility oh the .md file and not only on the __init__.py . That solved it.

_plugins/printhistory.md
```diff
@@ -0,0 +1,34 @@
...
+archive: https://github.com/imrahil/OctoPrint-PrintHistory/archive/master.zip
+
+featuredimage: /assets/img/plugins/printhistory/printhistory.png
+
+tags:
+- history
+- log
+
+screenshots:
+- url: https://github.com/imrahil/OctoPrint-PrintHistory/raw/master/printhistory.png?raw=true
```
jneilliii, line 25: this one should also be `/assets/img/plugins/printhistory/printhistory.png`

jneilliii (changes requested):
@guysoft sorry about the python compatibility flag, used to seeing it at the bottom. 

Looks like you missed one image url though, see my comments.

_plugins/printhistory.md
guysoft: Done

guysoft:
Ok, Fixed it, I am not sure if the review has been resolved because is stell flashing red, did mark it as resolved.

PR #929, opened 2021-08-19
Added BLOCKS Plugin

HugoCLSC:
A plugins that was built with the BLOCKS 3D printers in mind.
Offers a different layout of the interface.


_plugins/BLOCKS.md
```diff
@@ -0,0 +1,72 @@
...
+homepage: https://github.com/HugoCLSC/BLOCKSUI
+source: https://github.com/HugoCLSC/BLOCKSUI
+archive: https://github.com/HugoCLSC/BLOCKSUI/archive/master.zip
+
+tags:
+- a list
+- of tags
+- that apply
+- to your plugin
+- (take a look at the existing plugins for what makes sense here)
```
jneilliii, line 22: remove this

_plugins/BLOCKS.md
```diff
@@ -0,0 +1,72 @@
...
+
+screenshots:
+- url: /assets/img/plugins/ScreenShot#1.PNG
```
jneilliii, line 25: all these image URLs are incorrect, need to be `/assets/img/plugins/BLOCKS/...` paths are case sensitive.

_plugins/BLOCKS.md
```diff
@@ -0,0 +1,72 @@
...
+  caption: Dark
+- url:  /assets/img/plugins/ScreenShot#3.PNG
+  alt:  Light Connected
+  caption: Light connected
+- url:  /assets/img/plugins/ScreenShot#4.PNG
+  alt:  Dark Connected
+  caption: Dark connected
+
+
+featuredimage: /assets/img/plugins/Blocks_Logo.png
```
jneilliii, line 39: same as above, but also missing the image as part of the PR.

jneilliii (changes requested):
There is quite a few things that need to be corrected in this PR. Have yet to review your plugin directly, but will do so and comment here if I see anything wrong.

_plugins/BLOCKS.md
HugoCLSC: Fixed

_plugins/BLOCKS.md
HugoCLSC: And also fixed

_plugins/BLOCKS.md
jneilliii: the URLs appear to still be incorrect for all the images.

_plugins/BLOCKS.md
jneilliii: Still don't see the image `Blocks_Logo.png` included with this PR.

cp2004:
Preview URL (manually updated by me) https://cp2004-plugins-octoprint-org-git-pr-929-cp2004.vercel.app/

HugoCLSC:
You are completely right. I'm going to correct that asp.
Appreciate all the help

cp2004:
I've updated the preview but it still doesn't seem to be showing the images right. It is manually updated by syncing to my own git repo, and I probably won't do it again this evening.

You can setup a local preview of the site using jekyll, follow step 4 here https://plugins.octoprint.org/help/registering/ - doing it on Linux is much easier than windows.

cp2004:
It's likely the `#` in the URLs, don't think you can have that character.

HugoCLSC:
Sorry for all the trouble. I'll get to it when I get a chance.
Yea I was trying the Jekyll, but I'm on windows and it always bugs :/

_plugins/BLOCKS.md
```diff
@@ -0,0 +1,66 @@
...
+# Blocks Plugin
+
+This Plugin was made with the intention to offer a personalized interface to the users of the Blocks 3D Printers.
+The Plugin offers more notifications, a new theme and the ability to change between Light and Dark themes.
+
+The plugin is still in development.
+http://plugins.octoprint.org/plugin/BLOCKS/
+
+
+###Acknowledgments
```
cp2004, line 61:
```suggestion
### Acknowledgments
```
Spotted this issue in the preview, Markdown wants a space

_plugins/BLOCKS.md
```diff
@@ -0,0 +1,66 @@
...
+
+First of all, Thanks OctoPrint devs.
+Used and studied the code from:
+
+*https://github.com/LazeMSS/OctoPrint-UICustomizer/
```
cp2004, line 66:
```suggestion
* https://github.com/LazeMSS/OctoPrint-UICustomizer/
```
Same as above. Need a space for formatting as a list (see preview)

cp2004 (changes requested):
Just a review on the content of the page, I will leave plugin review alone and just commenting on the formatting.

Updated the preview, at least the images are showing right - my two suggestions are just to polish it off.

_plugins/BLOCKS.md
HugoCLSC: Alright, just changed it

cp2004 (approved):
Alright, looks good from page content perspective - plugin review still to come.

HugoCLSC:
Great! Looking forward to it.

jneilliii:
I did take an initial look at your plugin and there is a concern on the effect of loading in bootstrap 5 from a CDN. Not only will that not work if the device is not connected to the internet (some people run OP isolated from the internet and access is only local) but may cause some havoc with the underlying bundled bootstrap. 

I'll try to get to a deeper review of the plugin this week just to see if there is anything else that stands out, but that was one that jumped out very quickly.

HugoCLSC:
You think i can correct that by downloading the source files of Bootstrap v5 and then just store them in my assets folder?

jneilliii:
Messing with the UI is fine, but not at the expense of messing up other plugins or core libraries. If you want to completely replace the UI that is also possible utilizing the UIPlugin mixin, which would also require handling the websocket connection for data and remove the ability to have plugin integration unless you also incorporate that functionality. Your method of replacing components is totally acceptable, and is how I do the PlotlyTempGraph replacing the built-in temperature tab. 

If your plugin requires the use of the newer bootstrap is it just css styling classes that aren't part of the bundled bootstrap? If so, maybe just incorporating the missing parts would suffice.

HugoCLSC:
Yes that does seem to be a way to fix the bootstrap problem, i'll give that a try and just add what i need from Bv5. 
For now the UIPlugin is in a shelve for me, might look into that in the near future.

bell47 (commented):
Tidak buruk untuk plugin yang lebih baik

_plugins/BLOCKS.md
```diff
@@ -0,0 +1,96 @@
...
+
+compatibility:
+
+  octoprint:
+  - 1.7.2
+
+  os:
+  - linux
+
+  python: ">=2.7,<4"
```
jneilliii, line 53: Python compatibility does not match plugin's definition. This should be `">=3.3, <4"` here and in init.py or you need to remove the python_requires from [here](https://github.com/HugoCLSC/OctoPrint-BLOCKS/blob/a36374821b27e8523ec754a7f96fc1cf7c1ffdfa/setup.py#L62).

jneilliii (changes requested):
So it seems there is one small python version mismatch noted which needs to be resolved. It also appears as if you are now bundling the bootstrap 5 library. I still think that could have some adverse effects to other plugins, but I assume the point of your plugin is that it's only yours that will be running?

_plugins/BLOCKS.md
HugoCLSC: Yes the idea is to run the plugin stand alone, and continue to work on it in case there is need for new features. 
Alright just changed the python version on the .md file and on init.py.

jneilliii:
One last thing @HugoCLSC. After installing this plugin to test and verify I noticed that you are info logging on every network fetch and wifi status check. This can lead to flooding the octoprint.log, and I highly recommend switching these to `self._logger.debug` instead of `self._logger.info` to prevent log bloat. You can always change your log level for the plugin in the logging section and then get those messages when needed.

HugoCLSC:
I see what you mean, i'll apply those changes asap.

HugoCLSC:
Alright i've changed the logging levels as recommended.

jneilliii (approved):

PR #939, opened 2021-08-30
Safety Printer Plugin

SinisterRj:


jneilliii:
@SinisterRj one thing I see right off the bat is your plugin's repo has `__pycache__` folder. I don't think that should be included.

SinisterRj:
> 
> 
> @SinisterRj one thing I see right off the bat is your plugin's repo has `__pycache__` folder. I don't think that should be included.

Ok, fixed!

jneilliii (approved):

PR #949, opened 2021-10-05
Added plugin psucontrol_mqtt

oerkel47:
Subplugin for PSU Control to control generic mqtt switches


jneilliii:
With python 3 only plugin, we also recommend setting [this](https://github.com/oerkel47/OctoPrint-PSUControl-MQTT/blob/82a15ed43231dc1901f058856cc751f0d454438e/setup.py#L33) line to
```
additional_setup_parameters = {"python_requires": ">=3, <4"}
```
to prevent attempts at installing directly from github in older python environments.

I would also recommend mentioning on the plugin repo entry the requirement for the MQTT plugin as well. Will help you in the long run with support requests.

oerkel47:
Thanks, good call on the mentioning of mqtt plugin. Made the requested adjustments.

kantlivelong:
Haven't looked yer but you could add a banner in the plugin settings similar to how it checks for PSUControl being installed/enabled.

oerkel47:
> 
> 
> Haven't looked yet but you could add a banner in the plugin settings similar to how it checks for PSUControl being installed/enabled.

What do you mean by banner, something like this?

```
if not psucontrol_helpers or 'register_plugin' not in psucontrol_helpers.keys():
            self._logger.warning("The version of PSUControl that is installed does not support plugin registration.")
            return
```

kantlivelong:
Oh you don't have one at all. See https://github.com/kantlivelong/OctoPrint-PSUControl-TPLink/blob/7b1fc8717802e182206c05ad60047b835e930490/octoprint_psucontrol_tplink/templates/psucontrol_tplink_settings.jinja2#L1

Not required but it helps reduce support noise.

oerkel47:
Thats a cool feature which I just copied. Thank you!

oerkel47:
I also updated the description and name of my other plugin to make the two more distinctable due to the similiar names and focus on mqtt and psucontrol.

jneilliii (approved):

PR #959, opened 2021-11-11
Add OctoPrint-CrealityCloud Plugin

crealitycloud:


_plugins/OctoPrint-CrealityCloud.md
```diff
@@ -0,0 +1,56 @@
...
+  - linux
+  - windows
+  - macos
+  - freebsd
+
+  python: ">=2.7,<4"
+---
+* # OctoPrint-Crealitycloud
+
+  A OctoPrint-Plugin that connect the Octoprint to Creality Could
```
jneilliii, line 39: typo? Could instead of Cloud.

_plugins/OctoPrint-CrealityCloud.md
```diff
@@ -0,0 +1,56 @@
...
...643934386465392e706e67266f726967696e4865696768743d383536266f726967696e57696474683d31323338266f726967696e616c547970653d62696e61727926726174696f3d312673697a653d3435323636267374617475733d646f6e65267374796c653d6e6f6e65267461736b49643d7561653039623264332d373838632d343132622d393362342d66333964343862396632392677696474683d333130)](https://camo.githubusercontent.com/1f27dc04799d2ed0f52a741ba865eec6100faea48a8cd715c3fc7a7ed6076987/68747470733a2f2f63646e2e6e6c61726b2e636f6d2f79757175652f302f323032312f706e672f32323739353335362f313633323634373430373435322d66313961313237342d623131342d343636392d613636662d6361656532643430646130652e706e6723636c69656e7449643d7536373365356333652d643166662d342666726f6d3d64726f70266865696768743d3231342669643d756362353139636235266d617267696e3d2535426f626a6563742532304f626a656374253544266e616d653d34623637323032382d313935652d343638332d616135392d6539643663643934386465392e706e67266f726967696e4865696768743d383536266f726967696e57696474683d31323338266f726967696e616c547970653d62696e61727926726174696f3d312673697a653d3435323636267374617475733d646f6e65267374796c653d6e6f6e65267461736b49643d7561653039623264332d373838632d343132622d393362342d66333964343862396632392677696474683d333130)
```
jneilliii, line 54: please include all images as part of your PR in the folder /assets/img/plugins/crealitycloud/ instead of referencing them from an outside source for GDPR related concerns.

jneilliii (changes requested):
While reviewing the plugin I notice that the plugin is configured for python 2 compatibility but the `aliyun-iot-linkkit` module defined in plugin_requires is only python 3 compatible based on pypi's information so I think your plugin needs to be python 3 only. 

The software update hook is using the wrong user [here](https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/0ead7d5b3b3e8066b601ad624d8886b987ab565d/octoprint_crealitycloud/__init__.py#L91) it should match the GitHub user for the repo that's being reviewed/distributed, ie `crealitycloud`. 

While also reviewing your code it seems that you are including system files that are compiled binaries and cannot be reviewed. I believe this would be a concern for some people and therefore may need to get @foosel's input on the adoption of this plugin.

jneilliii:
You're still missing some things related to plugin_identifier in your submission @crealitycloud. The folder you put the images in needs to match your plugin's identifier `crealitycloud` from setup.py. So all images need to be in the folder `/assets/img/plugins/crealitycloud/`. Also your md file needs to also be named `crealitycloud.md` to match your plugin_identifier (case sensitive), as well as the `id` field in the `crealitycloud.md` file needs to be the same `crealitycloud`.

_plugins/crealitycloud.md
```diff
@@ -0,0 +1,68 @@
...
+
+  - https://github.com/SimplyPrint/OctoPrint-Creality2xTemperatureReportingFix/archive/master.zip
+  - https://github.com/OllisGit/OctoPrint-DisplayLayerProgress/releases/download/1.26.0/master.zip
+
+  ## Configuration
+
+  1、Make sure OctoPrint is up and running ,and Raspberry Pi is online. 
+  2、Make sure the above three plugins are installed.
+  3、In Octoprint select Setting->Crealitycloud Plugin->click botton to get bind code 
+  ![Configuration](/assets/img/plugins/OctoPrint-CrealityCloud/1.png)
```
jneilliii, line 62: forgot to update path, should be `/assets/img/plugins/crealitycloud/1.png`

_plugins/crealitycloud.md
```diff
@@ -0,0 +1,68 @@
...
+  4、In Creality Could APP select print->Add Device->choose Raspberry Pi 
+  ![Configuration](/assets/img/plugins/OctoPrint-CrealityCloud/2.png)
```
jneilliii, line 64: forgot to update path, should be `/assets/img/plugins/crealitycloud/2.png`

_plugins/crealitycloud.md
```diff
@@ -0,0 +1,68 @@
...
+  5、input code or scan QR Code 
+  ![Configuration](/assets/img/plugins/OctoPrint-CrealityCloud/3.png)
```
jneilliii, line 66: forgot to update path, should be `/assets/img/plugins/crealitycloud/3.png`

jneilliii (changes requested):
just a couple of minor issues here with the image paths as noted.

jneilliii (commented):
All my requested changes have been completed with the submission. It's now just a matter of @foosel reviewing since this has compiled binaries and such included.

crealitycloud:
Thank you for your patience.

foosel (changes requested):
I went through the plugin's source and sadly have found some issues...

  * Big issue: Large parts of the plugin functionality are implemented in form of intransparent shell scripts and other binaries that get spawned from python code. That is a no-go. It would be fine if you instructed the user to install certain well documented and publicy available third party video streaming services or similar and interfaced with those via documented APIs, with error handling that would guarantee graceful failing. The way things are done now is intransparent, prone to unrecoverable failure on even slight network hiccups, completely unaware of the fact that OctoPrint targets more than Linux/POSIX compatible systems on ARMv7 or x86/64, and operating outside of the boundaries of the OctoPrint plugin itself, which is a no-go. I'm sorry, but this needs to be changed for this plugin to have a shot at registration.
  * General issue: The code is littered with print statements used for debugging. Remove those.
  * General issue: There is next to no error handling present, especially when interacting with files or API endpoints
  * General issue: You need to make it crystal clear on the plugin settings page and the listing that this requires your cloud to function, that user data will be exchanged, and where to find the privacy policy for your cloud service.
  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/5713934183db54373e87a5c21bf3353949cd8e1d/setup.py#L21
    Description should be set
  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/5713934183db54373e87a5c21bf3353949cd8e1d/setup.py#L46
    1. You forgot to include these in MANIFEST.in as instructed by the comment 
    2. These binaries are something you'd need to also provide the sources of, or you'd need to make this VERY clear in your README and the plugin description that this plugin relies on CLOSED SOURCE COMPONENTS that the user has to trust.
  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/5713934183db54373e87a5c21bf3353949cd8e1d/octoprint_crealitycloud/__init__.py#L13-L19
    You forgot to remove this
  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/master/octoprint_crealitycloud/__init__.py#L132
    Do not call arbitrary OS functions, least of all without any meaningful error handling and/or making sure you are actually running under the correct OS
  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/5713934183db54373e87a5c21bf3353949cd8e1d/octoprint_crealitycloud/__init__.py#L149-L168
    This has to be as performant as possible as it will be caused for every single sent line. It currently is anything but as you are doing multiple full text searches on every single call. Use the `gcode` parameter.

    ``` python
    if gcode == "M220":
      # handle further
    elif gcode == "M27":
      # handle further
    ```

    Your `M27` handling currently doesn't even do anything and thus should most likely be removed altogether, meaning this results in

    ``` python
    if gcode == "M220":
      # handle futher
    ```

    Furthermore, under *no* circumstances must you do any kind of potentially blocking I/O processing here, like the M220 data upload. Enqueue this to be handled in a separate thread, NOT in the send loop, you will block the print processing and make fast     prints stutter.
  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/5713934183db54373e87a5c21bf3353949cd8e1d/octoprint_crealitycloud/__init__.py#L171-L189
    Same as before, this needs to be as performant as possible. if/if should be if/elif. The tons of text processing should ideally be reduced to a single precompiled regular expression.

    It also doesn't make much sense parsing standard protocol messages that OctoPrint already parses and processes itself, you probably can get rid of your own receive handler altogether by hooking into the progress and file selection reporting of the     printer interface.
  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/5713934183db54373e87a5c21bf3353949cd8e1d/octoprint_crealitycloud/perpetual_timer.py#L4
    PerpetualTimer looks very much like it just completely reimplements the OctoPrint utility class `octoprint.util.RepeatedTimer`. You should probably not reinvent the wheel but use what's already there. Also: get rid of the debug print statements.
  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/5713934183db54373e87a5c21bf3353949cd8e1d/octoprint_crealitycloud/crealityprinter.py#L151-L161
    What for? This is potentially personally identifiable information that you should make VERY CLEAR in your plugin listing & a LINKED PRIVACY POLICY that you are collecting this.
  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/5713934183db54373e87a5c21bf3353949cd8e1d/octoprint_crealitycloud/cxhttp.py#L9-L10
    If this is still supposed to be a secret, now it isn't anymore, so you probably should revoke these
  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/5713934183db54373e87a5c21bf3353949cd8e1d/octoprint_crealitycloud/config.py#L14
    Don't pollute the logging namespace please, "octoprint.plugins.crealitycloud.config"
  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/5713934183db54373e87a5c21bf3353949cd8e1d/octoprint_crealitycloud/filecontrol.py#L15
    Do *not* create secondary instances of things like file manager, get the existing ones injected instead. Even if you replicate object creation as it is now, nothing ensures you'll keep it up to date with future versions of OctoPrint, and that might cause file storage corruption for uses that stays even if your plugin is removed. Also do *not* create a second analysis queue, there's absolutely no reason to have two of it running.

lllhy9428:
All binary files and functions are deleted and some codes are optimized

lllhy9428:
Hello！Have you forgotten this plugin?@jneilliii @foosel

foosel:
Not forgotten, I never saw the notification that you recently commented and actually was under the impression that YOU had forgotten about this PR given that I never saw any kind of reaction to my review ;) Glad to know this is not the case :)

I'm currently swamped with other tasks, but I'll see that I can squeeze in another review ASAP, unless @jneilliii wants to?

lllhy9428:
Thank you : )

foosel (changes requested):
Thank you for your patience and for taking care of so many of the aforementioned points! Things are looking much better now. Sadly I still see some issues in your plugin that need to be addressed before it can get published on the repository. Especially the utter lack of a privacy policy is a grave concern here as it is with every cloud plugin. But there are also several code quality issues to address:

  * General issue: You need to make it crystal clear on the plugin settings page and the listing that this requires your cloud to function, that user data will be exchanged, and where to find the privacy policy for your cloud service.
    
    I've seen your plugin upload:
      * the internal IP of the interface via which the default gateway is reached (for which a ping is sent to Google's DNS server without the user's consent and for which alternatives exist in `octoprint.util.net`),
      * the full file list, 
      * the name of the current file,
      * printer telemetry,
      * something called `video` which I'm not even sure yet what it is
    
    I understand that at least everything apart from the first item in this list is needed to implement remote control, however this should still be made transparent to the user in form of a privacy policy. And I'd still like to know what the purpose is of the internal IP address.

  * General issue: There's still a signficant lack of error handling happening, especially where integer conversion and network communication takes place. E.g. in your CrealityPrinter implementation, and also in the CrealityCloud class. Stuff can go wrong, you need to anticipate that and handle it gracefully without the user's instance getting flooded with exceptions caught by OctoPrint's plugin layer originating from your plugin.

  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/b7f4c535ea22ef0253081846c5bed02e2b245007/octoprint_crealitycloud/__init__.py#L83 & https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/b7f4c535ea22ef0253081846c5bed02e2b245007/octoprint_crealitycloud/__init__.py#L86
    Your repository is called `OctoPrint-CrealityCloud`, not `OctoPrint-Crealitycloud`. Should match.

  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/b7f4c535ea22ef0253081846c5bed02e2b245007/octoprint_crealitycloud/__init__.py#L95
    Remove css entry, the CSS file is empty.

  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/b7f4c535ea22ef0253081846c5bed02e2b245007/octoprint_crealitycloud/__init__.py#L99-L118
    This should probably have some form of access control, most likely ADMIN access. Otherwise any registered user can just happily reconfigure the plugin.

  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/b7f4c535ea22ef0253081846c5bed02e2b245007/octoprint_crealitycloud/__init__.py#L120-L136
    This should maybe also have some form of access control?

  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/b7f4c535ea22ef0253081846c5bed02e2b245007/octoprint_crealitycloud/__init__.py#L156-L161
    I think you want to use https://docs.python.org/3/library/stdtypes.html#str.split instead of reinventing the wheel here. Also, OctoPrint already parses these messages from the printer and evaluates the position and total size to be used in its progress reports, you are reinventing the wheel here instead of using data that's already parsed and available. You might want to take a look at the progress reports available on the PrinterCallback you implemented again.

  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/b7f4c535ea22ef0253081846c5bed02e2b245007/octoprint_crealitycloud/crealitycloud.py#L64-L66
    Why create a repeated timer that you then cancel after one run? Also a property get access should never have a side effect, and your's here not only opens up a connection to the Google DNS server but also uploads data to your cloud endpoint. That's just dirty code.

  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/b7f4c535ea22ef0253081846c5bed02e2b245007/octoprint_crealitycloud/crealitycloud.py#L52
    If an SD card is detected, OctoPrint will itself already regularly query the SD printing status as needed. There's no reason for you to reimplement behaviour here and spam the serial line. Also, `M27C` is not a GCODE and just blindly sending it to unsuspecting printers here might cause severe problems for users but at the very least won't do anything but spam the serial line. You probably meant `M27 C`? Do you really need that on every single loop? I would imagine the file name won't change after selection for printing. In any case, this command will likely not be supported by every printer out there, so make sure you check for related capability reports before sending arbitrary commands that are not yet widely available, some firmware out there might crash on seeing something like this. 

  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/b7f4c535ea22ef0253081846c5bed02e2b245007/octoprint_crealitycloud/crealitycloud.py#L120-L129
    This will make the whole file get parsed fully every 2s if there's no `TIME` comment in there, or if the structure of the line causes your int conversion to fail thanks to an exception (no error handling either). Excessive resource usage is the consquence. If you absolutely need to (instead of just using what OctoPrint already provides you with and which also might have gotten enhanced by further plugins such as PrintTimeGenius) then fetch the print time rendered into the GCODE file *once* on print start and reuse that, and make sure to fail gracefully in case of GCODE files that don't contain the metadata encoded in comments that you expect here (not everyone uses the exact same slicer as you).

  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/b7f4c535ea22ef0253081846c5bed02e2b245007/octoprint_crealitycloud/crealitycloud.py#L300
    Should be a constant too, `Events.STARTUP`

  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/b7f4c535ea22ef0253081846c5bed02e2b245007/octoprint_crealitycloud/crealitycloud.py#L387-L410
    Launching arbitrary shell scripts that are not even longer located in your repository is a no go. However, as far as I can see this code was simply forgotten to be removed during the great shell script purge, so please do that now.

  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/b7f4c535ea22ef0253081846c5bed02e2b245007/octoprint_crealitycloud/config.py#L6
    You might want to fix that typo

  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/b7f4c535ea22ef0253081846c5bed02e2b245007/octoprint_crealitycloud/crealityprinter.py#L84-L86
    I'm not sure what you are trying to accomplish there, but it looks like something that should be done a different way instead of messing around with `self.__dict__` when you have perfectly fine attributes to use.

  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/b7f4c535ea22ef0253081846c5bed02e2b245007/octoprint_crealitycloud/crealityprinter.py#L157-L159
    This does not do what you expect it to do here. Sending `M114` happens asynchronously, as does receiving the response. So you need to actually wait for that (e.g. via the `PositionUpdate` event). By immedaitely using the current position data, you are actually using stale position data.

  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/master/octoprint_crealitycloud/static/js/qrcode.min.js
    There already is a QR Code library available in OctoPrint which even is usable through a knockout binding `qrcode`. Example usage see https://github.com/OctoPrint/OctoPrint/blob/1d4043736774dd9b4ea3eb8381100a38d8943fd4/src/octoprint/templates/dialogs/usersettings/access.jinja2#L38

  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/tree/master/octoprint_crealitycloud/static/less & https://github.com/crealitycloud/OctoPrint-CrealityCloud/tree/master/octoprint_crealitycloud/static/css
    Unused, should be deleted.
---

I'd also like to point that out that I noticed you are recommending to install [a plugin that fixes your printer firmware's behaviour](). I'd like to suggest again to not only rely on workarounds provided by the community but to actually take care that the firmware you ship with your printers is in working condition. We are still seeing a *ton* of issues in the community from users with Creality stock firmware shipped with their newly bought printers, from temperature parsing issues, to backpowering problems and outright refusal to communicate on serial if either an SD card is in the printer or if one is not in the printer, depending on firmware version apparently. So please forward this internally that there are still issues with stock firmware shipped with Creality printers (which I've also already repeatedly raised with Creality).

HaoyangLe:
By the way, how can i  use the current position data. I can't find the relevant code.

HaoyangLe:
About M27: Connecting the printer while the printer is printing to the SD card does not get information about the print job. How can I get relevant information in this situation?

foosel:
> 2\. Regarding print progress and print time: Because we print the file immediately after downloading, even though we prioritize the progress reports available on the PrinterCallback, the time to wait for octoprint analyze gcode files to complete is obviously too long, and we have to reinvent the wheel

Ah, thank you for clarifying. Be aware though that not all Slicers embed the information, and even if they do, the format differs widely. You should definitely prevent parsing the whole file looking for one specific format that you expect. Fail early, try multiple formats, be prepared to not find anything that you know.

> By the way, how can i use the current position data? I can't find the relevant code.

Subscribe to the PositionUpdate event. That will get fired whenever OctoPrint sees a position report from the printer, either triggered through a `M114` or if autoreporting of positions is enabled also that.

> About M27: Connecting the printer while the printer is printing on the SD card does not get information about the print job. How can I get relevant information in this situation?

So first of all, many printers out there still reset when connecting to them, so usually connecting to a printer currently printing from SD card is something not recommended to not risk that. Then: OctoPrint will detect an ongoing SD card job when it sees it get started (via the output that generates on the serial line). There's no way to retrieve the selected file from the printer (at least no universally supported one), so OctoPrint cannot track prints that haven't been started while it was connected as it cannot retrieve all information it requires for that. It might be possible to at least allow minimal monitoring by showing a progress report here with an "unknown" file, but that would require changes in OctoPrint.

Long story short, OctoPrint currently only supports monitoring prints started either through it or while it was connected to the printer. For everything else it lacks the necessary information to do its job properly, and it's also risky depending on the printer to connect to it while it's already working.

HaoyangLe:
Thank you for your patience. At present, I have fixed all the items in the list except for access control, because these two interfaces are connected to setting.js, and other roles except administrators have no access to it.
- [https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/b7f4c535ea22ef0253081846c5bed02e2b245007/octoprint_crealitycloud/__init__.py#L120-L136](url)
- [https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/b7f4c535ea22ef0253081846c5bed02e2b245007/octoprint_crealitycloud/__init__.py#L99-L118](url)

foosel:
> because these two interfaces are connected to setting.js, and other roles except administrators have no access to it.

Limiting access only client side isn't sufficient, it's trivial to bypass this completely and go against the backend directly, e.g. `curl http://octopi.local/plugin/crealitycloud/status` or `curl -X POST http://octopi.local/plugin/crealitycloud/get_token`

HaoyangLe:
OK, I just add it.

foosel:
Alright. Will take another look ASAP, but be aware that I'm currently in the process of finalizing and preparing a first release candidate for the next stable version of OctoPrint, and that will probably take up most of my time next week, so it might be late next week or even the week after until I get around to another review.

foosel:
Almost done, thank you for all the work you did in working those review notes in! Only some small open points that need addressing (the first of which I only now spotted, sorry for that):

  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/945ad023bcc5ed23511ab38be512d180cf0b35c4/setup.py#L36
    Pinning the dependencies like this might cause incompatibility issues with other plugins. I know for a fact that the MQTT plugin for example also depends on `paho-mqtt`. If that now pins (or ranges) a version that is not 1.5.1, either your plugin will fail to install or nuke the other in the process. In general I'd strongly suggest to code against version *ranges* of third party dependencies (as wide as possible), to avoid exactly these kind of issues to arise.
  * You are still sending `M27C` instead of [`M27 C`](https://reprap.org/wiki/G-code#M27:_Report_SD_print_status). Whitespace matters.
  * Thank you for adding the privacy statement to the settings page. As stated before we'll also need this information on the plugin listing itself (so added to the files of this PR here). You also should provide a link so that people can take a look at the policy *before* installing any apps or plugins.
  * The latest release of your plugin on GitHub is still 0.5.7, the latest code states it's 0.6.0. Be sure to create a release whenever you change the version number or people will install versions that are further ahead than what the software update plugin can pick up on.

There's also some stuff I noticed that you might want to correct as well, though I'd classify this as FYI/nice-to-haves:

  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/945ad023bcc5ed23511ab38be512d180cf0b35c4/setup.py#L36
    You are depending on OctoPrint 1.3.8+, but your plugin is Python 3 only, which only is supported since OctoPrint 1.4.0. So, `>=1.4.0` would probably make more sense here. I'd rather suggest however to go with the version you've actually tested it against though as the lower range boundary here, so if you've mostly tested it against OctoPrint 1.7.0, go with `>=1.7.0`. Not actually an issue, just a suggestion.
  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/945ad023bcc5ed23511ab38be512d180cf0b35c4/octoprint_crealitycloud/__init__.py#L53-L60
    This definition is a left-over you should remove. It gets overwritten by the correct implementation further below.
  * The way that Python's `with open...` works is, it will open the file for you, return the handle to you, and once you leave the `with` block the file will be closed. So the `f.close()` statements at the end of various of your `open` routines are unnecessary and can be removed.
  * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/945ad023bcc5ed23511ab38be512d180cf0b35c4/octoprint_crealitycloud/templates/README.txt
    Can be removed

HaoyangLe:
- About PAHO mqtt: if the version is greater than 1.5.1, it will lead to catastrophic bugs. Now the version requirement has been set to < = 1.5.1


- About M27 and m27c: the current policy is to send once when the printer is connected. After receiving the reply, if it is printing, it will stop when the printing is completed or stopped. If the printer is not printing, it will stop sending only once. Used to improve the problem that octoprint cannot obtain the printer status when connecting to the printer while the printer is printing

- About privacy agreement: where does PR file refer?

foosel:
> * About PAHO mqtt: if the version is greater than 1.5.1, it will lead to catastrophic bugs.

Why? If that is down to your code or the dependency using it, you should fix that accordingly. Simply using an old version of a library is no long term solution, especially not if it pins that version for everyone installing your plugin also for other plugins.

> * About M27 and m27c:

I think there is a misunderstanding here. My problem is with you sending `M27C` instead of `M27 C`. The latter (with the added space between `M27` and `C`) [is specified in what we have in terms of a specification](https://reprap.org/wiki/G-code#M27:_Report_SD_print_status). The former, without the whitespace, is not. I'm not arguing about sending two commands, I'm arguing about the spelling of the second one, which is lacking a whitespace character.

> * About privacy agreement: where does PR file refer?

[See `_plugins/crealitycloud.md` in this here PR where we are having our conversation in](https://github.com/OctoPrint/plugins.octoprint.org/pull/959/files#diff-d72909abf5cea3fbdf44b67a136c511f07a9d1790087a828f09564e2be32bcf7).

crealitycloud:
> Almost done, thank you for all the work you did in working those review notes in! Only some small open points that need addressing (the first of which I only now spotted, sorry for that):
> 
> * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/945ad023bcc5ed23511ab38be512d180cf0b35c4/setup.py#L36
>   Pinning the dependencies like this might cause incompatibility issues with other plugins. I know for a fact that the MQTT plugin for example also depends on `paho-mqtt`. If that now pins (or ranges) a version that is not 1.5.1, either your plugin will fail to install or nuke the other in the process. In general I'd strongly suggest to code against version _ranges_ of third party dependencies (as wide as possible), to avoid exactly these kind of issues to arise.
> * You are still sending `M27C` instead of [`M27 C`](https://reprap.org/wiki/G-code#M27:_Report_SD_print_status). Whitespace matters.
> * Thank you for adding the privacy statement to the settings page. As stated before we'll also need this information on the plugin listing itself (so added to the files of this PR here). You also should provide a link so that people can take a look at the policy _before_ installing any apps or plugins.
> * The latest release of your plugin on GitHub is still 0.5.7, the latest code states it's 0.6.0. Be sure to create a release whenever you change the version number or people will install versions that are further ahead than what the software update plugin can pick up on.
> 
> There's also some stuff I noticed that you might want to correct as well, though I'd classify this as FYI/nice-to-haves:
> 
> * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/945ad023bcc5ed23511ab38be512d180cf0b35c4/setup.py#L36
>   You are depending on OctoPrint 1.3.8+, but your plugin is Python 3 only, which only is supported since OctoPrint 1.4.0. So, `>=1.4.0` would probably make more sense here. I'd rather suggest however to go with the version you've actually tested it against though as the lower range boundary here, so if you've mostly tested it against OctoPrint 1.7.0, go with `>=1.7.0`. Not actually an issue, just a suggestion.
> * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/945ad023bcc5ed23511ab38be512d180cf0b35c4/octoprint_crealitycloud/__init__.py#L53-L60
>   This definition is a left-over you should remove. It gets overwritten by the correct implementation further below.
> * The way that Python's `with open...` works is, it will open the file for you, return the handle to you, and once you leave the `with` block the file will be closed. So the `f.close()` statements at the end of various of your `open` routines are unnecessary and can be removed.
> * https://github.com/crealitycloud/OctoPrint-CrealityCloud/blob/945ad023bcc5ed23511ab38be512d180cf0b35c4/octoprint_crealitycloud/templates/README.txt
>   Can be removed

Hello, I have fixed the above issues and updated to version 1.0.1.

jneilliii (approved):

foosel:
Thank you, can confirm that the points have been covered and I took care of adjusting your plugin description here to also point users to where they can find a privacy policy. Merging.

foosel:
It has come to our attention that there are allegations of content theft from prominent members of the community against your platform, as of yet unaddressed by you:

  * https://twitter.com/agepbiz/status/1514611716985790475

Due to this, we are removing this listing from the plugin repository again. Once these complaints have been resolved we will reconsider adopting the plugin into the official repository by reverting the above commit.

foosel:
Further handling of this situation to be done in #1048

PR #965, opened 2021-11-21
Add OctoApp Companion Plugin

crysxd:
This PR adds the OctoApp Companion plugin to the repository. I was not able to test the page with Jekyll due to some weird errors in Jekyll's startup...


cp2004:
Preview URL (manually updated): https://cp2004-plugins-octoprint-org-git-pr-965-cp2004.vercel.app/

crysxd:
Thanks! Added a second image, should be fine now :)

jneilliii (changes requested):
Your plugin's update hook needs to be fixed [here](https://github.com/crysxd/OctoApp-Plugin/blob/11e144877ff00d098f816c316a227969f2aa9564/octoprint_octoapp/__init__.py#L246) to be `octoapp` not `octolight` so it matches the plugin's identifier. 

I'm not 100% sure on this, but I think maybe all your threads should include `daemon=True` when initialized to avoid any kind of lock up issues with OctoPrint's shutdown. It probably would never effect your plugin, but just something that should be adjusted I think.

jneilliii:
@crysxd just noticed also that the repo is wrong in your update check as well [here](https://github.com/crysxd/OctoApp-Plugin/blob/11e144877ff00d098f816c316a227969f2aa9564/octoprint_octoapp/__init__.py#L254), should be `OctoApp-Plugin`.

crysxd:
- Fixed the `octolight`, must have missed it. Also double checked it's nowhere else mentioned
- I looked into the deamon flag bit and it seems you are right. Added the deamon flag. I don't think this would be a big issue as the thread will only finish the current network request, but it's cleaner this way for sure.
- Fixed the wrong plugin key

All changes from the plugin: https://github.com/crysxd/OctoApp-Plugin/commit/2bf9215d5d888cdc08b58eaad979e5b3d9d586f3

I also pushed a new commit changing the plugin version to include the above changes.


Thanks for the feedback!

crysxd:
And a small feature was added as well to allow the app to get the printer firmware data :) https://github.com/crysxd/OctoApp-Plugin/commit/e1da044c9280ce6e735e1e2047a1bc7321cb02ec

_plugins/octoapp.md
```diff
@@ -0,0 +1,117 @@
...
+- Christian Würthner
+#- second autor name
+license: AGPL v3
+
+# today's date in format YYYY-MM-DD, e.g.
+date: 2021-11-21
+
+homepage: https://play.google.com/store/apps/details?id=de.crysxd.octoapp&hl=en&gl=US
+source: https://github.com/crysxd/OctoApp-Plugin
+archive: https://github.com/crysxd/OctoApp-Plugin/archive/refs/tags/1.0.3.zip
```
cp2004, line 17: It would be better if this pointed to a branch or something - otherwise you will be doing PRs here for every update.

It can be pointed to the default branch using `https://github.com/crysxd/OctoApp-Plugin/archive/master.zip`

cp2004 (changes requested):

crysxd:
Change the archive to the release branch

cp2004 (approved):
My comments have been addressed so I guess I need to turn my review green 🙂

jneilliii (approved):

PR #974, opened 2022-01-02
Added plugin custom_gcode_events

RoboMagus:


jneilliii:
Unable to review plugin as repo https://github.com/RoboMagus/OctoPrint-Custom_gcode_events doesn't exist.

RoboMagus:
Ah damn, forgot to change it to public.
I changed that now.

jneilliii (changes requested):
Great utility function @RoboMagus and everything looks good except for a couple of minor things listed below.

Add description to setup.py [here](https://github.com/RoboMagus/OctoPrint-Custom_gcode_events/blob/a5b4e5490c33958115d46a8de279688daa1495d9/setup.py#L21). 

Add email to setup.py [here](https://github.com/RoboMagus/OctoPrint-Custom_gcode_events/blob/a5b4e5490c33958115d46a8de279688daa1495d9/setup.py#L27)

You might consider changing all your self._logger.info to self._logger,debug to reduce writes to octoprint.log unless needed. You can always enable debug logging for the plugin in OctoPrint's logging section and get more details when you need it.

We typically encourage plugin's to include a screenshot as part of the plugin listing. It's not required at all, but might be helpful to show the settings page like you did in your repo. If you do decide to add it make sure to include it in `/assets/img/plugins/custom_gcode_events/` folder. The featuredimage attribute in the template section at the top would include that in the new plugin notification, twitter posts, etc. Check other plugin's md files for examples, you can embed the image as part of your full description or add a screenshots list to the template section at the top. Again, totally optional, but encouraged.

RoboMagus:
Thanks for the feedback @jneilliii.
I've updated the setup.py fields as requested and reworked the least usefull log entries to debug (with the exception of the logging events when a gcode hook matches).

I left out a screenshot for the plugin listing on purpose, since in the near future the plugin configuration page might be revamped. Then the image would become outdated.
At the moment I don't see adding a screenshot as added value, so for now I'll leave that as is ;)

jneilliii:
Ok, no problem on the picture. If you ever do want to update the listing just send in a PR, those get merged fairly quickly for this specific reason. Thanks for the changes, it taught me about GitHub's private commit email address (although it doesn't actually seem to work for my account for some reason as the domain has no mx record).

jneilliii (approved):

PR #979, opened 2022-01-08
Added GPIOStatus and ShutdownButtonLEDBuzzer plugins

danieleborgo:


jneilliii:
[this](https://github.com/danieleborgo/OctoPrint-ShutdownButtonLEDBuzzer/blob/b3dde65599b1980823854392418ee92d260cc370/octoprint_shutdownbuttonledbuzzer/__init__.py#L238) needs to match your plugin_identifier, including case so they are equal. 

Both plugins are missing email in setup.py.

Would recommend adding `additional_setup_parameters = {"python_requires": ">=3, <4"}` to both plugin's setup.py to prevent people from side-loading the plugin in older python 2 environments.

danieleborgo:
> [this](https://github.com/danieleborgo/OctoPrint-ShutdownButtonLEDBuzzer/blob/b3dde65599b1980823854392418ee92d260cc370/octoprint_shutdownbuttonledbuzzer/__init__.py#L238) needs to match your plugin_identifier, including case so they are equal.
> 
> Both plugins are missing email in setup.py.
> 
> Would recommend adding `additional_setup_parameters = {"python_requires": ">=3, <4"}` to both plugin's setup.py to prevent people from side-loading the plugin in older python 2 environments.

Ok, I fixed these three issues. Thank you

jneilliii (approved):

PR #989, opened 2022-01-18
Add OctoPrint-PrintWatch plugin

printpal-io:
Adding the OctoPrint-PrintWatch plugin to the OctoPrint plugin repository.


jneilliii:
Am I correct that your service does not have a maker/free plan available at all?

printpal-io:
> Am I correct that your service does not have a maker/free plan available at all?

At the moment all we have is the free trial period.

jneilliii:
The only thing that I see that stands out is that your get_update_information hook does not include the necessary bits for actually performing updates. It's missing user, repo, and pip at minimum. And the returned dictionary needs to be the same as the plugin_identifier (case sensitive) so the line [here](https://github.com/printpal-io/OctoPrint-PrintWatch/blob/6a836ae5fcf4f0b0bd81ee93e86b292a489ce13c/octoprint_printwatch/__init__.py#L30) should be `printwatch=dict(`. 

Outside of that, I am going to request @foosel to review this one due to the nature of the plugin and make sure I didn't miss anything.

printpal-io:
> The only thing that I see that stands out is that your get_update_information hook does not include the necessary bits for actually performing updates. It's missing user, repo, and pip at minimum. And the returned dictionary needs to be the same as the plugin_identifier (case sensitive) so the line [here](https://github.com/printpal-io/OctoPrint-PrintWatch/blob/6a836ae5fcf4f0b0bd81ee93e86b292a489ce13c/octoprint_printwatch/__init__.py#L30) should be `printwatch=dict(`.
> 
> Outside of that, I am going to request @foosel to review this one due to the nature of the plugin and make sure I didn't miss anything.

Updated the plugin update hook to include github and pip in addition to your fix.

printpal-io:
@jneilliii Update on the review process?

foosel:
This review is waiting for me. And I'm only back from a break since last week, spent the whole last week playing fire fighter on some compatibility issues that suddenly arose in the whole OctoPrint world, and this week am now slowly but steadily working through my break backlog, which this review here is also a part of. So please, some more patience :)

printpal-io:
> This review is waiting for me. And I'm only back from a break since last week, spent the whole last week playing fire fighter on some compatibility issues that suddenly arose in the whole OctoPrint world, and this week am now slowly but steadily working through my break backlog, which this review here is also a part of. So please, some more patience :)

Great. Thank you! Take your time :)

foosel (changes requested):
Found a couple points that should be addressed:

  * Plugin URL doesn't match repo: https://github.com/printpal-io/OctoPrint-PrintWatch/blob/917fc559b34d491e863a8ab97fe6174ab0127953/setup.py#L30
  * Archive URL for updates points to master branch instead of releases, should rather be `https://github.com/printpal-io/OctoPrint-PrintWatch/archive/{target_version}.zip` or you risk pulling in unreleased changes: https://github.com/printpal-io/OctoPrint-PrintWatch/blob/917fc559b34d491e863a8ab97fe6174ab0127953/octoprint_printwatch/__init__.py#L38
  * stream_url only works on OctoPi default config, maybe read from master global instead? https://github.com/printpal-io/OctoPrint-PrintWatch/blob/917fc559b34d491e863a8ab97fe6174ab0127953/octoprint_printwatch/__init__.py#L46
  * Threads need [daemon=True](https://docs.python.org/3/library/threading.html#threading.Thread.daemon), e.g. inferencer & videostreamer
  * Settings tab & plugin description should contain a link to privacy policy given that this is a cloud service with data exchange

printpal-io:
> Found a couple points that should be addressed:
> 
> * Plugin URL doesn't match repo: https://github.com/printpal-io/OctoPrint-PrintWatch/blob/917fc559b34d491e863a8ab97fe6174ab0127953/setup.py#L30
> * Archive URL for updates points to master branch instead of releases, should rather be `https://github.com/printpal-io/OctoPrint-PrintWatch/archive/{target_version}.zip` or you risk pulling in unreleased changes: https://github.com/printpal-io/OctoPrint-PrintWatch/blob/917fc559b34d491e863a8ab97fe6174ab0127953/octoprint_printwatch/__init__.py#L38
> * stream_url only works on OctoPi default config, maybe read from master global instead? https://github.com/printpal-io/OctoPrint-PrintWatch/blob/917fc559b34d491e863a8ab97fe6174ab0127953/octoprint_printwatch/__init__.py#L46
> * Threads need [daemon=True](https://docs.python.org/3/library/threading.html#threading.Thread.daemon), e.g. inferencer & videostreamer
> * Settings tab & plugin description should contain a link to privacy policy given that this is a cloud service with data exchange

Made and committed all of the changes requested except for the stream_url with master global, which is giving issues. The stream_url is just a placeholder value in this case and the user can change it in the settings page to match their actual stream. Will add this as a work item for the future.

jneilliii:
> The stream_url is just a placeholder value in this case and the user can change it in the settings page to match their actual stream.

fairly easy...

```
    def get_settings_defaults(self):
        return dict(
            stream_url = self._settings.global_get(["webcam", "stream"]),
```

printpal-io:
> > The stream_url is just a placeholder value in this case and the user can change it in the settings page to match their actual stream.
> 
> fairly easy...
> 
> ```
>     def get_settings_defaults(self):
>         return dict(
>             stream_url = self._settings.global_get(["webcam", "stream"]),
> ```

I am getting the following error when trying to global_get the webcam url:

2022-01-27 17:13:01,489 - octoprint.server.api.settings - ERROR - Could not load settings for plugin PrintWatch (1.0.10)
  File "/home/pi/oprint/lib/python3.7/site-packages/octoprint/server/api/settings.py", line 353, in _get_plugin_settings
    result = plugin.on_settings_load()
  File "/home/pi/oprint/lib/python3.7/site-packages/octoprint/util/__init__.py", line 1890, in wrapper
    return f(*args, **kwargs)
  File "/home/pi/oprint/lib/python3.7/site-packages/octoprint/plugin/types.py", line 1585, in on_settings_load
    data = copy.deepcopy(self._settings.get_all_data(merged=True))
AttributeError: 'NoneType' object has no attribute 'get_all_data'

That is why the placeholder is being used for the time being.

cp2004:
To explain that - getting the default settings is part of loading OctoPrint's settings system, so you can't start accessing the settings before it's loaded.

As long as you are aware it won't work for every install and are prepared to make users configure it, I would say that's OK. We've definitely seen before people hardcoding OctoPi or certain OS specific stuff without any sort of handling for if it goes wrong.

printpal-io:
> 

I am aware that it won't work for every install and have documentation for users to configur

printpal-io:
Accidentally closed this pull request, my apologies.

jneilliii:
> To explain that - getting the default settings is part of loading OctoPrint's settings system, so you can't start accessing the settings before it's loaded.

Oh right, wasn't thinking that one through...my bad. I agree though, if you are defaulting the setting might as well use what the majority would be, and in this case I do think that is octopi installs based on the stats.

foosel (approved):
Happy to give this an approval now.

Still something for you to look at and maybe re-evaluate though:

[In your update information](https://github.com/printpal-io/OctoPrint-PrintWatch/blob/main/octoprint_printwatch/__init__.py#L38) you now use `https://github.com/printpal-io/OctoPrint-PrintWatch/archive/release.zip` as download URL for your updates. That STILL points to a branch that you now have to keep up-to-date with your latest release. Instead, I strongly suggest to simply use GitHub's release mechanism to create releases off your main branch and then use `https://github.com/printpal-io/OctoPrint-PrintWatch/archive/{target}.zip` (or `.../{target_version}.zip`) as your update URL - that way, if you do update checks via GitHub releases (which you have configured here), OctoPrint will take the release tag you created and use that for `{target}` here, which will always resolve to the tagged source tarball and hence the exact code at the point of release.

PR #991, opened 2022-01-22
Google Drive Files

### LINKED ISSUE #2222: [Request] Direct Access to Google Drive
Hi, thank you for such an incredible app,

I am aware of a cloud plugin existing in octoprint called Polar Cloud which also indirectly support google drive.

the reason why I'm requesting this feature is that if I am using a slicer on a different network I could have a specific folder on my google drive as my upload folder on my octoprint which I can upload gcodes to it.

I know I can access octoprint via the internet to which I can just drag and drop but if it is possible to take out that one extra step it would be relatively more convenient.

thank you :)

jneilliii:
register plugin Google Drive Files, closes https://github.com/OctoPrint/OctoPrint/issues/2222


jneilliii:
come on bot, I didn't touch that file...

_plugins/googledrivefiles.md
```diff
@@ -0,0 +1,61 @@
...
+- jneilliii
+license: AGPLv3
+
+date: 2022-01-22
+
+homepage: https://github.com/jneilliii/OctoPrint-GoogleDriveFiles
+source: https://github.com/jneilliii/OctoPrint-GoogleDriveFiles
+archive: https://github.com/jneilliii/OctoPrint-GoogleDriveFiles/archive/master.zip
+
+tags:
```
kantlivelong, line 17: These might not be the best terms for this plugin :smile:

kantlivelong (changes requested):
Hmm you're repo def hes some other commits in it or something? Rebase and force push maybe?

Other than that looks like just the tags need adjusting.

jneilliii:
Yeah, I was fixing stuff that foosel appears to have been fixing upstream around the same time. Should be cleaned up now.

kantlivelong (approved):

PR #992, opened 2022-01-22
Active Connections

### LINKED ISSUE #527: Connection Indicator
It would be a nice feature to have an indicator telling us how many active persons are connected to octoprint web. The reason why i ask this is that as i allow persons access my octoprint web watch my parts print, However i never can tell if someone is connected or not 

Actually i use very crude approach by manually start octoprint via putty and then if someone other than me connects i will see the following in the terminal (very crude)  

2014-07-30 21:29:21,952 - octoprint.server.util - INFO - New connection from client: 63.168.X .X

A cleaner approach would be awesome

jneilliii:
register active connections plugin, resolves https://github.com/OctoPrint/OctoPrint/issues/527


kantlivelong:
https://github.com/jneilliii/OctoPrint-ActiveConnections/releases :question:

jneilliii:
Does that matter for first release? Added.

kantlivelong:
> Does that matter for first release?

Honestly not sure if it would throw any errors when doing update checks but may as well do it for completeness.

Issue #983, opened 2022-01-16
[Abandoned Plugin] Octoprint-Smart-Filament-Sensor


drbrown-l3tech:
### Which Plugin?

https://plugins.octoprint.org/plugins/smartfilamentsensor/

### Link to ticket on plugin's issue tracker confirming abandonment

https://github.com/maocypher/Octoprint-Smart-Filament-Sensor

drbrown-l3tech:
The author has posted that it is abandoned on the bottom of the Readme.

jneilliii:
@Royrdan, few things I'll need from you. 

1. Enable issues on your fork of the repository.
2. Update the user [here](https://github.com/Royrdan/Octoprint-Smart-Filament-Sensor/blob/ce247f3f4ebb3e72e65dfe9ea95b9659ee1db3f2/octoprint_smart_filament_sensor/__init__.py#L348)
2. Update the url [here](https://github.com/Royrdan/Octoprint-Smart-Filament-Sensor/blob/ce247f3f4ebb3e72e65dfe9ea95b9659ee1db3f2/octoprint_smart_filament_sensor/__init__.py#L369).
3. Bump the version [here](https://github.com/Royrdan/Octoprint-Smart-Filament-Sensor/blob/ce247f3f4ebb3e72e65dfe9ea95b9659ee1db3f2/setup.py#L18) to `1.1.6`.
4. Create a new tag named `1.1.6` and release with the above changes included.

jneilliii:
@Royrdan I see you took care of 1-4, small correction needed. GitHub is case sensitive on the URL so make the line [here](https://github.com/Royrdan/Octoprint-Smart-Filament-Sensor/blob/bffb42fa6350539a523ee4db132235b5a56ad3c3/octoprint_smart_filament_sensor/__init__.py#L369) be `pip="https://github.com/Royrdan/Octoprint-Smart-Filament-Sensor/archive/{target_version}.zip"`, notice capital `R` in username. Also Go [here](https://github.com/Royrdan/Octoprint-Smart-Filament-Sensor/releases) click new release button in the upper right and enter 1.1.6 for the tag, give it a name and add any release notes you'd like to include about adopting the plugin. That will handle `#5`. Please reply here once complete so I know it's done.

Royrdan:
Hi jneilliii,
Thanks for your help with that.
It should all be completed now.
Let me know anything else.
Thanks

joshuacant:
[...]

PR #1002, opened 2022-01-31
Fix update URL for maintainer change

cp2004:
Overrides the pip url for downloading the update.

Bit of an experiment to be honest, but should work. This has come up a couple of times before but with the standard notice saying to 'reinstall the plugin' to get the new maintainer's version I guess it goes away on it's own.

Following up from https://github.com/OctoPrint/plugins.octoprint.org/issues/983#issuecomment-1026283233

I won't merge it straight away - it's late here so it would be good for someone else to double check the logic.

@Royrdan this will probably override it for your repo as well - if you ever need to change the pip URL, then let us know here and we'll sort it.

### DIFF
--- a/_data/update_check_overlays.yaml
+++ b/_data/update_check_overlays.yaml
@@ -41,3 +41,4 @@ HeaterTimeout:
   user: Andy-ch
 smartfilamentsensor:
   user: Royrdan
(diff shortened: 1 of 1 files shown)


kantlivelong:
Seems like it will work. Lets find out

PR #1005, opened 2022-02-09
Add CalibrationTools plugin

SergiuToporjinschi:
Adding CalibrationTools plugin.


cp2004:
What's the difference between this plugin and 'Calibration Companion'?

https://plugins.octoprint.org/plugins/calibrationcompanion/

Does your plugin do anything special/different or unique?

SergiuToporjinschi:
I did not know that there is another calibration plugin. But because you pointed on that, I can see that it does not having x-y-z steps calibration it has only steps for extruder.

SergiuToporjinschi:
Hi,
Just let me know if i have to change soething.
Thank you!
Have a nice day!

_plugins/CalibrationTools.md
```diff
@@ -0,0 +1,71 @@
...
+description: Marlin Calibration Tools
+authors: 
+  - Sergiu Toporjinschi
+license: GNUv3
+
+date: 2022-02-09
+
+homepage: https://github.com/SergiuToporjinschi/OctoPrint-CalibrationTools
+source: https://github.com/SergiuToporjinschi/OctoPrint-CalibrationTools
+archive: https://github.com/SergiuToporjinschi/OctoPrint-CalibrationTools/archive/master.zip
```
jneilliii, line 15: incorrect URL, should be `https://github.com/SergiuToporjinschi/OctoPrint-CalibrationTools/archive/master.zip`

_plugins/CalibrationTools.md
```diff
@@ -0,0 +1,71 @@
...
+
+compatibility:
+  octoprint: 
+   - 1.2.0
+  os:
+    - linux
+    - windows
+    - macos
+    - freebsd
+  python: ">=3,<4"
```
jneilliii, line 37: doesn't match what's defined in your init.py file [here](https://github.com/SergiuToporjinschi/OctoPrint-CalibrationTools/blob/82ede19af44d2086dfa578e0901393d5ee14f656/octoprint_CalibrationTools/__init__.py#L125).

_plugins/CalibrationTools.md
```diff
@@ -0,0 +1,71 @@
...
+![X-Y-Z Steps](/assets/img/plugins/CalibrationTools/featuredimage.png)
+
+![PID Autotune](/assets/img/plugins/CalibrationTools/PID-autotune.png)
+
+## Setup
+
+Install via the bundled [Plugin Manager](https://docs.octoprint.org/en/master/bundledplugins/pluginmanager.html)
+or manually using this URL:
+
+    https://github.com/SergiuToporjinschi/OctoPrint-CalibrationTools/archive/master.zip
```
jneilliii, line 67: same as above, incorrect URL.

jneilliii (changes requested):
Please see my comments and adjust accordingly.

_plugins/CalibrationTools.md
SergiuToporjinschi: Hi, Sorry but for me, the link posted by you is the same as I have in line 15 and for me both links are triggering an download of that archive.
or you are referring to the source url ? should get me to the source archive?

jneilliii: Oh sorry, I copied/pasted but didn't adjust. It should be `https://github.com/SergiuToporjinschi/OctoPrint-CalibrationTools/archive/main.zip`

jneilliii:
@SergiuToporjinschi I didn't know that GitHub was redirecting master.zip to the correct main.zip, but it's best to set it to what it actually is, which is based on your branch name `main`. I've added a comment above with corrected URL, sorry for the confusion.

SergiuToporjinschi:
> 
Yes, I think they just changed the name of the default branch and now they are just keeping a backward compatibility. But like you said is better if we are using the new naming.

jneilliii (approved):

PR #1016, opened 2022-02-22
add openmiko plugin

tmountjr:
Adding my new plugin to control IR LEDs (day and night modes) on webcams running openmiko firmware.


cp2004 (approved):
Quite a simple plugin so nothing wrong that I can see :)

Preview of the page looks good https://cp2004-plugins-octoprint-org-git-pr1016-cp2004.vercel.app/

jneilliii:
Wouldn't it be better to put [this](https://github.com/tmountjr/OctoPrint-Openmiko/blob/b8ef533b2d6adf1b0a550320e5fc38105840e83f/octoprint_openmiko/static/js/openmiko.js#L86-L89) bit at the top of getAdditionalControls? No reason to do any parsing and function creation at all if their is no stream I suspect.

tmountjr:
@jneilliii yeah, definitely - good catch. that line was a late addition and I didn't think to move it to the top. released v0.2.0 with that change (though nothing about this PR's file should change I think).

PR #1019, opened 2022-02-27
New Plugin - Github Token screen for software update plugin

NilsRo:
Created a simple plugin to set Github Token in Settings to avoid rate limit error in Software Update Plugin

Still asked in the forum how to delete the old commits in Pull Requests ;-) as my forked Repo seems in sync...


cp2004:
Does this really need to be a plugin? Why not a PR to core OctoPrint?

The token should not really be accessible over the settings API, to keep it secure - remember these tokens provide access to the GitHub account, so any user with access to OctoPrint could find it out.

It could simply be in the menu in OctoPrint's settings, a place where you can *set* the token but it can't be read back.

NilsRo:
Is there a guide what has to be taken care of if enhancing OctoPrint itself? Like plugins as they are easier to handle outside....

Would be very easy to have that behaviour. I can connect the knockout bind directly to the software updater settings. Then it would not show the token as it is not loaded. Added some code to get exactly that behaviour and see the token set in settings. :-) But sound useful to me, I am a private single user only.

cp2004:
> any special things to know or should link it via knockout directly (->very simple)?

The token is not exposed over the settings API so it's not available to link directly in KO. To try and keep it somewhat private (even though you probably could not call it *secure*), this should continue and it should be set-only in my opinion.

NilsRo:
But I can access it and it will behave like you described...ready only. Build it like this before...but at least it would be more consistent to add it to core octoprint as a new field in software updater.

NilsRo:
@cp2004 : created a demo (here)[https://github.com/OctoPrint/OctoPrint/pull/4443] for discussion but maintenance branch seems having problems as I cannot run it actually. Only master and some other branches are fine...but like to discuss it before.

PR #1030, opened 2022-03-17
Add plugin OctoPrint-Wyze

eshapiro42:


eshapiro42:
Just wanted to mention that this plugin only works with Python3.8+. That's because it relies on the `wyze_sdk` module which only works with newer Python versions.

That obviously severely limits who will be able to install it, since it looks like OctoPrint doesn't test on versions newer than Python3.7. But I figured some people might either already be running newer Python environments or find this useful enough to go the extra mile to use it.

Thanks!

jneilliii:
Highly recommend adding a [software update check hook](https://docs.octoprint.org/en/master/bundledplugins/softwareupdate.html#octoprint-plugin-softwareupdate-check-config) to your plugin at a minimum. This way as you release new versions the end users will get notices for upgrade automatically. 

You may want to also link to the reverse engineered wyze_sdk github and explain it may stop working if Wyze decides to make a change on their end just as a precaution.

Lastly, I notice on your GitHub you mention storing the password in plain text in config.yaml. You could add in some form of encryption on the password using hardware identifiers or `self._settings.global_get(["accessControl", "salt"])`. If you do go this route you would probably have to override [on_settings_save](https://docs.octoprint.org/en/master/plugins/mixins.html#octoprint.plugin.SettingsPlugin.on_settings_save) and [on_settings_load](https://docs.octoprint.org/en/master/plugins/mixins.html#octoprint.plugin.SettingsPlugin.on_settings_load) to encrypt/decrypt depending on which way it's going.

eshapiro42:
Thanks for the quick review! I've implemented the changes you requested.

cp2004:
> since it looks like OctoPrint doesn't test on versions newer than Python3.7.

We test against Python 3.7+, the automated ones run on 3.7, 8, 9 and 10.

OctoPi 0.18 is running 3.7, as it is Buster based, the newer images that are in testing are running 3.9. You can see a full breakdown if you like on https://data.octoprint.org/

eshapiro42:
Thanks for the clarification! This plugin might be useful to more people than I thought.

jneilliii (approved):

PR #1032, opened 2022-03-17
Update continuousprint.md

smartin015:
A quick cleanup of the plugin page - corrects the python version range and makes it a bit less wordy.


jneilliii:
As the checks show, you should never change the date after initial registration. Please put that back.

PR #1049, opened 2022-04-15
Added plugin Detector2.

mikulash:
AI detection on frontend


jneilliii:
I think the email password should not be stored in plain text if it's avoidable. The [Wyze](https://github.com/OctoPrint/plugins.octoprint.org/pull/1030#issuecomment-1070305168) plugin recently implemented this in their plugin and might be a good reference point on how to do it.

mikulash:
thanks for the quick response! I added the suggested encryption

jneilliii (approved):

jneilliii:
You need to crate a release on your GitHub repository that matches the current version 0.1.0 in setup.py. When new versions are ready update setup.py and create a new release with a matching version tag that way users that have the plugin installed will get update notices.

mikulash:
oh, did not know about this. it is done. thank you for reviewing this even on weekend

cp2004:
Is there any noticeable performance impact or benefit sending the images over the websocket vs. serving a HTTP request?

Other question, there is a get request for the snapshot image, but the response never seems to be used: https://github.com/mikulash/Octoprint-detector2/blob/bb561255255ae524f6e1267978159076fe4765bd/octoprint_detector2/__init__.py#L66
Is this needed for something I haven't spotted, or can it be removed?

jneilliii:
> Is this needed for something I haven't spotted, or can it be removed?

Good point, it doesn't appear to be necessary and would improve performance.

mikulash:
oh, did not notice that, thanks, i have removed it

PR #1050, opened 2022-04-16
add Bed Ready plugin

jneilliii:


cp2004:
Not a proper review, just a couple things I spotted when having a poke, I think it is implemented quite neatly and easily.

https://github.com/jneilliii/OctoPrint-BedReady/blob/ef3cdd7a456e1d189ec8096d921617c8faed1bd5/octoprint_bedready/__init__.py#L57 - requests always need a timeout, otherwise they *could* hang if the webcam server is broken. If it knows it is broken, it should return a 503 or something, but not always. 

https://github.com/jneilliii/OctoPrint-BedReady/blob/ef3cdd7a456e1d189ec8096d921617c8faed1bd5/octoprint_bedready/__init__.py#L117 - debug logging needed?

jneilliii:
Thanks Charlie, timeout added and logging lines cleaned up.

kantlivelong (approved):
:+1:

PR #1053, opened 2022-04-27
Added plugin psucontrol_fritz

not-na:
This adds a sub-plugin to the PSU Control plugin adding support for FRITZ!Box/FRITZ!Dect smart home devices.


_plugins/psucontrol_fritz.md
```diff
@@ -0,0 +1,36 @@
...
+author: not-na
+license: AGPLv3
+date: 2022-04-27
+tags:
+- power
+- psu
+- control
+- psucontrol
+- psucontrol subplugin
+- tplink
```
kantlivelong, line 18: Might want to adjust this :smirk:

kantlivelong (changes requested):
Just two things to adjust.

https://github.com/not-na/OctoPrint-PSUControl-Fritz/blob/4d59a10d7f507ef960e0074a6a7dfdceb19a75db/octoprint_psucontrol_fritz/__init__.py#L146

cp2004:
RE: https://github.com/not-na/OctoPrint-PSUControl-Fritz/blob/4d59a10d7f507ef960e0074a6a7dfdceb19a75db/octoprint_psucontrol_fritz/__init__.py#L57

Will this end up logging usernames and passwords? That is not really what we would want happening and is a security problem. Log files are designed to be shared in case of a problem.

kantlivelong:
> RE

Good catch. Note that I actually do (and thus this fork) log that with [DEBUG](https://github.com/not-na/OctoPrint-PSUControl-Fritz/blob/4d59a10d7f507ef960e0074a6a7dfdceb19a75db/octoprint_psucontrol_fritz/__init__.py#L51) but since this is INFO definitely worth filtering.

_plugins/psucontrol_fritz.md
not-na: Oops, sorry. I overlooked that while adjusting the description. I'll fix that right now.

not-na:
I've fixed the logging issue. It's probably an artifact of when I was debugging some issues while creating a session due to an outdated dependency.

kantlivelong (approved):

PR #1062, opened 2022-05-07
Replace TSD with Obico

kennethjiang:
@jneilliii I'm not 100% sure if I did it right, especially for the TSD plugin. I'm not sure how it'll look like in the plugin manager. My best guess is with what I have here, the users can still install TSD plugin, but they will get Obico plugin instead, which can be surprising. Please shed some light here.

Also it looks like we are on track to launch Obico Sunday or Monday. Will appreciate it if you can hold off merging this PR until I give you the final confirmation.

Thanks again for being so helpful!


jneilliii:
Yes, you are correct. This would basically list the plugin as TSD in plugin manager and then install Obico. It may be enough to just add a note at the top of your plugin's wizard implementation referencing the name change? Otherwise we can leave TSD as is and once installed they would get the notice about the change. Curious as to what @cp2004 and @kantlivelong think would be the better approach. 

An alternative to the notice/manual install would be adding a [update check overlay](https://github.com/OctoPrint/plugins.octoprint.org/blob/gh-pages/_data/update_check_overlays.yaml) for TSD to automate the upgrade over to Obico. I think in this scenario you would just have to make sure the Obico plugin release version was newer/higher than the current TSD version. 

Please insure that all links in plugin listing are not broken (ie https://obico.io/obico-cloud-vs-self-hosted.html) as this will undoubtedly lead to someone posting an issue.

In order for me to finalize the review I will need access to the Obico repository.

_plugins/thespaghettidetective.md
```diff
@@ -3,16 +3,17 @@ layout: plugin
 
 id: thespaghettidetective
 title: Access Anywhere - The Spaghetti Detective
-description: Securely monitor and access your full OctoPrint instance from anywhere, for free! Unlimited live webcam streaming. Free companion mobile app for iOS and Android. The best part? AI-powered failure detection. (TSD is the successor of OctoPrint Anywhere).
-author: TSD Team
+description: The Spaghetti Detective is now Obico! Please install the Obico plugin
+instead.
```
cp2004, line 7: Remove extra newline, description should be one line, that will fix one of the build failures.

cp2004:
Here's a preview of what it will look like (manually updated): https://cp2004-plugins-octoprint-org-git-pr1062-cp2004.vercel.app.

To test in the plugin manager, you can swap the URL used in the plugin manager's settings to be https://cp2004-plugins-octoprint-org-git-pr1062-cp2004.vercel.app/plugins.json then refresh the list from the repository. It should update with the new source, so you can see what it looks like in there. However, due to the build failures for that page it does not show up properly at all.

The build checks of course need to be fixed - there's something wrong with the formatting in the TSD file now, and also the lack of Python 3 compatibility in the 'new plugin' as the checks have noted.

_data/notices.yaml
```diff
@@ -190,3 +190,9 @@
     repository of the old maintainer. Please uninstall the old plugin from the plugin manager, then reinstall from the plugin
     repository to update it. The release notes of the version by the new maintainer can be found at the "Read more..." link below.
   link: https://github.com/egguy/OctoPrint-Prusa-Mini-ETA/releases/tag/0.1.4
+- plugin: thespaghettidetective
+  date: 2022-5-9 12:00:00Z
+  text: The Spaghetti Detective plugin is now replaced by Obico plugin. Please uninstall it (keep data) in the plugin manager.
+    Then install the "Obico for OctoPrint" plugin. Your current settings in the old plugin will be auotmatically migrated
```
cp2004, line 196:
```suggestion
    Then install the "Obico for OctoPrint" plugin. Your current settings in the old plugin will be automatically migrated
```

cp2004 (changes requested):

kennethjiang:
Thank you folks for providing the review and detailed info. And sorry for the stupid mistakes. Will try to set up a server and to test the behavior of overlay to see if it's a good way to go. I think the fallback plan is just to add a warning to TSD plugin wizard for the users who still choose to install.

I'm still working on these pages so the links will be there before tomorrow, including the new repo. I guess I created extra work for you by going hyper multi-tasking... :(

kennethjiang:
Oh one question: are new plugins required to support only python 3 now? It looks like there are still 17% of the active OctoPrint users on python 2. Not a huge number but don't want them to be stuck in TSD plugin without an option to move to Obico.

jneilliii:
New plugins are not required to be Python 3 only, but the build failures are because you didn't specify any python compatibility flag in the obico.md file.

kennethjiang:
@jneilliii @cp2004 Please ignore my pushes for now as I'm fixing problems and experimenting with what I can do with tsd.md. I'll signal you when it's ready for review again.

kennethjiang:
@jneilliii @cp2004 After testing it, I think the least confusing way is to let users install the TSD plugin if they miss the "TSD is succeeded by Obico" description. Then in TSD wizard I'll warning them about the switch.

I think everything is ready for this PR except the github links are still broken. Please ignore the links and review the rest of the PR again. Will ping you back when the links are ready to go, in a few hours. 

@cp2004 I think I have resolved both of the issues you brought up but somehow github still shows that changes are requested. Maybe I missed something?

cp2004:
> @cp2004 I think I have resolved both of the issues you brought up but somehow github still shows that changes are requested. Maybe I missed something?

I haven't re-reviewed it, so it will still show as changes requested until I re-review and approve it. The automatic checks are passing now.

Does @jneilliii have access to the Obico repo (listed as https://github.com/TheSpaghettiDetective/OctoPrint-Obico) to do the plugin review?

jneilliii:
> @cp2004 I think I have resolved both of the issues you brought up but somehow github still shows that changes are 
> Does @jneilliii have access to the Obico repo (listed as https://github.com/TheSpaghettiDetective/OctoPrint-Obico) to do the plugin review?

After a short DM in Discord it seems @kennethjiang is just going to be renaming the original TSD repo to the new name, and then forking/re-uploading TSD into the old name so there is still somewhere to go. He was concerned about losing his stars on GitHub. BTW Kenneth, you will lose plugin stats with this new name. We have been discussing internally for a way to have these merged, but as it stands now with the new plugin_identifier you will have new stats. 

Which leads me to the question if the new name should get a new date in the plugin listing. My feeling from previous experience with my own personal plugin name change that the date should remain the same as the original.

kennethjiang:
All repos links should be valid now.

> BTW Kenneth, you will lose plugin stats with this new name. We have been discussing internally for a way to have these merged, but as it stands now with the new plugin_identifier you will have new stats.
It's fine that the stats get reset. Any chance we can use the new date?  Just try to stay on top of the plugin list for a big longer to make up for the loss of stats. It's technically not a rename since the TSD plugin still exists lol.

jneilliii:
we discussed internally and the date should remain the same as original because this is not a new plugin.

kennethjiang:
> we discussed internally and the date should remain the same as original because this is not a new plugin.

No problem. Just pushed. 

Please let me know if there is anything else I need to change. I expect to go live in a couple hours the most.

jneilliii:
Quick look and setup.py version doesn't match release version?

jneilliii (approved):

kennethjiang:
Thanks @jneilliii @cp2004  !

PR #1072, opened 2022-05-24
Add karmen connector plugin

pavel-mxsf:


pavel-mxsf:
> Need to have version 0.1.0 released on GitHub repository. Otherwise, nothing else appears to be out of line.

Thanks, I've just done release on github, so it should be fixed.

jneilliii:
Thanks given this is a remote access app, you may want to include it in the list [here](https://github.com/OctoPrint/plugins.octoprint.org/blob/gh-pages/_topics/remote_access.md) as part of your PR. Since this is a remote access and potential security concerns, I'd also like to get a pair of seconds eyes on this PR from @kantlivelong or @foosel  and you should probably include a link to Karmen's security/privacy policy on the plugin page as well.

pavel-mxsf:
Hello,
I have:
- Restricted settings API access to admins only
- Port and host is loaded from `on_startup` method
- Added security and privacy info to the README, plugin settings and octoprint plugin repository readme
Thanks for review, it was really helpful

foosel (changes requested):
- I'm sorry I overlooked this the last time, but [do not inject your secrets into the template vars](https://github.com/fragaria/karmen-octoprint-plugin/blob/1adbb465aaa9a9ad46975b7573e5e48b526438a0/octoprint_karmen/__init__.py#L30-L34). You thankfully don't seem to use them in your template, but any other plugin could.
- This is still unsolved in my eyes: "Acts as a reverse proxy if I understand this correctly? Make sure to forward client IPs correctly then and/or make it perfectly clear to the user that the cloud will act like a local client here, meaning autologin might trigger." You state that the Karmen Cloud will get full control, but if you expose the web interface through your reverse proxy publicly and the headers aren't properly set, local autologin WILL login arbitrary visitors. That's a security risk. So, again, make sure that remote clients are identified as remote client and not logged in if autologin is enabled.

pavel-mxsf:
- Secret key is now hidden in the template (just first and last two characters are displayed when set).
- For GUI protection I've limited access to whitelisted addresses: by default `http://api_host:api_port/api/` and webcam snapshot url. 
- Whitelist default is `/api/` and can be extended in advanced settings (eg. we have octoprint, webcam and one more service behind haproxy on our custom installment). URL request path must start with some string from whitelist otherwise request is discarded. Localhost requests are allowed with only exception - webcam URL which must match octoprint's webcam snapshot setting. I hope this is enough to protect GUI autologin.
- List of allowed URLs is also displayed in settings.

jneilliii:
foosel is currently on vacation for a few weeks, so it will probably be a while before she's back to continue the review. Thank you for your patience.

foosel (changes requested):
All good now as far as I can see (obviously cannot vouch for the server side), apart from your version number. On GitHub you have published 0.2.1, in your `setup.py` you still state 0.1.0. The two must always match or you'll get endless update notifications. Once that is fixed, we can update the registration date here and go ahead with merging. Please re-request a review, once that's been taken care of.

martinbilek:
Hi @foosel, thanks for your review. We have changed version to 0.2.2 in `setup.py` to correspond with our latest plugin release. I also changed release date to today. If anything more is needed, please let us know. Thank you!

foosel (approved):

foosel:
Thank you!

PR #1076, opened 2022-06-06
Add plugin 'queue'

chennes:
A new plugin for managing a print queue. From the description:

Designed for use at a public library, this plugin allows staff to efficiently manage customer print 
jobs. Its simple interface and setup are easy to learn, and your organization can either implement its own
prioritization scheme or use the one provided with the plugin. It is primarily intended as a staff communication
mechanism, allowing staff to see which job they should start up next, as well as look up completed jobs
to provide information to customers.


_plugins/queue.md
```diff
@@ -0,0 +1,55 @@
...
+- url: /assets/img/plugins/queue/settings.png
+  alt: The settings screen.
+  caption: The settings screen showing the default priorities.
+
+featuredimage: /assets/img/plugins/queue/queue.png
+
+compatibility:
+  octoprint:
+  - 1.3.0
+  python: ">=3.0,<4"
```
jneilliii, line 37: Compatibility mismatch between this and your `__init__.py file [here](https://github.com/chennes/OctoPrint-Queue/blob/1b5bb570aac3b5fc79cb35654d4fe76b32a42200/octoprint_queue/__init__.py#L9). I would lean towards updating your plugin to be python 3 only as that is the way forward with future development.

jneilliii (changes requested):
see notes above as well as release a new version 2.0.0 on your plugin's repo, as that's what is defined in setup.py [here](https://github.com/chennes/OctoPrint-Queue/blob/1b5bb570aac3b5fc79cb35654d4fe76b32a42200/setup.py#L15).

jneilliii:
I'd also say remove the EventHandlerPlugin mixin implementation from your plugin class definition as it's not being used.

_plugins/queue.md
chennes: Good suggestion, thanks: https://github.com/chennes/OctoPrint-Queue/commit/b9dfac4610e9ea4fb37aa1dcf50de2b3b7b47460

chennes:
> I'd also say remove the EventHandlerPlugin mixin implementation from your plugin class definition as it's not being used.
Done now, in https://github.com/chennes/OctoPrint-Queue/commit/231e76ee165329853f6913aa3ae68f2fa1c1f042

chennes:
> see notes above as well as release a new version 2.0.0 on your plugin's repo, as that's what is defined in setup.py [here](https://github.com/chennes/OctoPrint-Queue/blob/1b5bb570aac3b5fc79cb35654d4fe76b32a42200/setup.py#L15).

Thank you for the review: I've incorporated it into my repo and tagged release 2:
https://github.com/chennes/OctoPrint-Queue/releases/tag/v2.0.0

I believe this addresses all of your feedback, please let me know if I've missed something.

jneilliii (approved):

PR #1078, opened 2022-06-09
OctoPrint-CreatbotUtil Registration

kForth:
See plugin repository here: https://github.com/kForth/OctoPrint-CreatbotUtil


_plugins/CreatbotUtil.md
```diff
@@ -0,0 +1,61 @@
...
+- temperature
+- heated
+- chamber
+- build
+- volume
+
+# featuredimage: null
+
+screenshots:
+- url: /assets/img/plugins/creatbotutil/settings.png
```
jneilliii, line 35: plugin identifier folder in img path should match id above.

jneilliii (changes requested):
your plugin's on_event callback should exit early and not process on every event. add something like 

```
def on_event(self, event, payload):
    if event not in ["PrintStopped", "PrintCancelled", "PrintDone", "PrintStarted"]:
        return
```

same goes for your gcode_sending_hook function, return early to avoid processing overhead on every command being sent to the printer.

jneilliii (approved):

PR #1089, opened 2022-07-18
Added plugin PersistTerminalCommandHistory

nivekmai:


nivekmai:
Plugin for context: https://github.com/nivekmai/OctoPrint-PersistTerminalCommandHistory

It's mostly boilerplate, just storing `terminalViewModel.cmdHistory` into the settings data, updated via waiting for `command` to go empty and then saving (I would rather subscribe to `cmdHistory`, but it's a plain array and not an observable, happy to make a commit to change it in the base repo if that's the way we wanna go).

jneilliii (changes requested):
One small change @nivekmai on the plugin repo. The tag on your latest release should match the version number defined in setup.py. That's what the software update checks will verify against. 

Very clean/simple solution for handling command history. Only potential issue I could see is having 300 additional lines saved into config.yaml. Would you consider overriding [on_settings_load](https://docs.octoprint.org/en/master/plugins/mixins.html#octoprint.plugin.SettingsPlugin.on_settings_load) and [on_settings_save](https://docs.octoprint.org/en/master/plugins/mixins.html#octoprint.plugin.SettingsPlugin.on_settings_save) on the python side to read and store those lines in the plugin's data folder rather than config.yaml? You could use `self.get_plugin_data_folder() + "/command_history.yaml"` to get a path to where to save and then add the logic for saving yaml to that path. Or use a json format and write direct to that file path. I've done similar for OctoDashCompanion plugin [here](https://github.com/jneilliii/OctoPrint-OctoDashCompanion/blob/faae265de9c7b83b5e2754edad4be649b9c40f26/octoprint_octodashcompanion/__init__.py#L43-L96) as an example.

nivekmai:
Plugin updated to v0.2.0 which implements the requested changes: https://github.com/nivekmai/OctoPrint-PersistTerminalCommandHistory/commit/cfc4b533a281d366fec53601362a5dc9a6c54726

jneilliii (approved):
Thanks, I think that will work out better in the long run, and those files also still get backed up by OctoPrint and will restore as expected.

PR #1091, opened 2022-07-21
Registering autprint plugin

chof747:
Register the first productive version of the auto print plugin


jneilliii:
Once the build and validate passes (need to put in featureimage which is what goes out with notices of your plugin) I can review. Technically the only difference between this plugin and PSU Control is that you are controlling two GPIO pins vs just the one?

PR #1092, opened 2022-07-21
Adding autoprint plugin

chof747:


jneilliii (changes requested):
Couple of points on your plugin @chof747. If you add RPi.GPIO [here](https://github.com/chof747/octoprint-autoprint/blob/357831cc1ab3b513754d4c25863e1839cad8e44c/setup.py#L36) in plugin requires as `plugin_requires = ["RPi.GPIO]` the dependency will automatically be installed and you can remove the manual install instructions from this PR lines 63-69. 

Since this is utilizing RPi.GPIO you should add 
compatibility:
```
  os:
  - linux
```
because it will only work on a Pi. 

You also need to create a release on your plugin's repository that is tagged the same version as you have in setup.py `0.1.0`, as you release new versions create a new release with the new version tagged and that way your users will get notified of updates.

chof747:
Dear @jneilliii, thanks for the review, I will implement the points. I just have one question regarding adding the RPi.GPIO dependency: I did not add this automatically as I am testing the system not on a pi but on my laptop computer and I am using a simulation library instead. Adding this would require me to modify the setup.py every time I install the plugin on my OctoPrint test environment. Do you see any way around that and still being able to make this a declared dependency? - Thanks chof747

jneilliii:
I think you can add additional logic to the setup.py to handle both situations by checking against the os it's running on and then define the plugin_requires differently. I can't think of an example doing this off the top of my head but something like this might work.

```
try:
    with io.open('/sys/firmware/devicetree/base/model', 'r') as m:
        if 'raspberry pi' in m.read().lower(): 
            plugin_requires = ["RPi.GPIO"]
except Exception:
  plugin_requires = []
```

chof747:
> I think you can add additional logic to the setup.py to handle both situations by checking against the os it's running on and then define the plugin_requires differently. I can't think of an example doing this off the top of my head but something like this might work.
> [...]

Thanks for the suggestion, this method worked I changed it now so that RPi.GPIO is only installed on the PI and it can be installed on test systems with special test instructions which are found in the README.md in the main directory. Can you please check it and conclude the review? - Thx

jneilliii (approved):

Issue #796, opened 2021-02-24
[Abandoned Plugin] OctoPrint-Filetypes


awenelo:
**Which Plugin?**

https://plugins.octoprint.org/plugins/filetypes/

**Link to ticket on plugin's issue tracker confirming abandonment**

https://github.com/TheLongRunSmoke/OctoPrint-Filetypes/issues/3

**Is there someone open to adopting it?**

_No response_

PR #1094, opened 2022-07-26
Claim OctoPrint-Filetypes

kForth:
I have forked and updated the OctoPrint-Filetypes plugin and would like the become the new maintainer, see [kforth/OctoPrint-Filetypes](https://github.com/kForth/OctoPrint-Filetypes).

This plugin has not been updated in almost 5 years and was marked as abandoned over a year ago (#796), no one has since claimed it.

If preferred, I can submit this as a new plugin instead of claiming the abandoned one.
Suggestions are always appreciated.


_plugins/filetypes.md
```diff
@@ -3,15 +3,15 @@ layout: plugin
 
 id: filetypes
 title: Filetypes
-description: Let you select file types visible in upload dialog
-author: thelongrunsmoke
+description: Lets you control which file extensions are allowed to be uploaded thru the 'Upload' and 'Upload to SD' buttons.
+author: Kestin Goforth
```
jneilliii, line 7: Please convert this to authors as a list.
```
authors:
- Kestin Goforth
- thelongrunsmoke

File: _plugins/filetypes.md [hunk 2]
```diff
@@ -20,15 +20,25 @@ tags:
...
+
+If you have installed plugins that support additional file types, add them to the list so that you can easily upload them.
+
+**Note:** The server will still reject uploads for unsupported file types. Only add extensions that are supported by your installed plugins.
+
+## Configuration
+
+Go to 'Settings' -> 'Filetypes Plugin' and configure the extensions you need.
+
+![Settings](https://raw.githubusercontent.com/kforth/OctoPrint-Filetypes/doc/settings.png)
```
jneilliii, line 44: change to internal image `/assets/img/plugins/filetypes/settings.png`

jneilliii (changes requested):
Just a couple of minor adjustments that need to be taken care of.

cp2004:
What is the use case for this plugin these days?

It mentions a lot on the repo about adding file types that other plugins support uploading to this plugin. However, other plugins can do that themselves with the [`octoprint.filemanager.extension_tree`](https://docs.octoprint.org/en/master/plugins/hooks.html#octoprint-filemanager-extension-tree) hook. By the looks of the code, this plugin would end up overriding any extensions that are added properly without configuring in the plugin. Maybe the custom file types that you've added will have a use case I don't know of, and there is already a plugin 'Upload Anything' that looks like it does similar.

kForth:
Ah yes, that does exactly what I wanted and with better functionality.

I was looking to add support for some other file types and found this plugin which fixed the first half of my problem, tho I see now that the file extension support solves both halves of my problem (client half and server half). I just figured I'd update this plugin while I was modifying its code anyway.

I suppose the use case is now fixing old or broken plugins while also breaking newer plugins if you don't enter every file type.

The functionality of the original plugin is pretty much obsolete then:
 - Server blocks unsupported types added by plugin
 - Plugin blocks support types added by other plugins

Closing this PR and leaving the plugin abandoned since it's probably not worth updating.

PR #1105, opened 2022-09-12
add HeaderAnnouncements plugin

wwsean08:
Allow administrators to setup announcements for their users in a shared environment, allowing for important information to be shared in the OctoPrint UI, like upcoming maintenance, and procedures.


_plugins/HeaderAnnouncements.md
```diff
@@ -0,0 +1,43 @@
...
+layout: plugin
+
+id: HeaderAnnouncements
+title: OctoPrint-HeaderAnnouncements
+description: Add an announcement to the header of your octoprint install for use in a shared environment
+authors:
+- Sean Smith
+license: MIT
+
+date: 2021-09-12
```
jneilliii, line 11: 2022-09-12

jneilliii (changes requested):
small fix for the plugin listing., should be 2022.

as for the plugin itself, a cursory look seems to indicate you don't need this line [here](https://github.com/wwsean08/OctoPrint-HeaderAnnouncements/blob/3cff6ef4b5cad19a41b5217bc44fa4a4b229b917/octoprint_HeaderAnnouncements/__init__.py#L30) as it's not an actual setting being used anywhere.

Also, I'm pretty sure since you're not using custom bindings for the settings jinja2 template you don't need to include the dependency in the binding definition [here](https://github.com/wwsean08/OctoPrint-HeaderAnnouncements/blob/3cff6ef4b5cad19a41b5217bc44fa4a4b229b917/octoprint_HeaderAnnouncements/static/js/HeaderAnnouncements.js#L27), so should be `dependencies: []` but I would definitely test that out yourself to know for sure.

jneilliii:
> For the `custom_bindings` being set to false, I found that without that, my settings were not getting persisted, it's possible that I was doing something wrong.

That is true for the get_template_configs function, but not necessary within the get_settings_default. 

Was discussing the way this plugin utilizes polling to retrieve data in our Discord server and it would probably be better to utilize [send_plugin_message](https://docs.octoprint.org/en/master/modules/plugin.html?highlight=send_plugin_message#octoprint.plugin.core.PluginManager.send_plugin_message) in an [on_settings_save](https://docs.octoprint.org/en/master/plugins/mixins.html#octoprint.plugin.SettingsPlugin.on_settings_save) override. Then in the js side you add a [onDataUpdaterPluginMessage](https://docs.octoprint.org/en/master/plugins/viewmodels.html#callbacks) callback to manipulate the UI. The only reason I mention this is because a future change may break the fetch polling you have now. 

Something like this on the python side...
```
def on_settings_save(self, data):
  octoprint.plugin.SettingsPlugin.on_settings_save(self, data)
  self._plugin_manager.send_plugin_message(self._identifier, {SettingsKeys.SETTING_ANNOUNCEMENT: self._settings.get([SettingsKeys.SETTING_ANNOUNCEMENT], merge=True})
```

With this on the js side...
```
self.onDataUpdaterPluginMessage = function(plugin, data) {
  if (plugin != "HeaderAnnouncements") {
    return;
  }

  // do stuff here
}
```

wwsean08:
Oh gotcha, I missed that instance of `custom_bindings`.  Got it I will look into doing that as opposed to polling.

jneilliii:
Happy to go ahead and approve and merge, just wanted you to be aware.

jneilliii (approved):

Issue #918, opened 2021-07-29
[Abandoned Plugin] LEDStripControl


awenelo:
### Which Plugin?

https://plugins.octoprint.org/plugins/ledstripcontrol/

### Link to ticket on plugin's issue tracker confirming abandonment

Github repository is archived, a ticket can't be created

Jordan-Morales:
What all is involved in caring for a plugin. 👀

jneilliii:
Mainly just dealing with support related issues, and if needed bug fixes/feature requests at the discretion of the adopter. If you are interested in taking over this plugin there is a small process that has to be completed to transfer the pointers in the plugin repo to your fork of the plugin.

LunkwillAndFook:
Give me a few days to look at the code. I may take this since I use the plugin.

dzikk:
Hi
I want to use this plugin only with white led, How can I modify the source code to make possible to toogle state of led with only one command? currently I have to use M150 W and M150

OllisGit:
Hi @dzikk,
this is not the right place to ask this question, because this repository is only for managing the plugins-page: https://plugins.octoprint.org. 
The plugins itself are maintained in there own repository....but this LED-Plugin is archived..no change to raise the question there.

You should use the community forum: https://community.octoprint.org/
There is a much bigger audience for answering your questions.

Regarding your question: I looked into the sourcecode and from my perspective you should implemente something like this (pseudo-code):

New command type: M150 TW:  T = toggle
```
if cmd == 150
  if cmd.contains(T)
      onOrOff = readGPI from gcode seleced LED (e.g. W or R or B)
      if onOrOff == true
        dutycycles = {'w':0.0}
        self._leds[l].ChangeDutyCycle(dutycycles[l])
  else
    ...do the old stuff
```

gbonesso:
Hi! I'm interested in adopting this plugin. This plugin and it's not much complicated and it's a good place to start contributing to the community. What should I do to start? I've already forked it here (https://github.com/gbonesso/OctoPrint-LEDStripControl)...

jneilliii:
Modify all the files in your fork to reference it's GitHub address instead of your own, including setup.py and update hook in init.py and then submit a PR here modifying the [ledstripcontrol.md](https://github.com/OctoPrint/plugins.octoprint.org/blob/gh-pages/_plugins/ledstripcontrol.md) template file adjusting the URLs and switching author to a list called authors and adding yourself to it, like this.

```
[...]

PR #1107, opened 2022-09-27
Adoption - LEDStripControl (#918)

gbonesso:
I've made the changes as suggested.


gbonesso:
I've send a new PR with a screenshot update...

PR #1124, opened 2022-11-28
Added plugin FusiBox

hm1996:


cp2004 (changes requested):
Hi @hm1996, apologies for the delay in getting this plugin reviewed for you.

I've taken a look, and have a some comments for some things that **need** adjusting:

* You use a 'basepath' that is the path of the plugin package for storing all sorts of things, like a database and images, audio recordings etc. This shouldn't be the plugin code/package, but instead stored in OctoPrint's data folder. It's quite likely that data in the plugin package is lost on every update to the plugin, but also there is no guarantee that you can just write data to that folder. I would recommend using [`self.get_plugin_datafolder()`](https://docs.octoprint.org/en/master/modules/plugin.html#octoprint.plugin.types.OctoPrintPlugin.get_plugin_data_folder) in your main plugin package and then pass that through as `basepath`. Note that this will mean the stuff you store here gets backed up by OctoPrint's backup & restore functions, but [you can exclude it](https://docs.octoprint.org/en/master/bundledplugins/backup.html#octoprint-plugin-backup-additional-excludes) if you would like.

* In your file `i2smic.py`, you attempt to run several system commands, including installing some packages on the OS. I can't see where this is triggered from (other than a commented-out reference [here](https://github.com/Fusibox/OctoPrint-Fusibox/blob/e196575f8479a266d163da91e5d1ec2f69b3b55b/octoprint_fusibox/__init__.py#L154)) but it won't work without `sudo` (& therefore requiring the user's password). If this is called anywhere, then it **must** only be done after user interaction fully explaining what the plugin will do. If this isn't used any more, then I assume you'll need some kind of guide/instructions on setting this up for users?

And also some **recommendations**:

* You might want to add to your plugin description that this is only supported on a Raspberry Pi - since it uses of the GPIO pins it won't work on other devices. It [imports `RPi.GPIO`](https://github.com/Fusibox/OctoPrint-Fusibox/blob/e196575f8479a266d163da91e5d1ec2f69b3b55b/octoprint_fusibox/modules/outputs.py#L7), which will only work properly on a Pi-system.

* You might also want to add some more descriptions about what the hardware this plugin links to. Many users will get a notification about this new plugin when it is published, and they may be questioning what 'FusiBox' is, to work out if this is something they are interested in. I can't find your hardware anywhere on the internet to see what this plugin is actually interacting with.



Let me know when you have addressed the above points by requesting a re-review here.

kantlivelong:
Stale - Closing

PR #1133, opened 2022-12-07
Added plugin SmartABL

scmanjarrez:


scmanjarrez:
Detected two bugs that need to be fixed:

1. If a mesh doesn't exist and SmartABL tries to load a mesh, it crashes the connection.
2. Commands sent through terminal must be skipped, SmartABL shouldn't process them, otherwise above bug could happen.

jneilliii (approved):

PR #1135, opened 2022-12-08
Update plugin links and description.

arekm:
Plugin has new home (old url redirects to new home). Support for Marlin2 added, so update description.


jneilliii (approved):
Appears to be a repo transfer from kanocz to arekm, no overlays needed in this case.

PR #1144, opened 2023-01-04
Added plugin esp32camUI

CyberSensei1:


jneilliii (commented):
Nice use of the hello world plugin example. You do know that webcam iframe plugin exists already?

jneilliii (approved):

PR #1150, opened 2023-01-23
Add OctoPrint-PowerFailure

paukstelis:
Submitted on behalf of both authors.

### DIFF
--- a/_plugins/powerfailure.md
+++ b/_plugins/powerfailure.md
@@ -1,0 +1,41 @@
+---
+layout: plugin
+
+id: powerfailure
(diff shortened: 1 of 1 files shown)


jneilliii (commented):
I wonder if maybe all the logger.info messages shouldn't be debug instead? At least where it may not make sense (ie if there isn't a failure).

paukstelis:
Changes in [241b694](https://github.com/pablogventura/OctoPrint-PowerFailure/commit/241b69433bc95f5fc5b7a383cea0f9928e26bf3f)

jneilliii (approved):

PR #1163, opened 2023-02-26
Added plugin InfluxDB2

MoltenBytes:
Here's the link to the [new plugin](https://gitlab.com/moltenbytes32/octoprint-influxdb2)

Not sure what else I can say that hasn't already been said in these documents or back on the homepage :sweat_smile:


jneilliii (changes requested):
I only see one change that I would request prior to merging this into the plugin repo. Your plugin is missing a software update hook, which would allow for your users to automatically get update notifications on new releases. You can find the docs for this [here](https://docs.octoprint.org/en/master/bundledplugins/softwareupdate.html#hooks). I can't say I remember another GitLab example of this but will try to find one for you.

jneilliii:
The only other GitLab example of a plugin update I could find actually utilizes PyPi release updates. I'm sure there's a way to do it similar to the GitHub version without PyPi releases, just don't have an example for you. 

https://gitlab.com/wolframmfg/octoprint-swapxy/-/blob/main/octoprint_swapxy/__init__.py#L17

MoltenBytes:
I think I got it -- gitlab doesn't have native support, but I was able to make it work with `git ls-remote` on a command line. A few things to note:

- It might be good to support `git ls-remote` as a ubiquitous option for any git server
- [The docs](https://docs.octoprint.org/en/master/bundledplugins/softwareupdate.html#hooks) you linked to say that you can use the `command_line` check, but digging through the source it's actually `commandline`

If you want to test it, you can install the package from v0.1.2 here and it should prompt and successfully update to v0.1.3
`https://gitlab.com/moltenbytes32/octoprint-influxdb2/-/archive/v0.1.2/octoprint-influxdb2-v0.1.2.zip`

jneilliii:
Test worked as expected, only downside to the commandline update method though means os compatibility needs to be flagged now to linux only in your plugin template.

jneilliii:
I remembered where I've seen GitLab update before. It was with my palette and it's related plugins.

https://gitlab.com/mosaic-mfg/palette-2-plugin/-/blob/master/octoprint_palette2/__init__.py

MoltenBytes:
Ah, that makes sense -- thanks for the pointer. It'll probably be towards the end of the week, but I can definitely make that work.

MoltenBytes:
Alright, this should be good to go now! I said it before, but thanks again for that link reference. Not sure I'd have figured out gitlab's archive URL without it. Plugin should support software updates from Gitlab via python now :+1: 

Not sure which template you were wanting me to update the release date for.

jneilliii (approved):

PR #1172, opened 2023-03-17
Added OctoPrint-TurnOffHeatOnPauseTimer plugin

Argon2000:


jneilliii (changes requested):
The only thing I'd like to warn against and ask you to change is the self._logger calls in the plugin. It would be better to change `info` to `debug` to avoid too much writing to disk. You can always just change the logging level to debug in OctoPrint's settings when you need the extra detail. Although I don't see your plugin probably doing too much of that, it would still be better that way.

Argon2000:
Should be changed now!

jneilliii (changes requested):
One last thing I didn't catch before. Please create a release on your repo for version 0.0.1. In the future when you make updates you'll update plugin_version in setup.py and init.py and then new tag/release matching the updated version number. This is where the software update hook will catch the change and then prompt users to upgrade.

Argon2000:
Fixed!

jneilliii (approved):

PR #1182, opened 2023-04-22
Added plugin webcamextras

poulou0:
# 🐙 📷 ✌️

**Edit:**
Failed a check about python compatibility.
Added:
```
compatibility:
  python: ">=2.7,<4"
```

Initially I just deleted the entire compatibility block as instructed [here](https://plugins.octoprint.org/help/registering/) 😇
> You only need the following if your plugin requires specific OctoPrint versions or
> specific operating systems to function - you can safely remove the whole
> "compatibility" block if this is not the case.

(Outdated documentation?)


jneilliii:
Have you tested your plugin against 1.9.0rc5 yet? There are several changes to the webcam implementation coming out very soon and fullscreen button is already part of that.

poulou0:
Hey, thanks for checking so quickly! :))

No, actually I wrote this such a long time ago that I'm surprised that is still working.
But I'm using it in literally in every print so I thought its time to put it in the repo, maybe someone will benefit :)

Good point, I didn't check with the latest RC but I just noticed in the code that I hide my buttons if you have the "Float/Fullscreen webcam" enabled in the UI Customizer to not confilct. Let me know if that was the case.

<details>
  <summary>"Float/Fullscreen webcam" option</summary>
  
  uicust
</details>

https://github.com/poulou0/OctoPrint-WebcamExtras/blob/main/octoprint_webcamextras/static/js/jquery-webcamextras.js#L20:L21

I added it also in the long description of the plugin 👍

jneilliii:
I tried with disabling UI Customizer and the button wasn't there either, so think it's related to the 1.9.0 webcam changes.

poulou0:
Just published a new version.
https://github.com/poulou0/OctoPrint-WebcamExtras/releases/tag/0.0.6

TBH it needs a re-write to be more readable/efficient..
..and im happy to do so if its relevant and catches up with others!

jneilliii (approved):

PR #1185, opened 2023-04-26
Add plugin Connect And Print

Maxinger15:
This plugin connects octoprint automaticlly to your printer if you upload a file.
This is useful if you turn of your printer put octoprint is powered on constantly.

### DIFF
--- a/_plugins/connectandprint.md
+++ b/_plugins/connectandprint.md
@@ -19,3 +19,8 @@ tags:
 - connect
 - upload
 - file
+
+compatibility:
+  python: ">=2.7,<4"
+
+---
--- a/_plugins/connectandprint.md
+++ b/_plugins/connectandprint.md
@@ -13,7 +13,7 @@ date: 2023-04-26
 
 homepage: https://github.com/Maxinger15/connectandprint
 source: https://github.com/Maxinger15/connectandprint
-archive: https://github.com/Maxinger15/connectandprint/archive/master.zip
+archive: https://raw.githubusercontent.com/Maxinger15/connectandprint/master/connect_and_print.py
 
 tags:
 - connect
--- a/_plugins/connectandprint.md
+++ b/_plugins/connectandprint.md
@@ -24,3 +24,18 @@ compatibility:
   python: ">=2.7,<4"
 
 ---
+
+With the Connect and Print plugin, you can save time and effort, as there's no need to manually connect to the printer and start the print job after uploading a file. 
+Just upload your G-code file and let the plugin handle the rest.
+
+#### Features
+
+- Automatically connects to the printer if not already connected upon file upload.
+- Starts printing the uploaded file immediately after a successful connection.
+- Implements a 2-minute timeout for printer connection attempts.
+- Works with OctoPrint's REST API to streamline the printing process.
+
+#### Note: 
+Please ensure that your printer is properly configured and connected to OctoPrint before using this plugin. 
+In case of a failed connection attempt, the plugin will log an error message after the 2-minute timeout.
+
--- a/_plugins/connectandprint.md
+++ b/_plugins/connectandprint.md
@@ -1,0 +1,21 @@
+---
+layout: plugin
+
+id: connectandprint
+title: Connect And Print
+description: Automatically connect to your printer and start the print job on file upload
+authors:
+- Max Grallinger
+
+license: AGPLv3
+
+date: 2023-04-26
+
+homepage: https://github.com/Maxinger15/connectandprint
+source: https://github.com/Maxinger15/connectandprint
+archive: https://github.com/Maxinger15/connectandprint/archive/master.zip
+
+tags:
+- connect
+- upload
+- file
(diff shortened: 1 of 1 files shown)


jneilliii (changes requested):
1. Should probably add some content below the template block that explains a little more in depth what the plugin does, including 2 minute timeout, etc. 

2. Your plugin is basically a single file plugin, so you don't need `octoprint.plugin.AssetPlugin`, `octoprint.plugin.SimpleApiPlugin`, or `octoprint.plugin.TemplatePlugin` [here](https://github.com/Maxinger15/connectandprint/blob/c2f70364266152a4a449dd589569503902f1cb8b/connect_and_print.py#L8-L11) but would want to add `octoprint.plugin.RestartNeedingPlugin`.

3. Your plugin structure in the repo is not going to work as-is to install. The simplest option would be to register as a single file plugin by using `https://raw.githubusercontent.com/Maxinger15/connectandprint/master/connect_and_print.py` as the archive URL after making the changes mentioned in 2 above. Otherwise, you would need to rename the connect_and_print.py file to `__init__.py` and move it into a subfolder named connectandprint. Then your setup.py file needs to be updated, it's missing a lot of stuff that should be in there to properly install. I'd recommend copying the hello world example one from [here](https://github.com/OctoPrint/Plugin-Examples/blob/master/helloworld/setup.py) and replace all the pieces appropriately if you decide to go that route. 

4. I recommend you test the install by removing it from your OctoPrint instance and copying/pasting the `archive` URL into Plugin Manager > Get More > ...from URL and click install to validate yourself. 

5. We highly encourage you include a [software update hook](https://docs.octoprint.org/en/master/bundledplugins/softwareupdate.html#hooks) in your plugin so users are able to receive new version as you release them on GitHub. Related to that, you'll also need to create a new release with tag that matches your initial 1.0.0 version.

jneilliii:
On a side note, on a linux machine the portlister plugin can automatically connect the printer when it is detected by the OS.

Maxinger15:
Thank you very much for the valueable feedback @jneilliii. I will implement the changes soon.

_plugins/connectandprint.md
```diff
@@ -0,0 +1,41 @@
+---
+layout: plugin
+
+id: connectandprint
```
jneilliii, line 4: plugin identifiers for single file plugins are based on the filename, so you should update this id to match `connect_and_print` or rename the file and update the `archive` URL to match. whichever approach you go with you then need to change `auto_connect_and_print` [here](https://github.com/Maxinger15/connectandprint/blob/4442b9837e6623e429565058753790f06d6207a2/connect_and_print.py#L39) to match as well.

you may also want to add a `__plugin_description__`, `__plugin_author__`, `__plugin_url__`, and `__plugin_license__` [control properties](https://docs.octoprint.org/en/master/plugins/controlproperties.html) to your py file to fill in the gaps since you're not using the setup.py file, otherwise these items remain blank I think.

image

jneilliii (changes requested):

_plugins/connectandprint.md
jneilliii: oh, and you also need to create the tag and release for 1.0.0 on your repo.

Maxinger15:
I applied the changes accordingly to your suggestions.

jneilliii (approved):

AstroMelly:
Hi @Maxinger15 - just to let you know that it looks like this plugin fails to update for me.  Not sure if you can help with this or maybe it's me!  I'm pretty new to OctoPrint and 3D Printing but I am also a programmer so thought I would send this through to you.

Here's the error from the octoprint log file ~/.octoprint/logs//plugin_softwareupdate_console.log:

2023-06-05 15:50:26,258   /home/pi/oprint/bin/python -m pip --disable-pip-version-check install https://raw.githubusercontent.com/Maxinger15/connectandprint/1.2.0/connectandprint.py --no-cache-dir
2023-06-05 15:50:28,835 > Looking in indexes: https://pypi.org/simple, https://www.piwheels.org/simple
2023-06-05 15:50:28,837 > Collecting https://raw.githubusercontent.com/Maxinger15/connectandprint/1.2.0/connectandprint.py
2023-06-05 15:50:28,838 > Downloading https://raw.githubusercontent.com/Maxinger15/connectandprint/1.2.0/connectandprint.py (3.2 kB)
2023-06-05 15:50:28,842 ! ERROR: Cannot unpack file /tmp/pip-unpack-0l6n6qdc/connectandprint.py (downloaded from /tmp/pip-req-build-6i9jbhbf, content-type: text/plain; charset=utf-8); cannot detect archive format
2023-06-05 15:50:28,843 ! ERROR: Cannot determine archive format of /tmp/pip-req-build-6i9jbhbf

Maxinger15:
hi @AstroMelly ,
yeah this is a bug in that version.
Please remove the plugin and install it again.
More info here: https://github.com/Maxinger15/connectandprint/issues/6

AstroMelly:
Got it.  Thank you.  Also tracking the reported OctoPrint bug for eTag error post plugin install.  

Thanks again!

PR #1201, opened 2023-06-09
Added plugin cmdexec

alexis-coulombe:


jneilliii:
Not quite sure why the build failed here, but I think you may want to include some content at the end of the template that will be displayed on the plugin listing page. Check other plugins as example. We do encourage including screenshots in the plugin listing as well. 

Reviewing your plugin I think because of [this](https://github.com/Chargnn/Octoprint-CmdExec/blob/588868827b9e495421e649e6bf00c3b86bd319a2/setup.py#L6) line you need to rename the folder `CmdExec` to `octoprint_cmdexec` on your repo though for proper install.

alexis-coulombe:
Alright! Will come back with the changes

kantlivelong (changes requested):
Hi @Chargnn 

~~As it stands this plugin will execute shell commands from any authenticated user as the OctoPrint service account. This will give non-admin (including anonymous if enabled) users the ability to gain admin access to the app or run malicious code.~~

~~At the bare minimum you'll want to require permissions. Ideally commands should be pre-dedined to reduce the chance of such attacks.~~


Edit:

This is why I shouldn't review code in the morning.... on my phone... 🤦‍♂️

alexis-coulombe:
Hi @kantlivelong

fixed the pipeline and added permission (admin) to execute a command

jneilliii (approved):

jneilliii:
you may want to also add a data bind to your navbar for visibility based on permissions so that the button doesn't show up for non-admins. to do this use something like this after adding loginStateViewModel to your dependencies in js side and assigning it as a parameter (similar to settings). 

```py
def get_template_configs(self):
    return [{'type': "navbar", data_bind="visible: loginStateViewModel.isAdmin()"}]
```

https://docs.octoprint.org/en/master/plugins/mixins.html#octoprint.plugin.TemplatePlugin.get_template_configs

alexis-coulombe:
Thanks @jneilliii , will do the modification

PR #1216, opened 2023-07-19
New Plugin "Wattometer"

RealTryZ:


jneilliii:
Is your plugin repo private? I'm getting a 404 when attempting to review your plugin.

RealTryZ:
It was sorry for the mistake

jneilliii ***@***.***> schrieb am Mi., 19. Juli 2023, 20:14:

_plugins/Wattometer.md
```diff
@@ -0,0 +1,61 @@
...
+  - macos
+  - freebsd
+
+  python: ">=3,<4"
+
+---
+
+# Plugin Setup
+Once the plugin is installed, you will have to enter your FRITZ username and password in the settings tab.
+If you don't habe an account yet, the following [guide](https://en.avm.de/service/knowledge-base/dok/FRITZ-Box-7590/966_Creating-a-MyFRITZ-account-and-setting-it-up-in-the-FRITZ-Box/) will help you to create one.
```
jneilliii: small typo, `have`

jneilliii (changes requested):
besides the one small typo, your plugin needs [this](https://github.com/RealTryZ/Wattometer/blob/804699328b87ed3f9f18cb83940454480e7d38c4/setup.py#L36) line updated to include the fritzconnection module dependency so it automatically gets installed.

```
plugin_requires = ["fritzconnection"]
```

It's frowned upon slightly to use CDN for your js files. It's better to bundle them within your plugin in `static/js` folder and then include it in your get_assets callback. You can see an example of how this is achieved with Bed Level Visualizer [here](https://github.com/RealTryZ/Wattometer/blob/804699328b87ed3f9f18cb83940454480e7d38c4/octoprint_Wattometer/templates/wattometer_tab.jinja2#L2). 

once that gets updated create a `0.1.0` release on GitHub.

jneilliii:
Sorry I missed it before @RealTryZ, but it would be best if you also had a software update hook in your plugin, so that if you end up having to release new versions your users will get automatically notified of the new update as you make new releases. You can find the docs for that [here](https://docs.octoprint.org/en/master/bundledplugins/softwareupdate.html#octoprint-plugin-softwareupdate-check-config), with a simple example. Make sure to use `Wattometer` identifier in place of `updateplugindemo` in that example.

RealTryZ:
> Sorry I missed it before @RealTryZ, but it would be best if you also had a software update hook in your plugin, so that if you end up having to release new versions your users will get automatically notified of the new update as you make new releases. You can find the docs for that [here](https://docs.octoprint.org/en/master/bundledplugins/softwareupdate.html#octoprint-plugin-softwareupdate-check-config), with a simple example. Make sure to use `Wattometer` identifier in place of `updateplugindemo` in that example.

Thanks for the tip, I fixed it now and included that hook.

jneilliii:
You missed the bit about changing the identifier in the return [here](https://github.com/RealTryZ/Wattometer/blob/d2eb3d50ac9c1241f8f06fd97d5cd7c94fe3e839/octoprint_Wattometer/__init__.py#L65). Change `updateplugindemo` to `Wattometer` to match your plugin identifier.

RealTryZ:
> You missed the bit about changing the identifier in the return [here](https://github.com/RealTryZ/Wattometer/blob/d2eb3d50ac9c1241f8f06fd97d5cd7c94fe3e839/octoprint_Wattometer/__init__.py#L65). Change `updateplugindemo` to `Wattometer` to match your plugin identifier.

I'm really sorry, fixed it now. Thanks for your patience

jneilliii (approved):

PR #1228, opened 2023-09-16
Obico page refresh

kennethjiang:
Obico page refreshed as we have added a bunch of things since the last update. :)

Also the transition from The Spagehtti Detective has largely finished. So I'm removing it from the page.


jneilliii (approved):

kennethjiang:
Thank you @jneilliii for the prompt merge! :)

PR #1231, opened 2023-09-25
Fixed image links

printpal-io:
Broken image links previously.


jneilliii:
please start from new pull...make sure all images are in a `bedcheckai` folder (noticed this PR has inner caps on folder name which won't work).

PR #1248, opened 2023-11-27
adopt PrintJobHistory plugin

dojohnso:
Updated references to point to my fork of the repo for v1.17.1 (previously 1.17.0). The release includes a couple bug fixes people were waiting for, otherwise it's the same. 

https://github.com/dojohnso/OctoPrint-PrintJobHistory/releases/tag/1.17.1

Let me know if I missed anything or need to do anything else. Thank you!

### DIFF
--- a/_data/notices.yaml
+++ b/_data/notices.yaml
@@ -240,7 +240,6 @@
   pluginversions: ["<1.17.1"]
   versions: []
   date: 2023-11-27 12:00:00Z
-  text: Version 1.17.1 of this plugin is available from a new maintainer, but your version still looks for updates at the
-    repository of the old maintainer. Please re-install the plugin from the repository to update and keep your
-    settings. The release notes of the version by the new maintainer can be found at the "Read more..." link below.
+  text: Version 1.17.1 of this plugin is available from a new maintainer. The release notes of the version by the
+    new maintainer can be found at the "Read more..." link below.
   link: https://github.com/dojohnso/OctoPrint-PrintJobHistory/releases/tag/1.17.1
--- a/_data/notices.yaml
+++ b/_data/notices.yaml
@@ -236,3 +236,11 @@
     repository of the old maintainer. Please re-install the plugin from the repository to update and keeping your
     settings. The release notes of the version by the new maintainer can be found at the "Read more..." link below.
   link: https://github.com/thomst08/OctoLight/releases/tag/v0.1.5
+- plugin: PrintJobHistory
+  pluginversions: ["<1.17.1"]
+  versions: []
+  date: 2023-11-27 12:00:00Z
+  text: Version 1.17.1 of this plugin is available from a new maintainer, but your version still looks for updates at the
+    repository of the old maintainer. Please re-install the plugin from the repository to update and keep your
+    settings. The release notes of the version by the new maintainer can be found at the "Read more..." link below.
+  link: https://github.com/dojohnso/OctoPrint-PrintJobHistory/releases/tag/1.17.1
--- a/_data/update_check_overlays.yaml
+++ b/_data/update_check_overlays.yaml
@@ -56,4 +56,4 @@ octolight:
   user: thomst08
 PrintJobHistory:
   user: dojohnso
-  pip: https://github.com/dojohnso/OctoPrint-PrintJobHistory/archive/{target_version}.zip
+  pip: https://github.com/dojohnso/OctoPrint-PrintJobHistory/releases/download/{target_version}/main.zip
--- a/_data/update_check_overlays.yaml
+++ b/_data/update_check_overlays.yaml
@@ -54,3 +54,5 @@ webhooks:
   pip: https://github.com/derekantrican/OctoPrint-Webhooks/archive/{target_version}.zip
 octolight:
   user: thomst08
+PrintJobHistory:
+  user: dojohnso
--- a/_data/update_check_overlays.yaml
+++ b/_data/update_check_overlays.yaml
@@ -56,3 +56,4 @@ octolight:
   user: thomst08
 PrintJobHistory:
   user: dojohnso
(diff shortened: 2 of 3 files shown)


_plugins/PrintJobHistory.md
```diff
@@ -3,15 +3,17 @@ layout: plugin
...
-description: The OctoPrint-Plugin stores all print-job informations of a print in a database.
-author: Olli
+description: The OctoPrint-Plugin stores all print-job information of a print in a database.
+authors:
+    - dojohnso
+    - Olli
 license: AGPLv3
 
-date: 2020-05-28
+date: 2023-11-27
```
jneilliii: you shouldn't change this date, please change it back to the original.

jneilliii (changes requested):
Just a minor change necessary, date should remain the same.

_data/update_check_overlays.yaml
```diff
@@ -53,4 +53,6 @@ webhooks:
   user: derekantrican
   pip: https://github.com/derekantrican/OctoPrint-Webhooks/archive/{target_version}.zip
 octolight:
-  user: thomst08
\ No newline at end of file
+  user: thomst08
+PrintJobHistory:
+  user: dojohnso
```
cp2004, line 58: You need to add the pip URL as well, otherwise existing users will be prompted to upgrade to a new version, but it tries to download from the old location still.

_data/notices.yaml
```diff
@@ -235,4 +235,12 @@
...
-  link: https://github.com/thomst08/OctoLight/releases/tag/v0.1.5
\ No newline at end of file
+  link: https://github.com/thomst08/OctoLight/releases/tag/v0.1.5
+- plugin: PrintJobHistory
+  pluginversions: ["<1.17.1"]
+  versions: []
+  date: 2023-11-27 12:00:00Z
+  text: Version 1.17.1 of this plugin is available from a new maintainer, but your version still looks for updates at the
+    repository of the old maintainer. Please re-install the plugin from the repository to update and keep your
+    settings. The release notes of the version by the new maintainer can be found at the "Read more..." link below.
```
cp2004: This is less a comment that needs addressing here, but the wording used in these notices is outdated since the update check overlays (when configured properly) allow for updates to be redirected to the new version without needing to reinstall. It's not wrong, so no harm in keeping it in.

cp2004 (changes requested):

_data/notices.yaml
dojohnso: @cp2004 I'd probably rather let people keep their current data rather than uninstall and reinstall to avoid loss of history. Should I just remove this text? Or is there better text to use today?

jneilliii: I'd probably just make it something like this:

```
Version 1.17.1 of this plugin is available from a new maintainer. The release notes of the version by the new maintainer can be found at the "Read more..." link below.
```

jneilliii:
Looks like GitHub is having issues right now, so once it catches up and the additional commit(s) show up with these changes we can finalize the adoption.

_data/update_check_overlays.yaml
```diff
@@ -53,4 +53,7 @@ webhooks:
...
+  pip: https://github.com/dojohnso/OctoPrint-PrintJobHistory/archive/{target_version}.zip
```
jneilliii: This is a little different with OllisGit plugins, similar to what cp2004 does, and the url is to an attached zip asset called main.zip instead of the source zip file, so would need to be like this I think.
```
  pip: https://github.com/dojohnso/OctoPrint-PrintJobHistory/releases/download/{target_version}/main.zip
```

jneilliii (changes requested):

_data/update_check_overlays.yaml
dojohnso: ok yeah that works. I just copied what I saw from before, figured would change as needed. Thanks for reviewing all this so quickly!

cp2004: Yeah since you've got the actions working to make the releases, that makes sense. I do it this way to have the version number automatically bumped based on the tag, not sure what OllisGit did it for but definitely worth fully understanding. People will very quickly come running if you get it wrong!

cp2004 (approved):
My comments resolved, don't want to hold up the process so back over to @jneilliii :)

_data/update_check_overlays.yaml
jneilliii: Assuming they are working on @dojohnso's fork. I do see build failures on the actions though. Going to go ahead and approve and merge the adoption as that is not specific to the plugin repository listing and as 

https://github.com/dojohnso/OctoPrint-PrintJobHistory/actions/runs/7008953602/job/19066379869#step:11:26

IIRC correctly the reason OllisGit was doing releases that way was to get download stats of the asset download for the readme badges, which doesn't work with source zip downloads.

jneilliii (approved):

dojohnso:
Thanks again everyone! FYI the abandoned notice still pops up in addition to the update. I'm sure people will figure it out, just wanted to mention it in case it got overlooked. I'm not sure of your process of course so just putting it out there.  🙂

jneilliii:
Yeah, the notices typically go out quickly, while the software update caches the list of available versions for a day I think. You can always use software update settings and force check for updates to bypass the cache.

image

dojohnso:
Just curious when this abandoned notice and status goes away? 
image
image

jneilliii:
Ah, you didn't remove the abandoned line from the PrintJobHistory.md file. That was my mistake for missing it during review. Will be updated once this PR passes checks and gets merged.

https://github.com/OctoPrint/plugins.octoprint.org/pull/1249

dojohnso:
Ok sounds good, thank you! I saw that line in there but wasn't sure who managed it so I left it alone. Makes sense and again I appreciate everyone's help getting this moved. I've already had great feedback and interactions with users who are excited to see development pick up on it again. Cheers!

PR #1251, opened 2023-12-01
Added plugin ActivatePrusaHostTimer

sarusani:
This plugin "pings" the printer every x seconds with a "M79 S"OP"". (The interval can be configured in the settings page, default is 20 seconds)

This ping is needed to activate the print host capabilities of Prusa printers and is expected at least every 30 seconds.
See https://github.com/prusa3d/Prusa-Firmware/pull/4520 for more information.

### DIFF
--- a/_plugins/ActivatePrusaHostTimer.md
+++ b/_plugins/ActivatePrusaHostTimer.md
@@ -1,0 +1,51 @@
+---
+layout: plugin
+
+id: ActivatePrusaHostTimer
+title: Activate Prusa HostTimer
+description: Plugin for Octoprint to activate Prusa host features
+author: sarusani
+license: AGPLv3
+
+date: 2023-12-01
+
+homepage: https://github.com/sarusani/OctoPrint-ActivatePrusaHostTimer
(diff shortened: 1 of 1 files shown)


jneilliii:
Why was the timer check implemented in the firmware in the first place? I just reviewed your plugin and don't really see anything wrong technically with your plugin. 

I think I should point out that your implementation of the solution to send commands on a set interval however will only work if the web interface is loaded and logged into on a client computer. You might want to look into moving that repeat logic to the python side so it doesn't matter if the UI is loaded or not. 

Couple of pointers: 
[octoprint.util.RepeatedTimer](https://docs.octoprint.org/en/master/modules/util.html#octoprint.util.RepeatedTimer)
`self._settings.get(["interval"])`
[self._printer.commands](https://docs.octoprint.org/en/master/modules/printer.html#octoprint.printer.PrinterInterface.commands)

sarusani:
Thanks for pointing that out! I totally missed that.
I moved the logic to python. Now it always registers the interval during startup. Much better 😄 

I also found some more information on the why: https://github.com/prusa3d/Prusa-Firmware/pull/4460

jneilliii (approved):

PR #1253, opened 2023-12-02
Update ActivatePrusaHostTimer.md

sarusani:
I got some requests to explain better what this plugin is for.

So I added a link to the homepage in the feature description. This way I can keep users up to date on use cases without having to update this file all the time.

### DIFF
--- a/_plugins/ActivatePrusaHostTimer.md
+++ b/_plugins/ActivatePrusaHostTimer.md
@@ -3,7 +3,7 @@ layout: plugin
 
 id: ActivatePrusaHostTimer
 title: Activate Prusa HostTimer
-description: Plugin for Octoprint to activate Prusa host features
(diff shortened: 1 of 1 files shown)


jneilliii (approved):

jneilliii:
https://github.com/sarusani/OctoPrint-ActivatePrusaHostTimer#set-ready

This is the stupidest thing. It's the same start command as reprint. WTF is Prusa developing without looking at what's out there already. This kind of blinders on solution is really annoying.

foosel:
Next they'll reinvent the wheel...

sarusani:
I'm not 100% sure I understand how they intend to use "set ready". But yes, the only use case I can see, is to use it exactly like the start action... But they also have a "set not ready" option, so I'm probably missing the point. 🤷

PR #1264, opened 2023-12-27
Add xmpp plugin

tobser:
Adds a plugin to send print progress notifications via xmpp


jneilliii (changes requested):
In your plugin, your software update hook should return a dictionary named after your plugin_identifier.

https://github.com/tobser/octoprint-xmpp-plugin/blob/75dc6406075b77326e56cb97801815c24c075198/octoprint_xmpp_plugin/__init__.py#L207C17-L207C33

change that to
```
xmpp_plugin=dict(
```

jneilliii (approved):

PR #1267, opened 2024-01-02
add mqttchambertemperature

synman:
this plugin enables chamber temperature reporting using an mqtt topic subscription


jneilliii:
Not sure if you just missed this, but [here](https://github.com/synman/OctoPrint-MqttChamberTemperature/blob/cf0b13c976bfc7e6823fa93920a79b96edb3617d/octoprint_mqttchambertemperature/__init__.py#L44) you can use `self._settings.get_int(["convertFromFahrenheit"])`

jneilliii (changes requested):
We prefer to have registered plugins have a software update hook, and it appears your plugin is missing this.

synman:
completely forgot to implement the version management block.  Thanks for the full PR review.


update hook:  https://github.com/synman/OctoPrint-MqttChamberTemperature/blob/0.0.1/octoprint_mqttchambertemperature/__init__.py#L132

convertFromFahrenheit should be boolean (doh): https://github.com/synman/OctoPrint-MqttChamberTemperature/blob/0.0.1/octoprint_mqttchambertemperature/__init__.py#L44

jneilliii (approved):

PR #1281, opened 2024-02-26
Added Nozzle-Filament-Validator plugin data to the repo

Rylan-Meilutis:
Added Nozzle-Filament-Validator plugin data to the plugin repo


jneilliii:
You might consider switching your AJAX calls to use https://docs.octoprint.org/en/master/jsclientlib/base.html#OctoPrintClient.simpleApiGet and https://docs.octoprint.org/en/master/jsclientlib/base.html#OctoPrintClient.simpleApiCommand in your js file to avoid issues related to CSR validation. Other than that don't see any issues with merging.

jneilliii:
Another thing I thought of is you might want to implement https://docs.octoprint.org/en/master/modules/printer.html#octoprint.printer.PrinterInterface.job_on_hold in your print started event. I've used this in my BedReady plugin as a real-world example [here](https://github.com/jneilliii/OctoPrint-BedReady/blob/c004a9b9ff912fa0d560bb13aa859374406fb8b2/octoprint_bedready/__init__.py#L145). It's slightly different because I'm using an @ command to intercept the process, but you should be able to do the same in your event callback.

Rylan-Meilutis:
Thanks for the suggestions, I implemented the job_on_hold along with the simple_api_command. when I tried implementing the simpleApiGet, it would fail with a 404 on the page load requiring the function to be called again manually with the refresh button. I was able to get around it be waiting for a 1/2 second before initially loading the data.

PR #1291, opened 2024-04-27
Register Spoolman plugin

mdziekon:
This PR registers a new plugin to the repository, the [Spoolman plugin](https://github.com/mdziekon/octoprint-spoolman).

The plugin is allowing Octoprint users to integrate their Octoprint-based printers with [Spoolman](https://github.com/Donkie/Spoolman/). It is somewhat similar (and somewhat based on) to [Spoolmanager plugin](https://plugins.octoprint.org/plugins/SpoolManager/), however Spoolmanager is limited to Octoprint only, while Spoolman is a generic solution, as long as you can connect to it.

As of right now, the plugin supports absolute minimum, but more features are planned (as seen in the description of the plugin).


jneilliii (approved):

jneilliii:
I've been expecting this plugin after seeing the Spoolman project linked elsewhere.

PR #1296, opened 2024-05-17
Added plugin hologram

MarkSlat:


jneilliii:
ooh...I wonder if your build plate selector is a better option for my bed ready plugin masking selection. I struggled to get it to work but eventually got it and recently noticed alignment issues when reloading the settings in some display resolutions.

_plugins/hologram.md
```diff
@@ -0,0 +1,79 @@
...
+description: A simple OctoPrint plugin to display a render based on the preloaded G-code.
+authors:
+- MarkSlat
+license: AGPLv3
+
+date: 2024-05-16
+
+homepage: https://github.com/MarkSlat/OctoPrint-Hologram
+source: https://github.com/MarkSlat/OctoPrint-Hologram
+archive: https://github.com/MarkSlat/OctoPrint-Hologram/archive/master.zip
```
jneilliii: should be main.zip instead of master.zip. Although this does typically work with GitHub redirects, it's better to be specific.

jneilliii (changes requested):
one minor change to the listing. in your plugin, you may want to consider replacing your ajax calls with https://docs.octoprint.org/en/master/jsclientlib/base.html#OctoPrintClient.simpleApiCommand in order to avoid any potential [CSFR ](https://docs.octoprint.org/en/master/api/general.html#csrf-protection) related issues.

_plugins/hologram.md
MarkSlat: Perfect, I didn't even notice that I will change that 👍

jneilliii (approved):

PR #1310, opened 2024-07-10
Updated OctoLight API information

thomst08:
- Updated OctoLight API information
- Added notification regarding older version of OctoLight as it changes the API

Please let me know if this is not a good use of the ``notices`` file as I am OK if it shouldn't be used in this way.
Thank you.

### DIFF
--- a/_data/notices.yaml
+++ b/_data/notices.yaml
@@ -250,3 +250,10 @@
   text: Version 1.7.1 of this plugin is available from a new maintainer. The release notes of the version by the
     new maintainer can be found at the "Read more..." link below.
   link: https://github.com/dojohnso/OctoPrint-SpoolManager/releases/tag/1.7.1
+- plugin: octolight
+  pluginversions: ["<1.0.0"]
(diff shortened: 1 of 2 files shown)


cp2004:
Would it be easier to keep the API documentation just on your plugin's GitHub readme, and just provide a link from the plugin repository? That should save you making lots of PRs here if you want to change the docs (even small corrections you might find) in the future?

thomst08:
Honestly, I don't see it changing much beyond this but you have a point, if you would prefer I do that then let me know and I will change the file straight way.

thomst08:
@cp2004, thanks for the suggestion, thinking about it, I think you have a point.  The change is done.

jneilliii:
hmm...we've never had plugin notices outside of plugin adoption ones before. Need @foosel input on this one.

thomst08:
Sorry for the delay, I removed the notification addition.
Sorry for causing any issues.

jneilliii (approved):

PR #1315, opened 2024-08-11
new file:   _plugins/filamentmotionsensor.md

tabahi:
The old smart sensor plugin by Royrdan stopped working with the new updates. I have revised the plugin to make it functional again and added new features. There will be more updates soon.


jneilliii (changes requested):
There's a few spots in your repo that needs to be adjusted to where the plugin_identifier hasn't been updated from the original...

plugin_idnetifier needs to be updated to `filamentmotionsensor` to match the id here in the registry.
https://github.com/tabahi/Octoprint-Filament-Motion-Sensor/blob/d926f7cbad44a83a35ed6756eaaa29a25977e042/setup.py#L8

update hook needs to also be updated to return the plugin_identifier `filamentmotionsensor` instead of `smartfilamentsensor`, new user, and new repo name, and pip url. 
https://github.com/tabahi/Octoprint-Filament-Motion-Sensor/blob/d926f7cbad44a83a35ed6756eaaa29a25977e042/filamentmotionsensor/__init__.py#L394-L422

Hit me up in Discord if you have any specific questions.

tabahi:
Should be installable now. One big requirement is python >3.8

jneilliii:
Thanks for those plugin_identifier fixes, one small thing, I think the user property is case sensitive in the update check, so should be `tabahi` [here](https://github.com/tabahi/Octoprint-Filament-Motion-Sensor/blob/995902001f27e4c33a16dd92e60963f09f16296b/octoprint_filamentmotionsensor/__init__.py#L595).

jneilliii:
Oh, and on your plugin repo you need to create a release with a tag that matches the version number in setup.py.

https://github.com/tabahi/Octoprint-Filament-Motion-Sensor/releases

tabahi:
@jneilliii How is it looking now?

jneilliii (approved):

PR #1317, opened 2024-08-27
Add SkipTo plugin

awe-source:
Plugin to start printing at a specified layer.


jneilliii:
@awe-source I see only two issues with your plugin that needs to be addressed before release. First your software update hook needs to return the exact same name as your plugin_identifier here: <https://github.com/awe-source/OctoPrint-SkipTo/blob/master/octoprint_skipto/__init__.py#L76>. skipTo should be SkipTo. Once that change is made on your repo, then you'll need to create a release and tag that matches the version number configured in setup.py (0.1.0) here: <https://github.com/awe-source/OctoPrint-SkipTo/releases>. Once those two things are addressed comment back here and let me know and I can move forward with the process.

awe-source:
Have, hopefully, made the correct changes

jneilliii:
It looks like the software update hook may still be off. Should be like this:

```py
        # Define the configuration for your plugin to use with the Software Update
        # Plugin here. See https://docs.octoprint.org/en/master/bundledplugins/softwareupdate.html
        # for details.
        return {
            "SkipTo": {
                "displayName": "SkipTo",
                "displayVersion": self._plugin_version,

                "type": "github_release",
                "user": "awe-source",
                "repo": "OctoPrint-SkipTo",
                "current": self._plugin_version,

                # update method: pip
                "pip": "https://github.com/awe-source/OctoPrint-skipTo/archive/{target_version}.zip",
            }
        }
```

awe-source:
Fixed?

jneilliii:
Yes, that looks good, thanks.

awe-source:
Um do I need to recreate the release?
Never mind- yes I do…

awe-source:
Can I ask… am I supposed to put things into the release manually?

Eg I don’t fully understand how the whole ecosystem works but a lot of
plugins have 0 active instances” but perhaps this is because a lot of op
are blocked or offline.

But also the “master.zip” is (I assume) the current live/latest version not
a specific version… but I don’t know GH that well either…

Anyway just curious and trying to
Aa. Do the right thing by the ecosystem
And
B. Trying to judge the usefulness/point of this plugin

So far all I have is a bit of offline feedback and GitHub issues

cp2004:
No stats sometimes means there's a mismatch in plugin ids. Your plugin ID should be the same as the name of the .md file in your registration, and in the metadata at the top of this file. If you're seeing comments/issues, then I'd expect to also see some numbers here.

If your software update hook is configured correctly, you should just need to create a release and then it will be picked up by OctoPrint. Installing from the plugin repository (new installs) using the ...master.zip URL will give you whatever is currently on that branch. I'd advise to develop on a separate branch, and then when you're ready to release you merge to master, and then create a release so existing installs can also update.

(Note I've not actually checked if your update hook is correct, but this would have been done as part of the review which I trust 😉)

jneilliii:
Yeah, looks like the plugin ids are consistent throughout. Pretty sure we went through all that in the initial review, and what I'm seeing on your plugin's repo for releases appears to match up correctly with what's in setup.py. The only issue I'm currently seeing on the plugin listing is a missing image, because of capitalization related issues, skipto.png instead of SkipTo.png. I'll correct that here in a minute. I do not understand why the github and anonymous usage stats aren't populating, but I suspect it's something related to capitalization as well. @foosel any ideas on that?

awe-source:
Does it have anything to do with no stats on any releases “there are no
resources associated with this release “ when I goto stats in GitHub, but
if I goto the releases as another user would see it there is a zip and read
me with each one…?!?

I don’t know GitHub that well so not sure if I’ve configured something wrong

jneilliii:
I don't think it's that because I don't do anything special outside of taking a release like you've done.

foosel:
It's nothing you did wrong, there still was a bug in data enrichment here on the plugin repo that I had to touch when this whole stats manipulation happened.

I introduced a new `attributes` field for that, set that to unset by default, and that caused a bug where I looked at it in the enrichment code, as I tried to use it as a list when in fact it was just set to `None`. That caused an exception in all plugin listings that had the default `attributes:` definition, incl. `skipto`. This has now been fixed and the new page build is up, the plugin now has stats:

https://plugins.octoprint.org/plugins/SkipTo/

jneilliii:
287 instances already, not bad @awe-source

foosel:
s/years/instances/g ;)

jneilliii:
oops...edited.

awe-source:
lol
Cool thanks for the info and fix (and laugh)!

PR #1324, opened 2024-09-12
Adopt SpoolManager

WildRikku:
Two bugfixes created, readme etc updated, release 1.7.4 created.

### DIFF
--- a/_data/notices.yaml
+++ b/_data/notices.yaml
@@ -244,9 +244,9 @@
     new maintainer can be found at the "Read more..." link below.
   link: https://github.com/dojohnso/OctoPrint-PrintJobHistory/releases/tag/1.17.1
 - plugin: SpoolManager
-  pluginversions: ["<1.7.1"]
+  pluginversions: ["<1.7.4"]
   versions: []
-  date: 2023-12-06 12:00:00Z
-  text: Version 1.7.1 of this plugin is available from a new maintainer. The release notes of the version by the
+  date: 2024-09-13 12:00:00Z
+  text: Version 1.7.4 of this plugin is available from a new maintainer. The release notes of the version by the
     new maintainer can be found at the "Read more..." link below.
-  link: https://github.com/dojohnso/OctoPrint-SpoolManager/releases/tag/1.7.1
+  link: https://github.com/WildRikku/OctoPrint-SpoolManager/releases/tag/1.7.4
--- a/_data/update_check_overlays.yaml
+++ b/_data/update_check_overlays.yaml
@@ -58,5 +58,5 @@ PrintJobHistory:
   user: dojohnso
   pip: https://github.com/dojohnso/OctoPrint-PrintJobHistory/releases/download/{target_version}/main.zip
 SpoolManager:
-  user: dojohnso
-  pip: https://github.com/dojohnso/OctoPrint-SpoolManager/releases/download/{target_version}/main.zip
+  user: WildRikku
+  pip: https://github.com/WildRikku/OctoPrint-SpoolManager/releases/download/{target_version}/main.zip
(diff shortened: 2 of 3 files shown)


jneilliii:
Small correction needed on the fork. The software update hook needs to be updated to point to your fork/github username, etc. I would re-create the 1.7.4 tag and release after making this change. Otherwise, future updates won't get picked up by users.

https://github.com/WildRikku/OctoPrint-SpoolManager/blob/dd5189cf82f18e95240166fa8f9e13ae3e4888d2/octoprint_SpoolManager/__init__.py#L881-L918

WildRikku:
> Small correction needed on the fork. The software update hook needs to be updated to point to your fork/github username, etc. I would re-create the 1.7.4 tag and release after making this change. Otherwise, future updates won't get picked up by users.
> 
> https://github.com/WildRikku/OctoPrint-SpoolManager/blob/dd5189cf82f18e95240166fa8f9e13ae3e4888d2/octoprint_SpoolManager/__init__.py#L881-L918

Ah, one more thing where that needs to be set. Thanks for pointing me to the right place, I updated it now and updated the release, too.

I also deactivated the prerelease channel / branch as the configuration seemed wrong to me anyway (or I did not understand it). Now that I think about it, I wonder what happens to users having that channel active for the plugin? Will they switch to the stable channel? If they might end up with no channel at all, I'd rather activate it again.

Edit: From briefly looking at the source of the softwareupdate plugin, it seems users would automatically get the stable updates.

WildRikku:
Since I forgot to reference the adoption issue in my commits, I'll link the version diff here as it includes mainly changes for the plugin adoption. Maybe it helps someone in the future to take care all the things. https://github.com/WildRikku/OctoPrint-SpoolManager/compare/WildRikku:OctoPrint-SpoolManager:1.7.3...1.7.4

jneilliii:
> Edit: From briefly looking at the source of the softwareupdate plugin, it seems users would automatically get the stable updates.

yes, because all the channels included the main branch, then when the release was put out on main they would get notified of update. 

OliisGit used teh CI to generate the main.zip and attach to the release assets in order for the badge counting to work on the GitHub Repo. It's way simpler to just link to the automatically generated release assets, but they don't get counted by download stats on GitHub. Here's a super simple example of what I typically do since I don't use badges. 

https://github.com/jneilliii/OctoPrint-JogWheelControl/blob/c3c1c57f6db5931ef596d4e688cadb7905f8fc45/octoprint_jogwheel/__init__.py#L30-L45

Here's more information related to what OllisGit was doing: <https://github.com/OllisGit/OctoPrint-KnowledgeBase/wiki/DownloadCounter>

jneilliii (approved):

WildRikku:
> > Edit: From briefly looking at the source of the softwareupdate plugin, it seems users would automatically get the stable updates.
> 
> yes, because all the channels included the main branch, then when the release was put out on main they would get notified of update.
> 
> OliisGit used teh CI to generate the main.zip and attach to the release assets in order for the badge counting to work on the GitHub Repo. It's way simpler to just link to the automatically generated release assets, but they don't get counted by download stats on GitHub. Here's a super simple example of what I typically do since I don't use badges.
> 
> https://github.com/jneilliii/OctoPrint-JogWheelControl/blob/c3c1c57f6db5931ef596d4e688cadb7905f8fc45/octoprint_jogwheel/__init__.py#L30-L45
> 
> Here's more information related to what OllisGit was doing: https://github.com/OllisGit/OctoPrint-KnowledgeBase/wiki/DownloadCounter

Thank you so much! For now I just generated the zip file manually and created a release like Olli and Doug did, but I noticed the Github configuration files. I will check them out and your suggestions, too :)

PR #1330, opened 2024-10-07
Add segmentedbed plugin

DoubleStrike:
Added my first plugin for registration!  SegmentedBed.


DoubleStrike:
wrong path on image

Issue #1322, opened 2024-09-08
[Abandoned Plugin] PrintJobHistory


jneilliii:
Yeah I don't have the time to work on it so you all can move ahead with whatever is needed to take this over. same for my PrintJobHistory plugin

_Originally posted by @dojohnso in https://github.com/OctoPrint/plugins.octoprint.org/issues/1319#issuecomment-2332344915_

Larry2368:
I don't have a lot of free time and know very little about GITHUB projects, but I have done some code work in my day.  If someone would like to get in touch via some type of private message, I be interested in thinking about taking these projects over.

Larry

KevinVidomski:
I have 20+ years experience in software development including web technologies.  I have a lot of time on my hands right now.  I'm going to look through the two plugins and see if this is something I could handle.

vojtakaniok:
I created [PR](https://github.com/dojohnso/OctoPrint-PrintJobHistory/pull/31) with fix for one [issue](https://github.com/dojohnso/OctoPrint-PrintJobHistory/issues/23).
Let me know if this plugin is owned by someone once again or I can maintain it for a while.
@jneilliii

jneilliii:
If you'd like to adopt the plugin @vojtakaniok, it's just a matter of submitting a PR similar to the previous adoption PR that dojohnso did [here](https://github.com/OctoPrint/plugins.octoprint.org/pull/1248/files).

vojtakaniok:
Thanks for answer !
I will do some cleanup in the repo and pubslih the PR for adopting.

Larry2368:
I would be willing to help when I can.  My time is often limited.  I am new
to using GitHub and I am not sure what language the plug-in is written in,
but I have been coding since the late '70s.

vojtakaniok:
@Larry2368
Thanks for your willingness to help!  You can checkout my repo and help me to maintain it !
https://github.com/vojtakaniok/OctoPrint-PrintJobHistory

jneilliii:
Thanks @vojtakaniok adoption PR has been merged.

Kjubyte:
@vojtakaniok Do you still actively maintain the project? The disabled issue tracker makes me believe you have abandoned it.

vojtakaniok:
@Kjubyte Yes, I'm still maintaining the project. The issues was disabled by default when forking the repo. 
Fixed

PR #1352, opened 2025-02-23
Adopting PrintJobHistory

vojtakaniok:
Adopting PrintJobhistory #1322


_data/notices.yaml
```diff
@@ -237,12 +237,12 @@
     settings. The release notes of the version by the new maintainer can be found at the "Read more..." link below.
   link: https://github.com/thomst08/OctoLight/releases/tag/v0.1.5
 - plugin: PrintJobHistory
-  pluginversions: ["<1.17.1"]
+  pluginversions: ["<1.17.3"]
   versions: []
-  date: 2023-11-27 12:00:00Z
-  text: Version 1.17.1 of this plugin is available from a new maintainer. The release notes of the version by the
+  date: 2025-02-23 12:00:00Z
```
vojtakaniok: Please let me know if reordering of notices.yaml is needed @jneilliii

_plugins/PrintJobHistory.md
```diff
@@ -56,16 +57,17 @@ screenshots:
...
-## *NOTE: this plugin has been abandoned by the original creator and adopted here by a new maintainer*
+#### *NOTE: this plugin was once again abandoned ! Hopefully I will keep support longer*
 
-**This plugin is under new management** and will focus on critical bug fixes to start. Please bear with me as I get acclimated to this new plugin. If you would like to support these new efforts, please consider buying me a coffee or two. Thank you!
+**This plugin is under new management** and will focus on critical bug fixes to start. Please bear with me as I get acclimated to this new plugin.
 
-<a href="https://www.buymeacoffee.com/djohnson.tech" target="_blank"><img src="https://djohnson.tech/images/white-button.png" width=300 /></a>
+If you would like to support these new efforts, please consider buying me a coffee or two. Thank you!
+
+[![ko-fi](https://ko-fi.com/img/githubbutton_sm.svg)](https://ko-fi.com/D1D21B0LA9)
```
jneilliii: it's preferable to have image saved as `/assets/img/plugins/printjobhistory/githubbutton_sm.svg` and linked using that path instead of linking to external images.

_data/notices.yaml
jneilliii: actually need to get feedback from @foosel on this question. not sure if this secondary adoption should be added as a separate entry or replacing this entry is the way to go.

jneilliii (changes requested):

_plugins/PrintJobHistory.md
vojtakaniok: I added it, thanks for comment

_data/notices.yaml
foosel: I think we so far didn't have that case, but replacing feels like it makes more sense here. We don't need to track the history on the notifications, we have the goal to notify people about the latest maintainer change, so replacing the old one it should be.

vojtakaniok: What do you think about reordering ?
Now it is out of order. Should I move it to the end or is it ok like this ?
@foosel @jneilliii

foosel: It isn't needed, but it would make finding things easier for humans, so please reorder, thank you!

vojtakaniok: Solved, Thanks

jneilliii (approved):

PR #1353, opened 2025-02-27
Update gcodeleveling.md to reflect dependencies

Willmac16:
I updated my plugin's description to note an additional system dependency for numpy that I've gotten a couple of reports about


jneilliii (approved):
Yeah, I've ran across that numpy issue in the past. Older versions had the OS dependencies baked into the OctoPi image to allow easier to install, not sure if the upcoming 1.1.0 OctoPi image will have these new dependencies in it or not.

PR #1356, opened 2025-03-23
Added plugin WaveshareUPS

michaelszubartowicz:
This plugin adds a navbar icon to read out the Waveshare UPS HAT.


_plugins/WaveshareUPS.md
```diff
@@ -0,0 +1,73 @@
...
+
+id: WaveshareUPS
+title: OctoPrint-WaveshareUPS
+description: An OctoPrint plugin to support the Waveshare UPS HAT
+authors:
+- Michael Szubartowicz
+license: MIT
+
+# TODO
+date: 2025-03-16
```
jneilliii: update date

jneilliii (changes requested):
just a couple of small adjustments. please update the date and in your plugin, you need to add daemon=True to your Thread [here](https://github.com/michaelszubartowicz/OctoPrint-WaveshareUPS/blob/main/octoprint_WaveshareUPS/__init__.py#L28). If you don't do this, when you try to restart/shutdown OctoPrint it can get in the way of the process.

jneilliii:
curious also, @michaelszubartowicz if you need to add any instructions for enabling I2C, etc. from the system level?

_plugins/WaveshareUPS.md
michaelszubartowicz: date was updated, daemon property is now set. Thank you for the suggestions!

michaelszubartowicz:
> curious also, @michaelszubartowicz if you need to add any instructions for enabling I2C, etc. from the system level?

Yes, you are right. I have added this requirement to the readme as well as the md-file of this pull request.

jneilliii:
Your tagged release needs to match the version number in setup.py. setup.py is showing 0.1.1 but your release tag is V1.0-rc2.

image

michaelszubartowicz:
> Your tagged release needs to match the version number in setup.py. setup.py is showing 0.1.1 but your release tag is V1.0-rc2.
> 
> image

that was a typo, thanks for pointing it out, new tag is now V0.1-rc3

michaelszubartowicz:
@jneilliii no pressure, but is there anything else I can do before this merge request can go through?

jneilliii:
It's still wrong. 

image

The tag should match the release version in setup.py, which is still showing 0.1.1.

https://github.com/michaelszubartowicz/OctoPrint-WaveshareUPS/blob/main/setup.py#L17

So, delete the release, create a new release with tag 0.1.1 and we should be ready to go. Future updates on the plugin should follow that same strategy for the update hook to function properly.

michaelszubartowicz:
> It's still wrong.
> 
> image
> 
> The tag should match the release version in setup.py, which is still showing 0.1.1.
> 
> https://github.com/michaelszubartowicz/OctoPrint-WaveshareUPS/blob/main/setup.py#L17
> 
> So, delete the release, create a new release with tag 0.1.1 and we should be ready to go. Future updates on the plugin should follow that same strategy for the update hook to function properly.

Thank you for this clarification, I did not know how the update hook works. I did as suggested, the new release is now V0.1.1

jneilliii (approved):

PR #1365, opened 2025-04-19
Added plugin cvprinting

Spini11:


Spini11:
Could I ask someone to verify I've added hook for updates correctly into my code? It's at the end of the init file, thank you.
https://github.com/Spini11/OctoPrint-Cvprinting/blob/master/octoprint_cvprinting/__init__.py

jneilliii (commented):
https://github.com/Spini11/OctoPrint-Cvprinting/blob/ee3e29baa267b2ad7d02f84a2004bccbec6ff63f/octoprint_cvprinting/__init__.py#L49: should use https://docs.octoprint.org/en/master/modules/plugin.html#octoprint.plugin.types.OctoPrintPlugin.get_plugin_data_folder.

In your software update check change `updateplugindemo` to `cvprinting`.

I think there's a lot you can do to simplify your settings, you don't need to create your own blueprint for that if you bind properly in knockout. When I have some more time and not looking from my phone I'll give proper review and suggest changes related to that.

Spini11:
Will check the the first two things mentioned. The blueprints for updating settings and getting them is for changing it externally, not for octoprint UI. I could remove it for now from the master branch for now at least though if needed.
Thank you

_plugins/cvprinting.md
```diff
@@ -0,0 +1,107 @@
...
+  # Uncomment the appropriate setting
+
+  #python: ">=2.7,<3" # Python 2 & 3
+  python: ">=3,<4" # Python 3 only
+
+# TODO
+# If any of the below attributes apply to your project, uncomment the corresponding lines. This is MANDATORY!
+    
+attributes:
+#  - cloud  # if your plugin requires access to a cloud to function
```
jneilliii: I think this should be enabled since you have both telegram and discord notification options. 
```
attributes:
  - cloud
 ```

jneilliii (changes requested):
> The blueprints for updating settings and getting them is for changing it externally, not for octoprint UI

Oh, that makes sense then.

I noticed one other thing with your software update check. the pip property would work the way you have it, but would technically download the master branch every new tag release. It might be preferable to use the pattern `https://github.com/Spini11/OctoPrint-Cvprinting/archive/{target_version}.zip` instead. 

That should do it, those changes and the other one I flagged for cloud attributes.

Spini11:
Thank you, I've addressed both issues and created new release containing those updates.

jneilliii (approved):

PR #1378, opened 2025-06-06
Add SpoolTracker plugin

thatguymendel:
This PR adds the SpoolTracker plugin for simply tracking how much filament is left on your spool.


jacopotediosi:
Hi @thatguymendel,

I noticed that you closed your PR. Anyway, I still took a quick look at your plugin [OctoPrint-SpoolTracker](https://github.com/thatguymendel/OctoPrint-SpoolTracker).

In your [`__init__.py`](https://github.com/thatguymendel/OctoPrint-SpoolTracker/blob/2589010f7acc590110980d3682bf03663843badb/octoprint_spooltracker/__init__.py), you’re using the [`SimpleApiPlugin Mixin`](https://docs.octoprint.org/en/master/plugins/mixins.html#simpleapiplugin), which by default does not enforce authentication or authorization checks.
I recommend explicitly implementing such checks, for example using the `octoprint.access.permissions.Permissions` class. 

You can refer to how the OctoPrint built-in plugin `action_command_notification` does it here: 
https://github.com/OctoPrint/OctoPrint/blob/65b6b33b93e5af8270e477ca0e755ab7d00ad4f8/src/octoprint/plugins/action_command_notification/__init__.py#L86-L88

Here you can find a list of predefined permissions you can check:
https://github.com/OctoPrint/OctoPrint/blob/65b6b33b93e5af8270e477ca0e755ab7d00ad4f8/src/octoprint/access/permissions.py#L270-L448

Alternatively, you can define custom permissions using the [`additional_permissions_hook`](https://docs.octoprint.org/en/master/plugins/hooks.html#additional_permissions_hook).

I strongly suggest adding proper permission checks for both the `on_api_get` and `on_api_command` functions.  
Without them, your plugin poses a security risk, as unauthenticated attackers could use its APIs.

jneilliii:
The simplest approach is to add this to the plugin right @jacopotediosi?

https://docs.octoprint.org/en/master/plugins/mixins.html#octoprint.plugin.SimpleApiPlugin.is_api_adminonly

```
def is_api_adminonly(self):
    return True
```

jacopotediosi:
> The simplest approach is to add this to the plugin right @jacopotediosi?
> 
> https://docs.octoprint.org/en/master/plugins/mixins.html#octoprint.plugin.SimpleApiPlugin.is_api_adminonly
> 
> ```
> def is_api_adminonly(self):
>     return True
> ```

Not exactly the same thing. 

`is_api_adminonly` restricts API access to users with the `ADMIN` permission and applies to all API commands uniformly.  
By implementing the `octoprint.access.permissions.Permissions` class instead, you can define the required permissions more granularly and per individual command.

jneilliii:
Yeah, that makes sense. I've done permission implementations in several of my plugins.

PR #1383, opened 2025-07-13
Added Plugin remote_connection

mstarostik:
A plugin to allow connecting to devices on remote serial ports

Remote Connection

Enables connecting to devices on a remote serial port via the network

https://github.com/mstarostik/OctoPrint-Remote_connection/

Not at all

No

No


jneilliii:
nice, very similar to the unregistered [NetworkPrinting](https://github.com/hellerbarde/OctoPrint-Network-Printing) plugin. I noticed that plugin uses a octoprint.util.comm.BufferedReadlineWrapper and wonder if there's a benefit of doing so? Otherwise, this plugin seems fine to me.

jneilliii:
Might not be a bad idea to include a screenshot.

jacopotediosi:
I took a look, looks good to me from a security standpoint :+1: :smile:

mstarostik:
Thanks for the quick reply.
I wasn't aware of the [NetworkPrinting](https://github.com/hellerbarde/OctoPrint-Network-Printing) plugin as it's not registered and also didn't show up in Google.
Anyway, as it had its last commit 5 years ago and doesn't offer its own settings I would assert that my version does offer a little more functionality.
I did add a BufferedReadlineWrapper though as it looks like a good idea to do so indeed. Also actually passing connection parameters now.

Screenshots added. No, I don't have any bad feelings about having the RFC1918 IP used for testing in there.

jneilliii:
Looks like you released version update, but it's not the "latest".

Screenshot_20250713-202318

jneilliii (approved):

Issue #1417, opened 2026-03-09
[Abandoned Plugin] OctoPrint-PSUControl-Shelly


starlord1980:
### Which Plugin?

https://plugins.octoprint.org/plugins/psucontrol_shelly/

### Link to ticket on plugin's issue tracker confirming abandonment

Its on the front page "This repository was archived by the owner on Feb 16, 2026. It is now read-only."

### Is there someone open to adopting it?

Not that I know of, but I have some corrections in the code that need to be made as it currently doesnt work however found a way to fix it through a manual edit and manual install

starlord1980:
I'm not a developer and have no interest in adopting it. Would just like others to have the benefit of the fix.
The plug will only turn on once and then not respond to any further commands... heres an AI description of the fix that works, but not for anyone else who might want to use this plugin unless someone updates it with these fixes:

---------------------

The problem with the existing code was a **logical deadlock** caused by a safety mechanism called `self.transition`. This variable was designed to prevent the plugin from sending multiple commands at the same time, but it lacked a reliable "reset" trigger.

Here is the step-by-step breakdown of how the code was breaking:

### 1. The "Lock" is Set

When you clicked the button to turn the plug **ON**, the plugin ran the `change_psu_state` function.

* The code checked: "Are we already transitioning?" (`if self.transition: return`). Since it was the first time, the answer was "No."
* The code then set `self.transition = True`. This acted like a **padlock**, telling the plugin to ignore any future commands until the lock was removed.
* The command was sent to the Shelly, and the plug turned on.

### 2. The "Unlock" Mechanism is Missing

In the original code, the **only** way to set `self.transition` back to `False` (to remove the lock) was inside a different function called `get_psu_state`.

* `get_psu_state` is the function that asks the Shelly "Are you actually on right now?"
* The code assumed that OctoPrint would call this "Status Check" immediately after the "Turn On" command.

### 3. The Deadlock

If your main **PSUControl** settings were set to use **"Internal"** sensing (or no sensing at all), OctoPrint would never call `get_psu_state`.

* Because that function was never called, `self.transition` stayed `True` forever.
* The next time you tried to turn the plug OFF, the plugin checked the lock, saw it was still `True`, and simply logged: *"Still in transition... not sending command."*

### Summary of the Fixes

To fix this, we made two key changes to the logic:

1. **Immediate Reset:** We told the code to set `self.transition = False` as soon as the command is successfully sent for local devices, rather than waiting for a status check.
[...]

PR #1420, opened 2026-03-16
Adopt PSUControl-Shelly plugin

Triodes:
Adopting PSUControl-Shelly after abandonment: #1417 

@jneilliii I tried making this PR using previous adoption as suggestion. But I would appreciate a double check by someone who actually knows what he's doing 😆 

I noticed that the `_notices.yaml` (pun intended) contains messages about new maintainers. I added that one as well, but I'm not sure if I was supposed to do that myself


jneilliii:
Thanks, just had one small change needed on the notices. Awaiting the build validation and this is good to merge.

PR #1423, opened 2026-03-23
Added plugin Proxy4Artisan

coc815:
Added plugin Proxy4Artisan

Proxy4Artisan

It enables OctoPrint to correctly interpret messages Snapmaker Artisan.

https://github.com/coc815/OctoPrint-Proxy4Artisan

No

No

No


jneilliii:
I'm curious why the virtual serial here jsut to rewrite gcode command information. I feel like it would be better served to utilize gcode protocol hooks, which can both adapt commands being sent and received. 

https://docs.octoprint.org/en/main/plugins/hooks.html#octoprint-comm-protocol-gcode-phase

coc815:
@jneilliii 
You were absolutely right! This is so much easier!!
Thank you for the hint :)

jacopotediosi:
@coc815, thanks for your contribution, your plugin looks great!

For inclusion in the plugin repository, we usually require that log lines (and comments, if desired) be in English. This allows other non-German users to understand and contribute to your work. Could you please update them?

coc815:
@jacopotediosi all english now

jneilliii:
Thanks @jacopotediosi that was going to be something that I mentioned when performing a full review. The other thing I noticed was that version 0.1.0 was not released on GitHub yet. That's a good thing, as it will allow @coc815 to adjust the code prior to release and official review.

coc815:
I changed the answer to the question, if any kind of genAI was involved from "Copilot" to "No", because after I switched the code to utilize gcode protocol hooks, the code is truely mine ;)

@jneilliii But I must admit, that I am a bit lost now. you mentioned version 0.1.0. If you meant, that I should switch versioning back to 0.1.0, that is done now. If you meant something different, would you be so kind and explain a bit further? I am really new to all of that.

Thank you all for your patience

jacopotediosi:
@coc815 Automatic updates in OctoPrint plugins are managed by the `get_update_information` function, which looks for the latest release in your GitHub repository.

Currently, you haven't published your first release yet:

image

Usually the workflow is: implement features and fix bugs and - only when you're ready to release an update to the public - increment the version number in `pyproject.toml` and create a GitHub Release. Usually version numbers only go upwards, never backwards.

When you create a release, make sure to create a tag with the corresponding version number, e.g. `0.1.0`. This is something you can do on the same page while creating a Release.

What @jneilliii was asking you to do is to publish your first release.

For further information, you can refer to the [GitHub documentation about releases](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases).

coc815:
@jacopotediosi  Thank's for pointing me in the right direction. I published my first release

_plugins/Proxy4Artisan.md
```diff
@@ -0,0 +1,115 @@
...
+#  - free-tier  # if your plugin has a free tier
+
+---
+
+This plugin enables OctoPrint to handle Snapmaker Artisan
+
+What the plugin does
+
+- manipulates temperature responses, for OctoPrint to be able to show a bed temperature: 'B0:' -> 'B:' (just the first occurrence per line)
+- Filament runout will trigger @Pause
```
jneilliii: This technically no longer applies, rather instead of triggering a `@PAUSE` it will just pause the print.

jneilliii (changes requested):
One thing to note, you might want to add those releases with the version number as the release title. The updates will still work based on the version tag, but the software update plugin will show that title as the "available" version. As you have it now, that looks like this. 

image

You should be able to edit the title on the release if you just want to change it. 

I just have one minor edit on the plugin description now with this new gcode phasing approach.

_plugins/Proxy4Artisan.md
coc815: description is updated

coc815:
@jneilliii I changed the release title and the description

coc815:
Thank you all :)

PR #1427, opened 2026-04-13
Add InfillCode plugin

st7ma784:

InfillCode
Having filament run out can be a pain and waste perfectly good prints and filament. This plugin edits the spacing of linear infill layers so that a camera can detect where the print got to, and create gcode to resume from where the print stopped! 

https://github.com/st7ma784/infillcoder

Yes - Claude has helped with some of the logic

No, 

No

#### Further notes

InfillCode embeds a per-layer fingerprint into 3D print infill line spacing. When a print fails, a webcam snapshot identifies the last completed layer and generates a resume GCode file automatically. I quite like using Arc-welder (shout out!!) so this is set to auto-process .aw.gcode uploads, startup bed scan, mid-print health monitoring with rolling score, optional auto-pause, one-click resume button.

### DIFF
--- a/_plugins/infillcode.md
+++ b/_plugins/infillcode.md
@@ -1,0 +1,34 @@
+---
+layout: plugin
+
+id: infillcode
+title: InfillCode
+description: Embeds per-layer fingerprints into 3D print infill line spacing. When a print fails, a webcam snapshot identifies the last good layer and generates a resume GCode file automatically. Includes startup bed scanning, mid-print health monitoring, and one-click resume.
+author: Dr Steve Mander
+license: MIT
+date: 2026-04-13
+
+homepage: https://github.com/st7ma784/infillcoder
+source: https://github.com/st7ma784/infillcoder
+archive: https://github.com/st7ma784/infillcoder/releases/download/v0.2.0/infillcode-0.2.0.tar.gz
+
+tags:
+- gcode
+- resume
+- recovery
+- failure recovery
+- fingerprint
+- webcam
+- automation
+- filament runout
+
+compatibility:
+  octoprint:
+  - 1.4.0
+  os:
+  - linux
+  - windows
+  - macos
+  - freebsd
+  python: ">=3.7"
+---
--- a/_plugins/infillcode.md
+++ b/_plugins/infillcode.md
@@ -38,6 +38,6 @@ The Plugin embeds per-layer fingerprints into 3D print infill line spacing. When
 
 If you like it, I would be thankful about a cup of coffee :)
 
-[![paypal](/assets/img/plugins/bedlevelvisualizer/paypal-with-text.png)](https://www.paypal.com/donate/?business=FM3XGWAZJNGXU&no_recurring=0&currency_code=GBP)
+[![paypal](/assets/img/plugins/infillcoder/paypal-donate.svg)](https://www.paypal.com/donate/?business=FM3XGWAZJNGXU&no_recurring=0&currency_code=GBP)
 
 For implementation details please visit the [homepage]({{ page.homepage | absolute_url }}).
--- a/_plugins/infillcode.md
+++ b/_plugins/infillcode.md
@@ -10,7 +10,7 @@ date: 2026-04-13
 
 homepage: https://github.com/st7ma784/infillcoder
 source: https://github.com/st7ma784/infillcoder
-archive: https://github.com/st7ma784/infillcoder/releases
+archive: https://github.com/st7ma784/infillcoder/archive/refs/heads/main.zip
 tags:
 - gcode
 - resume
--- a/_plugins/infillcode.md
+++ b/_plugins/infillcode.md
@@ -10,8 +10,7 @@ date: 2026-04-13
 
 homepage: https://github.com/st7ma784/infillcoder
 source: https://github.com/st7ma784/infillcoder
-archive: https://github.com/st7ma784/infillcoder/releases/download/v0.2.0/infillcode-0.2.0.tar.gz
-
+archive: https://github.com/st7ma784/infillcoder/releases
 tags:
 - gcode
 - resume
@@ -31,4 +30,14 @@ compatibility:
   - macos
   - freebsd
   python: ">=3.7"
----
+
+
+The Plugin embeds per-layer fingerprints into 3D print infill line spacing. When a print fails, a webcam snapshot identifies the last good layer and generates a resume GCode file automatically. Includes startup bed scanning, mid-print health monitoring, and one-click resume.
+
+#### Support my Efforts
+
+If you like it, I would be thankful about a cup of coffee :)
+
+[![More coffee, more code](https://img.shields.io/badge/Donate-PayPal-green.svg)](https://www.paypal.com/donate/?business=FM3XGWAZJNGXU&no_recurring=0&currency_code=GBP)
+
+For implementation details please visit the [homepage]({{ page.homepage | absolute_url }}).
--- a/_plugins/infillcode.md
+++ b/_plugins/infillcode.md
@@ -30,7 +30,7 @@ compatibility:
   - macos
   - freebsd
   python: ">=3.7"
-
+---
 
 The Plugin embeds per-layer fingerprints into 3D print infill line spacing. When a print fails, a webcam snapshot identifies the last good layer and generates a resume GCode file automatically. Includes startup bed scanning, mid-print health monitoring, and one-click resume.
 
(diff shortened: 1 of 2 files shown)


jacopotediosi:
@st7ma784 First of all, thank you for your contribution!

Unfortunately, I'm unable to review your plugin until the [PR template](https://github.com/OctoPrint/plugins.octoprint.org/blob/gh-pages/.github/PULL_REQUEST_TEMPLATE.md) has been filled out - it should appear automatically when opening a PR in this repository.

That template helps us reviewers work more efficiently and ensures that plugin maintainers have read all the necessary documentation and accepted the terms.

We've noticed that a few recent PRs, also from other users, have been missing it, so just a friendly reminder - please take a moment to complete it, and I'll be happy to proceed with the review. Thanks!

_plugins/infillcode.md
```diff
@@ -0,0 +1,34 @@
...
+id: infillcode
+title: InfillCode
+description: Embeds per-layer fingerprints into 3D print infill line spacing. When a print fails, a webcam snapshot identifies the last good layer and generates a resume GCode file automatically. Includes startup bed scanning, mid-print health monitoring, and one-click resume.
+author: Dr Steve Mander
+license: MIT
+date: 2026-04-13
+
+homepage: https://github.com/st7ma784/infillcoder
+source: https://github.com/st7ma784/infillcoder
+archive: https://github.com/st7ma784/infillcoder/releases/download/v0.2.0/infillcode-0.2.0.tar.gz
```
jacopotediosi: `archive` should not be tied to a specific version

_plugins/infillcode.md
```diff
@@ -0,0 +1,34 @@
...
+compatibility:
+  octoprint:
+  - 1.4.0
+  os:
+  - linux
+  - windows
+  - macos
+  - freebsd
+  python: ">=3.7"
+---
```
jacopotediosi, line 33: Missing page content after `---`, please see how other plugins do

jacopotediosi (changes requested):

_plugins/infillcode.md
```diff
@@ -10,8 +10,7 @@ date: 2026-04-13
 
 homepage: https://github.com/st7ma784/infillcoder
 source: https://github.com/st7ma784/infillcoder
-archive: https://github.com/st7ma784/infillcoder/releases/download/v0.2.0/infillcode-0.2.0.tar.gz
-
+archive: https://github.com/st7ma784/infillcoder/releases
```
jneilliii: Should be something like this to not be version specific.
```
https://github.com/st7ma784/infillcoder/archive/refs/heads/main.zip
```

_plugins/infillcode.md
```diff
@@ -0,0 +1,43 @@
...
+compatibility:
+  octoprint:
+  - 1.4.0
+  os:
+  - linux
+  - windows
+  - macos
+  - freebsd
+  python: ">=3.7"
+
```
jneilliii: missing closing header indicator....add this here.
```
---
```

_plugins/infillcode.md
```diff
@@ -0,0 +1,43 @@
...
+
+The Plugin embeds per-layer fingerprints into 3D print infill line spacing. When a print fails, a webcam snapshot identifies the last good layer and generates a resume GCode file automatically. Includes startup bed scanning, mid-print health monitoring, and one-click resume.
+
+#### Support my Efforts
+
+If you like it, I would be thankful about a cup of coffee :)
+
+[![More coffee, more code](https://img.shields.io/badge/Donate-PayPal-green.svg)](https://www.paypal.com/donate/?business=FM3XGWAZJNGXU&no_recurring=0&currency_code=GBP)
```
jneilliii: we prefer to not have images load from external website/sources. look at how other plugins (all of mine include paypal links for example) are saving these and referencing.

_plugins/infillcode.md
```diff
@@ -0,0 +1,43 @@
...
+  python: ">=3.7"
+---
+
+The Plugin embeds per-layer fingerprints into 3D print infill line spacing. When a print fails, a webcam snapshot identifies the last good layer and generates a resume GCode file automatically. Includes startup bed scanning, mid-print health monitoring, and one-click resume.
+
+#### Support my Efforts
+
+If you like it, I would be thankful about a cup of coffee :)
+
+[![paypal](/assets/img/plugins/bedlevelvisualizer/paypal-with-text.png)](https://www.paypal.com/donate/?business=FM3XGWAZJNGXU&no_recurring=0&currency_code=GBP)
```
jneilliii: should be your own copy of the image you want to use and linked from a folder `/assets/img/plugins/infillcode/ ` as part of this PR.

jacopotediosi (changes requested):
I'm trying to write this comment as calmly as I can, because I want to be constructive, but I'm also quite frustrated.

This plugin reads as AI-generated slop, and submissions like this aren't respectful of reviewers' time: generating it takes a few prompts, while reviewing it and articulating what's wrong takes hours.

A standard OctoPrint plugin follows a well-defined structure, based on the [cookiecutter template](https://github.com/OctoPrint/cookiecutter-octoprint-plugin) and the [official documentation](https://docs.octoprint.org/en/dev/plugins/gettingstarted.html). This plugin's repository does not. It has an unusual nested structure: the top level is a wrapper for building documentation and the site that serves it, plus some library files. The actual plugin code lives under `octoprint_plugin`, but even that folder isn't self-contained, because the `core` folder is pulled in from the top level into release artifacts via a build step.

I'm not sure this layout is compatible with installation via the Plugin Manager at all, given that it expects a path containing the entire plugin code in one place, without intermediate build steps. It certainly won't work via the URL declared in the `archive` field of this PR's manifest, which points to the `main` branch - i.e., the entire repository, not just the plugin folder.

Beyond that, there are numerous issues:

- Some files (`README.md`, `LICENSE`) are duplicated between the top level and the inner plugin folder, with nothing visible that keeps them in sync.
- The "full documentation" link in the `README.md` returns 404.
- The README's install instructions claim the plugin can be installed via `pip install infillcode`, but the package is not registered on `pypi.org`. This smells like an AI hallucination.
- Both `requirements.txt` and `pyproject.toml` are present, which is an odd combination, and again nothing seems to keep the two dependency lists aligned.
- Inside `octoprint_plugin`, `setup.py` and `pyproject.toml` don't match the expected structure, miss several required fields, and contain other oddities (e.g., `long_description` is set to the contents of `../README.md`, a path that escapes the package directory).
- In the plugin's `__init__.py`, the SimpleApi endpoints are only placeholders, and the plugin update hook is not implemented.
- Under `octoprint_plugin/infillcode/templates/` there is an unexpected path `static/infillcode.js`, which from reading it appears to be a stale, broken copy of `octoprint_plugin/infillcode/static/infillcode.js`.

There are many other parts of the code that don't add up. The resume logic in particular [emits](https://github.com/st7ma784/infillcoder/blame/781b3a9fa468fe14d0499cf519ac15dc6b659388/core/resume.py#L232-L244) the following hardcoded G-code commands that I find questionable:

```py
# Position to resume Z with a hop
hop_z = round(resume_z + z_hop_mm, 4)
init_lines.append("; --- Position to resume layer ---\n")
init_lines.append("G90 ; absolute positioning\n")
init_lines.append("M83 ; relative extrusion\n")
init_lines.append(f"G1 Z{hop_z:.4f} F600 ; lift to clear ooze\n")
init_lines.append("G1 X5 Y5 F6000 ; move to prime corner\n")
init_lines.append(f"G1 Z{resume_z:.4f} F600 ; descend to resume Z\n")
# Short prime purge
init_lines.append(f"G1 E{prime_length_mm:.1f} F{prime_feedrate:.0f} ; prime nozzle\n")
init_lines.append("G92 E0 ; reset extruder\n")
init_lines.append("; --- Resume layers ---\n")
```

Some modes are set unconditionally (e.g. `G90`, `M83`), and then the file body from the resume layer onwards is concatenated and will run under these modes - not necessarily the ones that were active during the original print. This can result, for example, in the resume layer extruding far more than expected because of the switch from absolute to relative extrusion, producing a large blob at the resume point. There's also that absolute `G1 Z<resume_z>`: on many printers, after a reset the firmware loses the Z reference, and in response to that command the toolhead can end up anywhere.

Overall, I believe this plugin should be rejected from the official OctoPrint plugin repository. In my opinion, the author doesn't have a working understanding of what they're submitting, not least because this code structure is basically unmaintainable and the code unreadable.

It genuinely pains me to write this, and I'm deeply sorry that yet another opportunity for a real contribution to the OctoPrint ecosystem has been lost to AI slop.

foosel:
I sadly have to agree here. And I hate having to continue to close requests like this. But in this day and age of stochastic parrots puking out code faster than it can be reviewed by humans, I fear this won't be the last time.

PR #1428, opened 2026-04-17
Add OctoPrint-MFA-Passkeys to registry

daedalas1981:

OctoPrint-MFA-Passkeys

It brings native WebAuthn/Passkey authentication to OctoPrint. It allows users to securely authenticate through the native login interface using Biometrics (Windows Hello, Face ID, Touch ID) and hardware security keys (YubiKey), effectively replacing passwords.

https://github.com/daedalas1981/Octoprint-MFA-Passkeys

Yes. I used Google's DeepMind AI coding assistant (Antigravity) as a pair-programming partner to help me architect the WebAuthn integrations, frontend injections, and the HAProxy automation bash scripts. I have thoroughly tested, reviewed, and deeply understand the resulting architecture and am fully maintaining it myself!

No.

No. Authentication and credential negotiation happen 100% locally between the user's browser, hardware, and the OctoPrint server.

#### Further notes
Excited to share this with the community! Thank you for the review.

### DIFF
--- a/_plugins/mfa_passkeys.md
+++ b/_plugins/mfa_passkeys.md
@@ -12,7 +12,8 @@ date: 2026-04-16
 
 homepage: https://github.com/daedalas1981/Octoprint-MFA-Passkeys
 source: https://github.com/daedalas1981/Octoprint-MFA-Passkeys
-archive: https://github.com/daedalas1981/Octoprint-MFA-Passkeys/archive/refs/tags/v0.2.10.zip
+archive: https://github.com/daedalas1981/OctoPrint-Passkeys/archive/main.zip
+
 
 follow_dependency_links: false
 
@@ -31,7 +32,6 @@ compatibility:
   python: ">=3.9,<4"
   octoprint:
   - 1.8.0
-  - 1.9.0
 
 ---
 OctoPrint-Passkeys is a unified authentication plugin that brings modern, passwordless WebAuthn security directly to your printer.
--- a/_plugins/mfa_passkeys.md
+++ b/_plugins/mfa_passkeys.md
@@ -1,0 +1,43 @@
+---
+layout: plugin
+
+id: mfa_passkeys
+title: OctoPrint-MFA-Passkeys
+description: Seamless biometric and hardware security for OctoPrint. Drop-in WebAuthn support that integrates directly into the native login interface, replacing passwords with Face ID, Touch ID, and security keys.
+authors:
+- Daedalas1981
+license: AGPLv3
+
+date: 2026-04-16
+
+homepage: https://github.com/daedalas1981/Octoprint-MFA-Passkeys
+source: https://github.com/daedalas1981/Octoprint-MFA-Passkeys
+archive: https://github.com/daedalas1981/Octoprint-MFA-Passkeys/archive/refs/tags/v0.2.10.zip
+
+follow_dependency_links: false
+
+tags:
+- security
+- login
+- authentication
+- mfa
+- passkeys
+- webauthn
+- touch id
+- face id
+- biometric
+
+compatibility:
+  python: ">=3.9,<4"
+  octoprint:
+  - 1.8.0
+  - 1.9.0
+
+---
+OctoPrint-Passkeys is a unified authentication plugin that brings modern, passwordless WebAuthn security directly to your printer.
+
+### Powerful Security
+* **Native Integration:** Replaces passwords effortlessly on the native login screen.
+* **Biometrics:** Authenticate instantly with Windows Hello, Face ID, or Touch ID.
+* **Hardware Keys:** YubiKey and U2F compatible.
+* **Automated Security:** Contains a built-in automated TLS/HAProxy configuration deployment to instantly generate 10-year Chrome-compliant Secure Subject Alternative Name (SAN) Certificates right on your Pi.


_plugins/mfa_passkeys.md
```diff
@@ -0,0 +1,43 @@
...
+description: Seamless biometric and hardware security for OctoPrint. Drop-in WebAuthn support that integrates directly into the native login interface, replacing passwords with Face ID, Touch ID, and security keys.
+authors:
+- Daedalas1981
+license: AGPLv3
+
+date: 2026-04-16
+
+homepage: https://github.com/daedalas1981/Octoprint-MFA-Passkeys
+source: https://github.com/daedalas1981/Octoprint-MFA-Passkeys
+archive: https://github.com/daedalas1981/Octoprint-MFA-Passkeys/archive/refs/tags/v0.2.10.zip
```
jacopotediosi: This shouldn't point to a precise version, but to your release branch. It usually is something like `https://github.com/daedalas1981/Octoprint-MFA-Passkeys/archive/main.zip`.

_plugins/mfa_passkeys.md
```diff
@@ -0,0 +1,43 @@
...
+- webauthn
+- touch id
+- face id
+- biometric
+
+compatibility:
+  python: ">=3.9,<4"
+  octoprint:
+  - 1.8.0
+  - 1.9.0
```
jacopotediosi: `compatibility.octoprint` is the minimum compatible OctoPrint version, so it should just be `1.8.0`.

jacopotediosi (changes requested):

_plugins/mfa_passkeys.md
daedalas1981: Fixed, thanks!

_plugins/mfa_passkeys.md
daedalas1981: Fixed, thanks!

jacopotediosi:
Hi @daedalas1981 and thanks for your contribution!

I reviewed your plugin source and below are my initial thoughts. This plugin is **highly security relevant**, so expect additional in-depth reviews from me and maybe other contributors/maintainers to follow, and please understand that this PR may take significantly longer than other plugins.

### Packaging

In OctoPrint's modern plugin packaging system, `setup.py` usually contains only:

```py
import setuptools
# we define the license string like this to be backwards compatible to setuptools<77
setuptools.setup(license="AGPLv3")
```

And everything else is defined in `pyproject.toml` instead. If you need a reference example, you can look at how it's done in the [cookiecutter template](https://github.com/OctoPrint/cookiecutter-octoprint-plugin/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/pyproject.toml) o read the ["Migrating to pyproject.toml"](https://docs.octoprint.org/en/dev/plugins/migration_pyproject_toml.html) documentation.

Please also note that you need to provide a valid email in `plugin_author_email`, instead of the current `noreply@example.com`.

### XSS

[Here](https://github.com/daedalas1981/Octoprint-MFA-Passkeys/blob/fa899cf773b893cef748c8dd0d290eb55d9a3738/octoprint_mfa_passkeys/static/js/passkeys.js#L163-L165) you have a basic Cross-Site Scripting vulnerability since you concatenate user-supplied input into the page's HTML without sanitization.

You must sanitize the output of arbitrary data before concatenation, for example by using `_.escape()`.

### Path Traversal
[Here](https://github.com/daedalas1981/Octoprint-MFA-Passkeys/blob/fa899cf773b893cef748c8dd0d290eb55d9a3738/octoprint_mfa_passkeys/__init__.py#L610-L640) you have a path traversal vulnerability. Please check `candidate_paths` against an allowlist to prevent reading arbitrary files.

### Use of removed functions

In OctoPrint 2.0.0 (OctoPrint's `dev` branch) we are removing features that have been deprecated for years. Your plugin still uses some of them in a few places, and therefore will not work:

- Use of `user.asDict()` [here](https://github.com/daedalas1981/Octoprint-MFA-Passkeys/blob/fa899cf773b893cef748c8dd0d290eb55d9a3738/octoprint_mfa_passkeys/__init__.py#L538-L539): this function is no longer available. None of those `elif` fallbacks are necessary since `user.as_dict()` is guaranteed to exist starting from OctoPrint 1.4.0, which is way behind your declared minimum compatibility.

- Use of `current_user.is_authenticated` and `is_admin` [here](https://github.com/daedalas1981/Octoprint-MFA-Passkeys/blob/fa899cf773b893cef748c8dd0d290eb55d9a3738/octoprint_mfa_passkeys/__init__.py#L446-L453): see the ["Migrating to OctoPrint 2.0.0" guide](https://docs.octoprint.org/en/dev/plugins/migration_2_0_0.html#changes-in-octoprint-access-users). You can likely replace `current_user.is_authenticated` with `not current_user.is_anonymous()` and `is_admin` with `current_user.has_permission(Permissions.ADMIN)`.

jacopotediosi:
@jneilliii could you please assign this to me? Thanks

daedalas1981:
> Hi @daedalas1981 and thanks for your contribution!
> 
> I reviewed your plugin source and below are my initial thoughts. This plugin is **highly security relevant**, so expect additional in-depth reviews from me and maybe @jneilliii to follow, and please understand that this PR may take significantly longer than other plugins.
> 
> ### Packaging
> In OctoPrint's modern plugin packaging system, `setup.py` usually contains only:
> 
> ```python
> import setuptools
> # we define the license string like this to be backwards compatible to setuptools<77
> setuptools.setup(license="AGPLv3")
> ```
> 
> And everything else is defined in `pyproject.toml` instead. If you need a reference example, you can look at how it's done in the [cookiecutter template](https://github.com/OctoPrint/cookiecutter-octoprint-plugin/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/pyproject.toml) o read the ["Migrating to pyproject.toml"](https://docs.octoprint.org/en/dev/plugins/migration_pyproject_toml.html) documentation.
> 
> Please also note that you need to provide a valid email in `plugin_author_email`, instead of the current `noreply@example.com`.
> 
> ### XSS
> [Here](https://github.com/daedalas1981/Octoprint-MFA-Passkeys/blob/fa899cf773b893cef748c8dd0d290eb55d9a3738/octoprint_mfa_passkeys/static/js/passkeys.js#L163-L165) you have a basic Cross-Site Scripting vulnerability since you concatenate user-supplied input into the page's HTML without sanitization.
> 
> You must sanitize the output of arbitrary data before concatenation, for example by using `_.escape()`.
> 
> ### Path Traversal
> [Here](https://github.com/daedalas1981/Octoprint-MFA-Passkeys/blob/fa899cf773b893cef748c8dd0d290eb55d9a3738/octoprint_mfa_passkeys/__init__.py#L610-L640) you have a path traversal vulnerability. Please check `candidate_paths` against an allowlist to prevent reading arbitrary files.
> 
> ### Use of removed functions
> In OctoPrint 2.0.0 (OctoPrint's `dev` branch) we are removing features that have been deprecated for years. Your plugin still uses some of them in a few places, and therefore will not work:
> 
> * Use of `user.asDict()` [here](https://github.com/daedalas1981/Octoprint-MFA-Passkeys/blob/fa899cf773b893cef748c8dd0d290eb55d9a3738/octoprint_mfa_passkeys/__init__.py#L538-L539): this function is no longer available. None of those `elif` fallbacks are necessary since `user.as_dict()` is guaranteed to exist starting from OctoPrint 1.4.0, which is way behind your declared minimum compatibility.
> * Use of `current_user.is_authenticated` and `is_admin` [here](https://github.com/daedalas1981/Octoprint-MFA-Passkeys/blob/fa899cf773b893cef748c8dd0d290eb55d9a3738/octoprint_mfa_passkeys/__init__.py#L446-L453): see the ["Migrating to OctoPrint 2.0.0" guide](https://docs.octoprint.org/en/dev/plugins/migration_2_0_0.html#changes-in-octoprint-access-users). You can likely replace `current_user.is_authenticated` with `not current_user.is_anonymous()` and `is_admin` with `current_user.has_permission(Permissions.ADMIN)`.

Hi @jacopotediosi , thanks for your quick response!

### UPDATED
A few of your comments were a little harder to convert. I have tested the update in place via the plugin manager which worked as expected. I will need time to run the entire plugin again from a fresh install of Octopi in case of any unforeseen issues with the new updates. Will follow-up when completed.

jacopotediosi:
Hi @daedalas1981,

I'm always sorry to give a negative opinion on a contribution, but I have a bad gut feeling about this plugin.

In OctoPrint, it is possible to create plugins that make authentication **more secure** by adding a second factor **on top of** the password, using the [`MfaPlugin`](https://docs.octoprint.org/en/main/plugins/mixins.html#octoprint.plugin.MfaPlugin) Mixin.

However, this is not the case with your plugin, which does not use that Mixin and instead **replaces** the authentication method entirely. This is also why I think the “MFA” in your plugin name is misleading.

My perspective, as a security consultant, is that it is too dangerous to delegate such a critical responsibility to a third-party plugin.

The intended and documented way to replace authentication in OctoPrint is to configure an authentication proxy, so that the responsibility for the alternative authentication method lies with the proxy and is well separated from OctoPrint and its codebase. An example of how to set up such a configuration was recently discussed in issue https://github.com/OctoPrint/OctoPrint/issues/5279.

My concern is further reinforced by recurring code quality issues in your plugin (e.g., imports inside functions and other patterns that look AI-generated and that I believe are going to be difficult for a human to maintain in the long run) and by structural problems, such as the use of undocumented functionality like rewriting the user's Flask session or completely replacing the `<head>` tag of the login page.

Please note that this is only my personal opinion as a reviewer and may differ from the official OctoPrint stance, which I expect will come in the next few days/weeks. I am just a contributor to this project and have no decision-making authority.

jneilliii:
Currently, I completely agree with @jacopotediosi on the quality of code that the AI has made here. The MFAPlugin mixin approach is definitely the better way to go for MFA, and using things like https://github.com/daedalas1981/Octoprint-MFA-Passkeys/blob/5526db9b9a3468435b5dc22513abf4bb91596e1c/octoprint_mfa_passkeys/__init__.py#L93-L122 are not ideal. 

If the plugin is otherwise not actually MFA as eluded by name and really is a replacement of the login mechanism completely, there are other better approaches as well for this utilizing [octoprint-access-users-factory](https://docs.octoprint.org/en/main/plugins/hooks.html#octoprint-access-users-factory) and [UIPlugin mixin](https://docs.octoprint.org/en/main/plugins/mixins.html#uiplugin) potentially.  

Also, due to the use of bash shell scripting, the compatibility flags for OS needs to specify linux only.

jneilliii:
> If the plugin is otherwise not actually MFA as eluded by name and really is a replacement of the login mechanism completely, there are other better approaches as well for this utilizing [octoprint-access-users-factory](https://docs.octoprint.org/en/main/plugins/hooks.html#octoprint-access-users-factory) and [UIPlugin mixin](https://docs.octoprint.org/en/main/plugins/mixins.html#uiplugin) potentially.

I can't speak for the security aspects of this implementation, but I did remember an example of this that hasn't registered in the plugin repo: https://github.com/JoveToo/OctoPrint-MFA

foosel:
As much as it pains me to do this, I have to deny registration for this plugin.

As already outlined by @jacopotediosi and @jneilliii, to reiterate myself: 

1. There are heavy doubts about the code quality and maintainability of this plugin, most likely due to the heavy (and from the looks of the commit history frankly looking quite uncontrolled) use of genAI, which given that this plugin has a high security impact is just a no-go.
2. The plugin's name is misleading. This is not about adding any additional factors to OctoPrint's existing password based authentication workflow (e.g. through the existing MFA plugin interface), but rather fully replaces the default login, which is not supposed to be replaced, making the whole replacement approach hard to maintain and extremely easy to break. That's a no-go.
3. The plugin makes quite a number of assumptions about the runtime environment and will break on anything that doesn't match it. Possibly, it will also break the non-matching environment in the progress. That's a no-go.
4. Finally, this is not something that should be implemented through a plugin in the first place. There are [several IDM projects out there](https://github.com/awesome-foss/awesome-sysadmin#identity-management) that allow to take care of authentication through passkeys, existing SSO setups and whatever else, and OctoPrint is able to work with them by configuring the remote user related things in its `config.yaml` file.

You are welcome to use this plugin for yourself, share it etc. No one is keeping you from doing this. You do you! 

But for the sake of the security and safety of OctoPrint's user community, it will not be allowed on the official plugin repository.

PR #1430, opened 2026-04-30
bedcooldown: Migrate from github.com to codeberg.org

rfinnie:
Sorry for the rapid-fire PRs (two within a week!); it was actually the previous PR which kickstarted me looking into a Codeberg migration.

### DIFF
--- a/_plugins/bedcooldown.md
+++ b/_plugins/bedcooldown.md
@@ -10,9 +10,9 @@ license: MPL-2.0
 
 date: 2021-09-26
 
-homepage: https://github.com/rfinnie/OctoPrint-BedCooldown
-source: https://github.com/rfinnie/OctoPrint-BedCooldown
-archive: https://github.com/rfinnie/OctoPrint-BedCooldown/archive/release.zip
+homepage: https://codeberg.org/rfinnie/OctoPrint-BedCooldown
+source: https://codeberg.org/rfinnie/OctoPrint-BedCooldown
+archive: https://codeberg.org/rfinnie/OctoPrint-BedCooldown/archive/release.zip
 
 follow_dependency_links: false
 


jneilliii (approved):

jneilliii:
I merged this without testing the archive url, but the redirect on codeberg causes plugin manager to fail. 

image

jneilliii:
Using the "can't wait" link seems to work though.
```
https://codeberg.org/rfinnie/OctoPrint-BedCooldown/archive/release.zip?trans-rights=human-rights
```

I wonder if your software update hook may also have the same issue with version updates...

jneilliii:
Not sure if it's because of OctoPrint 2.0.0 or not, but after install from the above URL, I get prompted to upgrade and after upgrade end up getting a blank list of plugins in software update settings. I've reverted this PR until we've had a better chance at testing this process completely.

rfinnie:
Thanks, some weirdness there. I briefly published 0.8.1 with the switchover, but have nuked the releases and switched the release branches back to 0.8.0. You and I were (hopefully) the only people who actually tried to update during that window.

curl and wget against that release.zip download directly as expected, but codeberg seems to send something else (probably a redirect as you said) when downloaded through `pip`:

```
$ /home/ryan/venv3/bin/python -m pip     --disable-pip-version-check     install https://codeberg.org/rfinnie/OctoPrint-B
edCooldown/archive/release.zip     --no-cache-dir     --use-pep517     --no-build-isolation
Looking in indexes: https://pypi.org/simple, https://www.piwheels.org/simple
Collecting https://codeberg.org/rfinnie/OctoPrint-BedCooldown/archive/release.zip                                                                                           
  Downloading https://codeberg.org/rfinnie/OctoPrint-BedCooldown/archive/release.zip (603 bytes)
ERROR: Exception:              
```

As I said, everything causing an update should be reverted now, and I'll keep looking once I get my bearings straight. Thanks for the catch.

rfinnie:
Discussion started here: https://codeberg.org/Codeberg/Community/issues/2563

In doing so, I discovered the plugin manager UI mentions "Expected http, https or any of the [supported VCS URLs](https://pip.pypa.io/en/stable/reference/pip_install/#vcs-support).", except that functionality appears to be just plain broken. e.g. `pluginmanager/__init__.py` contains `URL_SCHEMES` which is not used anywhere; it appears to just want to download a zip and pass it to `pip install`.

Once I get some coffee, I'll verify further and open a new issue. But my preference is convincing Codeberg to not block these zips from pip.

rfinnie:
Just a follow-up here. The Codeberg issue ended up being a relatively new (at the time) anti-crawler protection they added, which they ended up reverting after more issues were brought up. Downloading zips from Codeberg should be good again.

PR #1435, opened 2026-05-06
Add octogoat plugin

ksmith1489:
- [x] You have read the ["Registering a new Plugin"](https://plugins.octoprint.org/help/registering/) guide.
- [x] You want to and are able to maintain the plugin you are registering, long-term.
- [x] You understand why the plugin you are registering works.
- [x] You have read and acknowledge the [Code of Conduct](https://octoprint.org/conduct/).

#### What is the name of your plugin?

OctoGoat

#### What does your plugin do?

OctoGoat helps users recover failed 3D prints when the partially completed print is still attached to the bed.

The plugin generates reconstructed resume G-code locally inside OctoPrint, using the original G-code file, the measured print height, and the user’s slicer layer height. It also provides a guided alignment flow so the user can safely align the printer to the real-world print position before choosing to resume.

It is designed to keep the user in control of printer movement and final resume confirmation. It does not force automatic Z homing into an existing print.

OctoGoat also includes optional Klipper/Moonraker support through a user-provided local Moonraker address.

#### Where can we find the source code of your plugin?

https://github.com/ksmith1489/octogoat-plugin

#### Was any kind of genAI (ChatGPT, Copilot etc) involved in creating this plugin?

Yes. ChatGPT, GitHub Copilot, and Codex were used as development assistants while building and refining the plugin.

The plugin was not simply generated blindly and published without review. I have been actively involved in the design, testing, debugging, and iteration of the plugin, including testing the OctoPrint UI, local resume G-code generation, license validation behavior, and Klipper/Moonraker command flow on actual printer setups.

I understand the main architecture and behavior of the plugin: the OctoPrint plugin UI gathers the file and recovery inputs, the local resume engine reconstructs the resume G-code, unsafe commands are removed before resuming, alignment commands are user-triggered, and the external service is used for subscription/license validation rather than for uploading or processing the print file.

I also have prior coding experience from building trading bots and indicators, including Pine Script and MetaTrader 4 related work, and I am comfortable maintaining and debugging this plugin going forward.

#### Is your plugin commercial in nature?

Yes.

OctoGoat is a commercial plugin with subscription-based licensing. Installing the plugin is free, but an active paid subscription is required before the plugin can generate, download, upload, or execute resume output.  There is only one price and plan for $29 per year.

License, terms, and privacy information are available at:

https://app.lazarus3dprint.com

#### Does your plugin rely on some cloud services?

Partially.

[...]

PR #1437, opened 2026-05-10
Add lazarus plugin

ksmith1489:

Lazarus

Lazarus helps users recover failed 3D prints when the partially completed print is still attached to the bed.

The plugin generates resume G-code locally inside OctoPrint from the original G-code file, the measured print height, and the detected layer height of that file. Yes it even considers and calculates if there is a different initial layer height.  Also works in spiral vase mode.  Lazarus provides a guided alignment workflow so the user can safely align the printer to the real-world print position before choosing to resume.  It takes a relatively complex series of problems and simplifies it so that even a novice to 3d printing can resume a failed print easily, safely, accurately, and surprisingly fast.  Along with generating a safe resume Gcode that carefully grabs the needed information and blocks anything that could cause a mishap it also calculates the exact layer of failure, a true alignment point which is the lowest x value or the highest x value depending on which side the user entered to be the lowest and paired with y at calculated z to provide an easy real world physical reference point at which a user can toggle to paper test perfect.  It displays a preview of the resume gcode for the user's inspection along with a variety of different methods for starting the alignment process based on a user's firmware and skill level.   An advanced Marlin user may choose to attest to their printer's current coordinate state while an advanced klipper user may want to begin with the coordinates that their printer parks at after a cancel_print event.  The beginner may want to stick with the safe resume homing which homes x and y and sets a temporary safe z without moving z until a true z has been established through user confirmed alignment.  It is way easier done than said.  Lazarus has been carefully developed to attend to most all of the nuances of different printers, gcode flavors and firmware and it does it very well keeping it simple for the user. What sets Lazarus apart is the ability to establish coordinate state starting with total loss of internal coordinates, grayed out and motors locked and align to the partial print and the reconstructed gcode just as accurately and even more so than G28 homing with no unsafe motion.  Actually only 2 motion commands in the whole process and one of them is a z hop after alignment so the user can begin to heat the nozzle without melting the partial print and the other does not move z at all.   
Lazarus not only makes resuming a failed print worth doing,  it actually makes it fun and very satisfying.  

Optional Klipper / Moonraker support is also included through a user-provided local Moonraker address.

https://github.com/ksmith1489/lazarus-plugin

Yes ChatGPT/ recently Codex were used as development assistants during development and cleanup. I understand the plugin’s architecture and behavior and I am maintaining it directly.  They just help with the code. The idea, the methods and systems, the processes, the flow were all human.  And I have had some experience coding trading bots and indicators on metatrader and trading view some time ago and have been working on this project nonstop over full time for several months and it works and I will keep it that way.

Yes.

Lazarus is commercial software. Installation is free, but resume generation and execution require an active subscription.

Pricing, activation, license, terms, and privacy information are available here:

https://app.lazarus3dprint.com/activate
https://app.lazarus3dprint.com/license
https://app.lazarus3dprint.com/terms
https://app.lazarus3dprint.com/privacy

Partially.

Resume G-code generation runs locally inside the OctoPrint plugin. The original G-code file is not uploaded to the Lazarus service for resume generation.

The plugin does use https://app.lazarus3dprint.com for activation and subscription validation.

The plugin is marked with the `cloud` attribute and includes a privacy policy link:
https://app.lazarus3dprint.com/privacy

If the license validation service is unavailable, Lazarus fails closed for resume generation and execution without causing OctoPrint itself to malfunction.

#### Further notes

This is a fresh registration under `Lazarus` after the previous `OctoGoat` submission was closed.

This resubmission addresses the review items raised on `#1435`:

- the plugin has been renamed from `OctoGoat` to `Lazarus`;
- packaging has been migrated to `pyproject.toml`, and the repository layout has been aligned with the current OctoPrint cookiecutter-style scaffold while avoiding the legacy `setup.py` packaging path;
- the promotional / watermarked hero assets were removed and the UI copy was simplified;
- unused settings, no-op startup code, and unreferenced API / frontend paths were removed;
- checkout remains outside OctoPrint, and the plugin keeps server-side permission checks, server-side license enforcement, the privacy policy link, and the software update hook;
- the source repo now includes a regression test for the resume engine.

If there are any remaining repository-specific issues with this Lazarus submission, I would appreciate a concrete list and I will address them directly.


foosel:
As has been pointed out to you on your former PR, the reason for not allowing the registration of your plugin on the official OctoPrint Plugin Repository to go forward was not specific isolated issues with your implementation, but rather general quality concerns given the existing code structure, implementation issues, and implementation of feedback, which strongly hinted at your LLMs playing an unhealthily large role in creation of the code here.

To quote the registration rules:

> This also means that you should be capable of fixing issues in your plugin even if your favourite genAI is down

Your statement in your (now deleted) comment:

> And with the progress of LLMs .... maintaining the plugin isn't going to be difficult.

gives the strong impression that this is not the case here, and together with the observed problems with the plugin's structure, the still not solved issues, some newly introduced issues and frankly also the fact that you do not seem to or want to grasp the actual systemic critique you received back, means that no, regardless of how you name it or how more often you run your LLMs against its codebase, this plugin will not be allowed on the plugin repository. 

As already pointed out, you are welcome to use this plugin for yourself, share it etc. No one is keeping you from doing this. You do you!

But for the sake of the security and safety of OctoPrint's user community, it will not be allowed on the official plugin repository. **This is our final verdict**. Do not keep opening PRs about this!

PR #1439, opened 2026-05-15
Added plugin git_backup

MauroDruwel:

Git Backup Plugin

Automatically pushes OctoPrint backups to a remote Git repository every time a backup is created. Once configured with a repo URL, every new backup is committed and pushed, keeping a full version-controlled history of your printer's configuration.

It also includes a live status panel in the OctoPrint settings UI that checks whether git and the GitHub CLI are installed and authenticated, with one-click install and setup buttons for apt-based systems (Raspberry Pi OS, Ubuntu, Debian).

https://github.com/MauroDruwel/OctoPrint-Git-Backup

Yes, GitHub Copilot was used as a coding assistant during development. I understand how the plugin works, actively directed all design decisions, and I am able to maintain it independently.

No.

No. The plugin runs `git push` to a user-configured remote repository. No data is sent to any service operated by the plugin author. Authentication is handled entirely at the OS level by the user (GitHub CLI or SSH key).

#### Further notes

- Python 3 only (`>=3,<4`)
- Requires OctoPrint >= 1.6.0 (backup plugin helpers)
- MIT licensed
- No credentials stored in the plugin


_plugins/git_backup.md
```diff
@@ -0,0 +1,74 @@
...
+
+featuredimage: /assets/img/plugins/git_backup/settings.png
+
+compatibility:
+  octoprint:
+  - 1.6.0
+
+  os:
+  - linux
+  - windows
```
jneilliii, line 41: I don't think this plugin is windows compatible due to the use of apt install commands. The other functions may work/exist in Windows though possibly.

jneilliii (commented):
Just a quick glance over the plugin code there are a couple of things that need some work. 

Threads should set the deamon property to True, not False as is the case [here](https://github.com/MauroDruwel/OctoPrint-Git-Backup/blob/de4d6807b45b6f6e79ee5cc9ac81285ee2b5ed2c/octoprint_git_backup/__init__.py#L78). 

Your plugin doesn't have a release yet matching the version in pyproject.toml. 

Is the plugin windows compatible (see review comment).

I've attached the scan results from [octoscanner](https://github.com/jacopotediosi/octoscanner) that points our a couple of security concerns that need to be addressed. 

[octoscanner_results.txt](https://github.com/user-attachments/files/27826875/octoscanner_results.txt)

When I have a little more time I'll go deeper through the code as there are some programming techniques (js client function injection into OctoPrint client) and command line functions (git) that I'm not familiar with.

jneilliii:
After a little further reading on the docs site, I think the proper method of injecting your plugin's code into the OctoPrintClient js library is to use [registerPluginComponent](https://docs.octoprint.org/en/main/jsclientlib/base.html#OctoPrintClient.registerPluginComponent) as mentioned in [OctoPrintClient.plugins](https://docs.octoprint.org/en/main/jsclientlib/base.html#OctoPrintClient.plugins). Honestly, the only reason you would want to do that I think is if you wanted your plugin's js code to be runnable by other plugins, which based on the usecase I think is not really needed. My recommendation would be to move all those functions into your viewmodel code like 
```js
self.startAuthLogin = function() {
// do stuff
}
```
for example and call those functions directly like `self.startAuthLogin()`

This just seems like another example of AI not understanding the context of OctoPrint's core code and best practices for plugins. Yes, it may work, but seems difficult to maintain and not quite up to the quality standards of what's expected for registration in the official plugin repository at its current state.

_plugins/git_backup.md
jacopotediosi: For the same reason I believe it is not compatible with macOS and freebsd either

jacopotediosi (changes requested):
First of all, I'd like to thank @MauroDruwel for his contribution. I really appreciate it!

Below are two additional issues I noticed:
- The first two lines of your `__init__.py` (`# coding=utf-8` and `from __future__ import absolute_import`) are leftovers and can be removed since your plugin declares itself as python3-only.
- The Python version declared in `__init__.py` (`__plugin_pythoncompat__ = ">=3,<4"`) is inconsistent with the version declared in this PR's manifest (`python: ">=3,<4"`) and with the one declared in `pyproject.toml` (`requires-python = ">=3.7, <4"`). They should all be aligned.

Below are some food for thought that came to mind while reading the plugin's code:
- I have some concerns about the security of backing up OctoPrint configurations to GitHub, and to the cloud in general, since they typically contain very sensitive information. I would suggest considering more alarming and transparent warnings/documentation for the user that require interaction to be dismissed (e.g., ticking a checkbox in a [setup wizard](https://docs.octoprint.org/en/dev/plugins/mixins.html#wizardplugin)), to ensure the consequences are really understood, and perhaps a check that prevents the push altogether if the repository is public.
- I'm not convinced about the use of `apt` and `sudo` to install dependencies (e.g., `git` and `gh`), because this constrains compatibility to specific Linux distros (not all of them have `apt` and `sudo`). It also won't work in environments where the user has configured a password for `sudo` (I seem to recall this is the default on OctoPi). I think there's a better approach, for example installing the dependencies in user space, so that `sudo` is not needed, and downloading the executables directly from their source instead of using `apt` or other package managers. Maybe there are cross-platform Python libraries to interface with `git`/`gh` or to achieve the same thing? Maybe you can look at `dulwich` or `pygit2`.

If needed, feel free to ask, I can help with these!

MauroDruwel:
> First of all, I'd like to thank @MauroDruwel for his contribution. I really appreciate it!
> 
> Below are two additional issues I noticed:
> - The first two lines of your `__init__.py` (`# coding=utf-8` and `from __future__ import absolute_import`) are leftovers and can be removed since your plugin declares itself as python3-only.
> - The Python version declared in `__init__.py` (`__plugin_pythoncompat__ = ">=3,<4"`) is inconsistent with the version declared in this PR's manifest (`python: ">=3,<4"`) and with the one declared in `pyproject.toml` (`requires-python = ">=3.7, <4"`). They should all be aligned.
> 
> Below are some food for thought that came to mind while reading the plugin's code:
> - I have some concerns about the security of backing up OctoPrint configurations to GitHub, and to the cloud in general, since they typically contain very sensitive information. I would suggest considering more alarming and transparent warnings/documentation for the user that require interaction to be dismissed (e.g., ticking a checkbox in a [setup wizard](https://docs.octoprint.org/en/dev/plugins/mixins.html#wizardplugin)), to ensure the consequences are really understood, and perhaps a check that prevents the push altogether if the repository is public.
> - I'm not convinced about the use of `apt` and `sudo` to install dependencies (e.g., `git` and `gh`), because this constrains compatibility to specific Linux distros (not all of them have `apt` and `sudo`). It also won't work in environments where the user has configured a password for `sudo` (I seem to recall this is the default on OctoPi). I think there's a better approach, for example installing the dependencies in user space, so that `sudo` is not needed, and downloading the executables directly from their source instead of using `apt` or other package managers. Maybe there are cross-platform Python libraries to interface with `git`/`gh` or to achieve the same thing? Maybe you can look at `dulwich` or `pygit2`.
> 
> If needed, feel free to ask, I can help with these!

Hi, thanks for reviewing!! I'll look into the code reviews later, but you're right a python user library is probably better. The reason I built it like this was because I first just used git commands, then I wanted to test this on the octopi docker and not all dependencies were installed, that's why I added the automatic installation. Also it does work on al OS'es even Windows, if the user themselves install git etc, I think I documented this, the auto install function was just a feature if a user finds it hard to configure git ;) But I'll look into those python libs

PR #1442, opened 2026-05-20
add ai-developed attribute

jneilliii:
Initial changes for identifying plugins where AI was involved in the development. This PR will simply add a badge similar to Cloud or Commercial is currently.


foosel:
Love it! I'll add that also to the built in repo browser! Love the choice of icon and color!

PR #1458, opened 2026-07-05
Added plugin maintenance_reminder

GarronAnderson:
Added a plugin which provides a simple maintenance reminder function. It counts prints and keeps track of what maintenance is due.

OctoPrint-Maintenance Reminder

It counts your prints and reminds you to perform periodic maintenance before each print if it is due. I wrote this because I wanted a plugin to remind me to wash my PEI sheet and one did not exist.

https://github.com/GarronAnderson/OctoPrint-Maintenance-Reminder

ChatGPT was used for some help with JavaScript (I'm a Python guy and the syntax is weird). No agents such as Claude Code were used. A lot of inspiration was me (the human) looking at code for octoprint-spoolman and Top Temp.

No, not at all.

No. I may add support for notifications through ntfy.sh in the future.


_plugins/maintenance_reminder.md
```diff
@@ -0,0 +1,57 @@
...
+source: https://github.com/GarronAnderson/OctoPrint-Maintenance-Reminder
+archive: https://github.com/GarronAnderson/OctoPrint-Maintenance-Reminder/archive/main.zip
+
+# TODO
+tags:
+    - maintenance
+    - reminders
+    - alerts
+    - monitoring
+
```
jneilliii, line 23: missing compatibility attribute, where python should be indented inside. this also is where you would have octoprint version and OS compatibility information. 

```yaml
compatibility:
  octoprint:
  - 1.9.0
  os:
  - linux
  - windows
  - mac
  python: ">=3,<4"
```

Not sure if it matters too, but looks like your yaml formatting is 4 space based and not 2 space based like most other plugins. Edited attribute above is only using 2 spaces, adjust as necessary until the build checks are happy.

jneilliii (changes requested):
There actually is an unlisted plugin that has been around for quite some time <https://github.com/OllisGit/OctoPrint-MaintenanceManager>. Looks like the PR validations are failing though, which most likely is a yaml formatting issue. 

[octoscanner ](https://github.com/jacopotediosi/octoscanner)also flagged these items...

```
╭──────────────────────────────────────────────────── OctoScanner ─────────────────────────────────────────────────────╮
│                                                                                                                      │
│ ╭─ Plugin C:\Users\Jim Neill\Documents\3D Prints\PrinterUprades\OrdBotHadron\OctoprintRelated\octoscanner\plugins_─╮ │
│ │ ╭───────────────────────────────────────── Summary (2 total findings) ─────────────────────────────────────────╮ │ │
│ │ │ 2 security                                                                                                   │ │ │
│ │ ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────╯ │ │
│ │ ╭─────────────────────────────────────────── Security (2 findings) ────────────────────────────────────────────╮ │ │
│ │ │ SEC-0001                                                                                                     │ │ │
│ │ │ Plugin implements SimpleApiPlugin but does not override is_api_protected(). Until OctoPrint 2.1.0, the       │ │ │
│ │ │ default implementation returns False, allowing unauthenticated access to SimpleApi endpoints.                │ │ │
│ │ │ File: octoprint_maintenance_reminder\__init__.py:10                                                          │ │ │
│ │ │        9 |                                                                                                   │ │ │
│ │ │   >   10 | class MaintenanceReminderPlugin(                                                                  │ │ │
│ │ │       11 |     octoprint.plugin.SettingsPlugin,                                                              │ │ │
│ │ │ Suggestion: Override is_api_protected() to explicitly return True. Note that this only controls              │ │ │
│ │ │ authentication; you should also verify permissions explicitly in each API handler (e.g.                      │ │ │
│ │ │ Permissions.PLUGIN_MYPLUGIN_ACCESS).                                                                         │ │ │
│ │ │                                                                                                              │ │ │
│ │ │ SEC-0007                                                                                                     │ │ │
│ │ │ Plugin implements TemplatePlugin but does not override is_template_autoescaped(). Until OctoPrint 2.1.0, the │ │ │
│ │ │ default implementation returns False, exposing the templates to XSS via unescaped variables.                 │ │ │
│ │ │ File: octoprint_maintenance_reminder\__init__.py:10                                                          │ │ │
│ │ │        9 |                                                                                                   │ │ │
│ │ │   >   10 | class MaintenanceReminderPlugin(                                                                  │ │ │
│ │ │       11 |     octoprint.plugin.SettingsPlugin,                                                              │ │ │
│ │ │ Suggestion: Override is_template_autoescaped() to return True and use the safe Jinja filter only for         │ │ │
│ │ │ variables that legitimately need to render HTML. See https://faq.octoprint.org/plugin-autoescape.            │ │ │
│ │ ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────╯ │ │
│ ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯ │
│                                                                                                                      │
│ ╭──────────────────────────── Rule Statistics (2 rules matched 2 times on 1/1 plugins) ────────────────────────────╮ │
│ │ Matches  Rule ID   Message                                                                                       │ │
│ │       1  SEC-0001  Plugin implements SimpleApiPlugin but does not override is_api_protected(). Until OctoPrint   │ │
│ │                    2.1.0, the default implementation returns False, allowing unauthenticated access to           │ │
│ │                    SimpleApi endpoints.                                                                          │ │
│ │       1  SEC-0007  Plugin implements TemplatePlugin but does not override is_template_autoescaped(). Until       │ │
│ │                    OctoPrint 2.1.0, the default implementation returns False, exposing the templates to XSS      │ │
│ │                    via unescaped variables.                                                                      │ │
│ ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯ │
│                                                                                                                      │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

Once these issues are addressed I can do a full review.

jneilliii:
I did just take a quick peak at the plugin code, and their is definitely better approaches over monkeypatching the UI in your js. I'd recommend looking at [onBeforePrintStart viewmodel callback](https://docs.octoprint.org/en/main/plugins/viewmodels.html#callbacks). This will allow you to achieve what you are trying to do without interfering with the core OctoPrint code/functions. An example of using this callback can be found in our Bambu plugin [here](https://github.com/OctoPrint/OctoPrint-BambuConnector/blob/406072480c6a5f2a2cdaaeeec6bc3c8f8452a9a1/octoprint_bambu_connector/static/js/bambu_connector.js#L23-L49).

GarronAnderson:
Thank you so much for the information @jneilliii ! I did not know about the viewmodel callback and was working off of what octoprint-spoolman had done.

jneilliii:
Another thing that first time plugin developers usually have trouble understanding is knockout bindings. For example, your progress bars could utilize the progress data-binding the way that OctoPrint uses it internally. Example is the gcode viewer that has this type of binding [here](https://github.com/OctoPrint/OctoPrint/blob/dev/src/octoprint/plugins/gcodeviewer/templates/gcodeviewer_tab.jinja2#L26) which on the js side is defined and updated via observables [here](https://github.com/OctoPrint/OctoPrint/blob/10bf10076ceffab137b96dbab19e8e81da093b7c/src/octoprint/plugins/gcodeviewer/static/js/gcodeviewer.js#L721). 

Another example, is instead of adding a click callback via jQuery in your plugin [here](https://github.com/GarronAnderson/OctoPrint-Maintenance-Reminder/blob/df15551f905f7f1c83fb423960329bbb4aeed017/octoprint_maintenance_reminder/static/js/maintenance_reminder.js#L188-L193) you could use the knockout data-bind 'click' in your jinja2 file. Are you doing it this way just because your  current implementation of the popover is being built on the fly via js for your navbar button? I would recommend looking at other examples that use native dropdown for this like [here](https://github.com/jneilliii/OctoPrint-MQTTPublish/blob/743d644b04229aea18330b536808091d0e4b8b63/octoprint_mqttpublish/templates/mqttpublish_navbar.jinja2#L1-L10) in my MQTTPublish plugin. This would also eliminate the need for assigning your viewmodel to the window context [here](https://github.com/GarronAnderson/OctoPrint-Maintenance-Reminder/blob/df15551f905f7f1c83fb423960329bbb4aeed017/octoprint_maintenance_reminder/static/js/maintenance_reminder.js#L10) I think. 

I'm typically online in the OctoPrint Discord server, so if you want to jump in and discuss these options further or need more clarification I'm happy to help in the #dev-plugins channel.

_plugins/maintenance_reminder.md
```diff
@@ -0,0 +1,60 @@
...
+description: Receive maintenance notifications when you complete a certain number of prints
+authors:
+  - Garron Anderson
+license: AGPL-3.0-or-later
+
+date: 2026-07-03
+
+homepage: https://github.com/GarronAnderson/OctoPrint-Maintenance-Reminder
+source: https://github.com/GarronAnderson/OctoPrint-Maintenance-Reminder
+archive: https://github.com/GarronAnderson/OctoPrint-Maintenance-Reminder/archive/main.zip
```
jacopotediosi: Your repository default branch is `master`, not `main`, so this link won't work. It should be `https://github.com/GarronAnderson/OctoPrint-Maintenance-Reminder/archive/master.zip`. Please remember to fix also the link in your plugin README.md, which has the same issue.

jacopotediosi (changes requested):
Aside from my comment below (the non working links) and @jneilliii's remarks (using `onBeforePrintStart` instead of monkeypatching the UI, and using knockout bindings correctly), the rest looks fine to me.

GarronAnderson:
Sorry it took me so long! I have rewritten the dropdown per jacopotediosi on the Discord. It now uses the dropdown class and Knockout bindings. The dialog also uses onBeforePrintStart. I have updated the branch name. Thank you!

_plugins/maintenance_reminder.md
```diff
@@ -0,0 +1,60 @@
...
+tags:
+  - maintenance
+  - reminders
+  - alerts
+  - monitoring
+
+compatibility:
+  octoprint:
+  - 1.9.0
+  python: ">=3,<4"
```
jacopotediosi: Should be >= 3.7, as indicated [here](https://github.com/GarronAnderson/OctoPrint-Maintenance-Reminder/blob/214223d88290048affd17cf061c7e6f55596dc60/pyproject.toml#L17). Please align also [this line](https://github.com/GarronAnderson/OctoPrint-Maintenance-Reminder/blob/214223d88290048affd17cf061c7e6f55596dc60/octoprint_maintenance_reminder/__init__.py#L125) to that version.

jacopotediosi (changes requested):
Well done! Your implementation of `onBeforePrintStart` looks good.

A potential bug tho: [here](https://github.com/GarronAnderson/OctoPrint-Maintenance-Reminder/blob/214223d88290048affd17cf061c7e6f55596dc60/octoprint_maintenance_reminder/__init__.py#L96) you call `self._settings.set` without `self._settings.save()`, so that value is likely not actually persisted after an OctoPrint reboot.

Beside that, and the below pending comment in this PR, the rest now looks good to me.

Please also remember to publish a Release, as we cannot merge this PR without that.

GarronAnderson:
Thank you for all the help! I have updated the Python version info to 3.7 or above. The settings should not save after reminder increment. 

I have published a GitHub release!

jacopotediosi (approved):
LGTM

GarronAnderson:
Thank you so much!

PR #1463, opened 2026-07-17
Add Alexa OctoPrint

RICLAMER:
- [x] You have read the ["Registering a new Plugin"](https://plugins.octoprint.org/help/registering/) guide.
- [x] You want to and are able to maintain the plugin you are registering, long-term.
- [x] You understand why the plugin you are registering works.
- [x] You have read and acknowledge the [Code of Conduct](https://octoprint.org/conduct/).

#### What is the name of your plugin?

Alexa OctoPrint

#### What does your plugin do?

It exposes selected OctoPrint and 3D printer actions as Philips Hue compatible devices that Alexa can discover and control directly over the local network. Supported actions include print control, motion, heating presets, emergency shutdown, configured files, and optional OctoPrint-Enclosure power/light outputs.

#### Where can we find the source code of your plugin?

https://github.com/RICLAMER/AlexaOctoPrint

The repository contains the public documentation and current source distribution package under plugin/.

#### Was any kind of genAI (ChatGPT, Copilot etc) involved in creating this plugin?

Yes. OpenAI Codex assisted with implementation, tests, debugging, and English documentation. The maintainer defined the requirements and performed iterative testing against a physical Raspberry Pi, OctoPrint installation, Alexa devices, printer, and Enclosure outputs. The registration metadata includes the required ai-developed attribute.

#### Is your plugin commercial in nature?

No.

#### Does your plugin rely on some cloud services?

No cloud backend or Alexa skill is used for discovery or control. Runtime device communication stays on the LAN. GitHub is used only for package distribution and Software Update version checks.

#### Further notes

Version 0.1.7 was installed from the public archive and validated on OctoPrint 1.6.1 with Python 3.7.3. OctoPrint's forced Software Update check reports the installed and available versions as 0.1.7 and up to date.

Alexa's local Hue integration requires root Hue routes on port 80. On OctoPi, the documented setup keeps HAProxy as the only port 80 listener and selectively forwards the required paths to OctoPrint. The plugin does not modify system services automatically.

PR #1464, opened 2026-07-17
Add Alexa OctoPrint (revised after #1463)

RICLAMER:

Alexa OctoPrint

It simulates local smart devices for selected OctoPrint and 3D printer actions. Alexa discovers and controls those actions on the LAN without an external backend, Alexa skill, or integration account. It includes Portuguese, English, and Spanish device names, configurable printer actions, and optional OctoPrint-Enclosure power/light outputs.

https://github.com/RICLAMER/AlexaOctoPrint/tree/0.2.0/Source

Tests: https://github.com/RICLAMER/AlexaOctoPrint/tree/0.2.0/tests

Release: https://github.com/RICLAMER/AlexaOctoPrint/releases/tag/0.2.0

Yes. OpenAI Codex assisted with implementation, tests, debugging, and English documentation. The maintainer defined the requirements and performed iterative testing against a physical Raspberry Pi, OctoPrint installation, Alexa devices, printer, HAProxy, and Enclosure outputs. The registration metadata includes the required `ai-developed` attribute.

No.

No cloud backend, Alexa skill, account, telemetry service, or relay is used for discovery or control. Runtime device communication stays on the LAN. GitHub is used only for source, releases, installation, and OctoPrint Software Update checks.

#### Changes made after the review in #1463

- Published the complete source and test suite instead of only a distribution ZIP.
- Replaced the previous license with `AGPL-3.0-or-later` and included the full license text.
- Replaced the custom update metadata with OctoPrint's `github_release` configuration and tagged source archives, following the `jneilliii/OctoPrint-BLTouch` pattern.
- Removed the shared static Hue-compatible username. A random 40-character username is generated per installation, validated on light routes, and redacted from logs.
- Restricted HAProxy ACLs by HTTP method and exact local smart-device paths; normal OctoPrint API routes such as `/api/job` and `/api/files` are excluded.
- Added explicit, optional Easy HAProxy Setup with listener detection, backup, validation, rollback, removal, and preservation of unrelated routes. The helper never kills a port owner.
- Simplified the settings UI and removed the temporary test device and mutable network/SSDP controls.
- Added optional Enclosure output dropdowns with controlled unavailable handling when the dependency is absent or disabled.
- Added Espalexa attribution as the protocol research/reference used for local interoperability.

#### Verification

Version 0.2.0 passed 30 automated tests and the gallery front-matter validator with no warnings. The tagged GitHub archive installs as `OctoPrint-AlexaOctoPrint 0.2.0`. It was installed on OctoPrint running Python 3.7.3 and validated on a Raspberry Pi with local Alexa discovery, HAProxy route isolation, and real Enclosure `POWER`/`LIGHT` switching.

This PR replaces the closed #1463.

### DIFF
--- a/_plugins/alexaoctoprint.md
+++ b/_plugins/alexaoctoprint.md
@@ -12,7 +12,7 @@ date: 2026-07-17
 
 homepage: https://github.com/RICLAMER/AlexaOctoPrint
 source: https://github.com/RICLAMER/AlexaOctoPrint
-archive: https://github.com/RICLAMER/AlexaOctoPrint/archive/0.2.0.zip
+archive: https://github.com/RICLAMER/AlexaOctoPrint/releases/latest/download/OctoPrint-AlexaOctoPrint.zip
 
 tags:
 - alexa
--- a/_plugins/alexaoctoprint.md
+++ b/_plugins/alexaoctoprint.md
@@ -1,0 +1,55 @@
+---
+layout: plugin
+
+id: alexaoctoprint
+title: Alexa OctoPrint
+description: Control selected OctoPrint and 3D printer actions through Alexa on the local network without an external backend.
+authors:
+- RICLAMER
+license: AGPL-3.0-or-later
+
+date: 2026-07-17
+
+homepage: https://github.com/RICLAMER/AlexaOctoPrint
+source: https://github.com/RICLAMER/AlexaOctoPrint
+archive: https://github.com/RICLAMER/AlexaOctoPrint/archive/0.2.0.zip
+
+tags:
+- alexa
+- automation
+- control
+- raspberrypi
+
+compatibility:
+  octoprint:
+  - ">=1.6.0"
+  os:
+  - linux
+  python: ">=3.7,<4"
+
+attributes:
+- ai-developed
+
+---
+
+Alexa OctoPrint simulates local smart devices for enabled printer actions. No
+external integration account, Alexa skill, cloud backend, telemetry service, or
+relay is required.
+
+Actions include pause, resume, guarded cancellation, homing, leveling, Z and
+extruder movement, configurable bed and nozzle temperatures, motors, selected
+files, emergency stop, and optional OctoPrint-Enclosure power and light outputs.
+Portuguese, English, and Spanish device names are included.
+
+Local discovery requires restricted root routes on TCP port 80. Existing
+HAProxy or another reverse proxy can provide them. The repository includes an
+explicit Easy Setup that inspects the current listener, preserves unrelated
+routes, validates HAProxy before restart, and saves rollback state. The plugin
+does not silently modify system services.
+
+Each installation generates its own local 40-character username. Proxy ACLs
+match only the description, username creation, and corresponding light paths;
+normal OctoPrint API routes are excluded.
+
+Updates use OctoPrint's `github_release` check and install the archive for the
+matching GitHub release tag.


_plugins/alexaoctoprint.md
```diff
@@ -0,0 +1,55 @@
...
+description: Control selected OctoPrint and 3D printer actions through Alexa on the local network without an external backend.
+authors:
+- RICLAMER
+license: AGPL-3.0-or-later
+
+date: 2026-07-17
+
+homepage: https://github.com/RICLAMER/AlexaOctoPrint
+source: https://github.com/RICLAMER/AlexaOctoPrint
+archive: https://github.com/RICLAMER/AlexaOctoPrint/archive/0.2.0.zip
```
jneilliii: Defining your archive this way will require updating the plugin repository on every new version. Rather than doing this we recommend that you link to either the latest tag or direct zip download of the repository.

latest tag with uploaded asset (assumes that OctoPrint-AlexaOctoPrint.zip will be included on every release)
```
https://github.com/RICLAMER/AlexaOctoPrint/releases/latest/download/OctoPrint-AlexaOctoPrint.zip
```

direct zip download
```
https://github.com/RICLAMER/AlexaOctoPrint/archive/main.zip
```

jneilliii (changes requested):
Thank you, this is a little better. To continue the review I'll need to know if you plan on using the latest tag or direct zip download option mentioned in the code comment.

_plugins/alexaoctoprint.md
RICLAMER: Thank you. I chose the latest stable GitHub Release asset and updated the `archive` entry to:

```
https://github.com/RICLAMER/AlexaOctoPrint/releases/latest/download/OctoPrint-AlexaOctoPrint.zip
```

This keeps the Plugin Repository URL permanent and installs the latest tested stable release instead of tracking untagged changes from `main`. It also matches the plugin's GitHub Release update strategy.

Release `0.2.0` already includes that exact asset name, the permanent URL has been download-tested, and every future stable Release will include `OctoPrint-AlexaOctoPrint.zip`. The project README now documents this release requirement as well.

jacopotediosi:
## Cookiecutter template

This plugin was written from scratch by some LLM, so it does not follow the structure of the [OctoPrint plugin cookiecutter template](https://github.com/OctoPrint/cookiecutter-octoprint-plugin). That template exists not only to make plugin reviews easier, but also to ensure that installation succeeds in all scenarios supported by OctoPrint.

A brief five-minute look was enough to spot the following issues:

- In `setup.py`, unlike in the template, the license string for backward compatibility is not defined, so this plugin might fail to install in environments with `setuptools<77`.
- In `pyproject.toml`, the `requires` entries differ from those in the template, and the author's email address is missing.
- There is an empty `requirements.txt` containing only the comment `# The plugin uses OctoPrint and Python standard-library modules only`. Usually, when `pyproject.toml` is used, there is no `requirements.txt` file at all.
- The plugin sources live in a `Source/octoprint_alexaoctoprint` subfolder instead of directly under `octoprint_alexaoctoprint`.
- There is a `plugin` subfolder which, again, as in the previous PR, contains an `OctoPrint-AlexaOctoPrint.zip` file, which in turn contains another copy of the plugin sources, with no automation whatsoever guaranteeing that those sources stay in sync with the ones in `Source/octoprint_alexaoctoprint`.

None of this makes any sense.

## Security

The plugin declares API routes through both `SimpleApiPlugin` and `BlueprintPlugin`. It overrides neither `is_api_protected()` nor `is_blueprint_csrf_protected()`. It sets `is_blueprint_protected()` to `False`, and it never imports the `Permissions` class anywhere.

Thus, anyone can use this plugin's APIs without authentication - APIs that allow, for example, turning the printer on and off, heating the nozzle, moving the axes, and so on.

Some routes take a 40-character "username" as input and use it as a "secret", validating it through the `_is_authorized_hue_username` -> `username_matches` -> `hmac.compare_digest` flow. However, a simple POST to the `/api` route [returns](https://github.com/RICLAMER/AlexaOctoPrint/blob/d9011c0a0f5c65209ef77f01aa370e80efa9fe7e/Source/octoprint_alexaoctoprint/__init__.py#L243) that username in full.

I understand that this design stems from the choice of emulating a Philips Hue bridge: since Alexa has no way to authenticate against the OctoPrint plugin in that scenario, authentication ends up being dropped altogether. But that only means the Philips Hue bridge emulation is the wrong approach to begin with, not that leaving the APIs unauthenticated is acceptable. Other protocols - Matter, for instance - would likely allow an actual pairing between Alexa and the OctoPrint plugin.

## AI usage

In this PR, as in the previous one, the author keeps talking like an excessively verbose clanker, which is another element we have historically disliked in past reviews.

The reason we closed the previous PR was not only the AI slop (which is very much present in the current codebase as well), but also the fact that the author demonstrated no understanding of how GitHub, or coding in general, works. The ["Registering a new plugin"](https://plugins.octoprint.org/help/registering/) page clearly states that we expect maintainers to understand why their plugin works, and not to rely solely on their favorite genAI tool.

---

For the reasons stated above, I am closing this PR as well as the previous one.

PR #1480, opened 2026-08-31
Update PiNozCam listing for the 1.1.0 stable release

DrAlexLiu:
PiNozCam just cut its first true stable release (1.1.0), so this refreshes
the plugin's listing to match current reality:

- **Description and content** rewritten to match the plugin's current
  README (Telegram/Discord remote control, on-device detection, mask
  zones), removing stale references to settings that no longer exist
  (Max Failure Count, old sensitivity thresholds).
- **Hardware speed table** updated with current measured numbers across
  Raspberry Pi 5/4/CM4, Jetson Orin, and the RK3566/RK3588 NPU boards.
- **New screenshots**: Telegram and Discord remote-control views, and the
  current OctoPrint tab/settings panel (the old screenshots were outdated
  PNGs).
- **`os` narrowed to `linux` + `macos`** (was previously also listing
  `windows` and `freebsd`, which the plugin has never supported — it
  ships a prebuilt native binary with no Windows/FreeBSD build).
- **`archive` now points at the `1.1.0` tag** instead of `master.zip`, so
  installs from this listing pull an immutable, released version instead
  of whatever happens to be on the default branch at install time.
- Added the `ai-developed` attribute (the plugin was developed with AI
  assistance) and a `privacypolicy` link, per the registration
  requirements.

Front matter validated locally with
`.github/scripts/validate_front_matter.py --check-id-match --check-asset-match`
(404 files, 0 fails).


_plugins/pinozcam.md
```diff
@@ -3,15 +3,15 @@ layout: plugin
...
+description: AI-driven FREE monitoring for 3D print failures, running entirely on the printer's own board. Requires Linux on ARM32, ARM64 or x86_64.
 author: DrAlexLiu
 license: AGPLv3
 
 date: 2024-03-06
 
 homepage: https://github.com/DrAlexLiu/OctoPrint-PiNozCam
 source: https://github.com/DrAlexLiu/OctoPrint-PiNozCam
-archive: https://github.com/DrAlexLiu/OctoPrint-PiNozCam/archive/master.zip
+archive: https://github.com/DrAlexLiu/OctoPrint-PiNozCam/archive/refs/tags/1.1.0.zip
```
jneilliii: you shouldn't set this directly with full version. the previous url will always install the code in your branch download link the way it was before. should leave as 

Image

```
archive: https://github.com/DrAlexLiu/OctoPrint-PiNozCam/archive/master.zip
```

_plugins/pinozcam.md
```diff
@@ -36,12 +53,19 @@ compatibility:
...
-  - windows
   - macos
-  - freebsd
 
   # Compatible Python version
   python: ">=3.7,<4"
+
+# Developed with the use of Artificial Intelligence.
+attributes:
+- ai-developed
```
jneilliii, line 63: not sure if the cloud attribute should have been flagged because of the telegram integration before. but now, with the discord integration as well looking for @foosel to chime in here.

jneilliii (changes requested):
would like @jacopotediosi just to take another look at the pre-built binary part of this solution. also, one question for @foosel about the cloud attribute tagging and any possible concerns from her.

_plugins/pinozcam.md
DrAlexLiu: Good catch, thanks -- reverted to `archive/master.zip`. `master` now always mirrors the content of the latest tagged release (that's part of what this same release cycle did), so the branch link stays correct without needing a PR every future version.

_plugins/pinozcam.md
foosel: Given that for any other plugins connecting to either Telegram or Discord we have the `cloud` tag applied (at least as far as I can see), this should also be applied here.

It is optional, yes, but this is about transparency.

foosel (changes requested):
As mentioned in my reply to @jneilliii, needs the `cloud` tag too.

_plugins/pinozcam.md
DrAlexLiu: Makes sense, thanks -- added `cloud` to the attributes list.

jacopotediosi:
Will review ASAP, I'm currently under a heavy refactor of the Telegram plugin but should be available again in a few days... Sorry for the delay

jacopotediosi (approved):
Reviewed both plugin sources and this PR, all looks good to me.

Regarding the security of the build and installation method, and the additional packages downloaded from PyPI, I gather that thoughtful choices were made with respect to supply-chain attacks. I found nothing particularly concerning from either a security or a safety standpoint.

[`octoscanner`](https://github.com/jacopotediosi/octoscanner) detected the following minor issues; some look like false positives and overall there is nothing blocking. @DrAlexLiu may want to use them as input for possible improvements.

```
╭──────────────────────────────────────────────────────────────────── Plugin /home/user/Desktop/OctoPrint-PiNozCam ────────────────────────────────────────────────────────────────────╮ │
│ │ ╭────────────────────────────────────────────────────────────────────────── Summary (13 total findings) ───────────────────────────────────────────────────────────────────────────╮ │ │
│ │ │ 6 breaking, 3 security, 2 deprecation, 2 packaging                                                                                                                               │ │ │
│ │ ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯ │ │
│ │ ╭───────────────────────────────────────────────────────────────────────────── Breaking (6 findings) ──────────────────────────────────────────────────────────────────────────────╮ │ │
│ │ │ STG-REM-0007                                                                                                                                                                     │ │ │
│ │ │ The webcam.flipH settings path has been removed in OctoPrint 1.9.0.                                                                                                              │ │ │
│ │ │ Breaking since: 1.9.0                                                                                                                                                            │ │ │
│ │ │ File: octoprint_pinozcam/camera.py:75                                                                                                                                            │ │ │
│ │ │       74 |         return (                                                                                                                                                      │ │ │
│ │ │   >   75 |             bool(self._settings.global_get_boolean(["webcam", "flipH"])),                                                                                             │ │ │
│ │ │       76 |             bool(self._settings.global_get_boolean(["webcam", "flipV"])),                                                                                             │ │ │
│ │ │ Suggestion: Remove usage of settings paths under webcam.flipH.                                                                                                                   │ │ │
│ │ │                                                                                                                                                                                  │ │ │
│ │ │ STG-REM-0007                                                                                                                                                                     │ │ │
│ │ │ The webcam.flipH settings path has been removed in OctoPrint 1.9.0.                                                                                                              │ │ │
│ │ │ Breaking since: 1.9.0                                                                                                                                                            │ │ │
│ │ │ File: octoprint_pinozcam/camera.py:260                                                                                                                                           │ │ │
│ │ │      259 |         """Read webcam flip and rotate flags in one call."""                                                                                                          │ │ │
│ │ │   >  260 |         return (bool(self._settings.global_get_boolean(["webcam", "flipH"])),                                                                                         │ │ │
│ │ │      261 |                 bool(self._settings.global_get_boolean(["webcam", "flipV"])),                                                                                         │ │ │
│ │ │ Suggestion: Remove usage of settings paths under webcam.flipH.                                                                                                                   │ │ │
│ │ │                                                                                                                                                                                  │ │ │
│ │ │ STG-REM-0008                                                                                                                                                                     │ │ │
│ │ │ The webcam.flipV settings path has been removed in OctoPrint 1.9.0.                                                                                                              │ │ │
│ │ │ Breaking since: 1.9.0                                                                                                                                                            │ │ │
│ │ │ File: octoprint_pinozcam/camera.py:76                                                                                                                                            │ │ │
│ │ │       75 |             bool(self._settings.global_get_boolean(["webcam", "flipH"])),                                                                                             │ │ │
│ │ │   >   76 |             bool(self._settings.global_get_boolean(["webcam", "flipV"])),                                                                                             │ │ │
│ │ │       77 |             bool(self._settings.global_get_boolean(["webcam", "rotate90"])),                                                                                          │ │ │
│ │ │ Suggestion: Remove usage of settings paths under webcam.flipV.                                                                                                                   │ │ │
│ │ │                                                                                                                                                                                  │ │ │
│ │ │ STG-REM-0008                                                                                                                                                                     │ │ │
│ │ │ The webcam.flipV settings path has been removed in OctoPrint 1.9.0.                                                                                                              │ │ │
│ │ │ Breaking since: 1.9.0                                                                                                                                                            │ │ │
│ │ │ File: octoprint_pinozcam/camera.py:261                                                                                                                                           │ │ │
│ │ │      260 |         return (bool(self._settings.global_get_boolean(["webcam", "flipH"])),                                                                                         │ │ │
│ │ │   >  261 |                 bool(self._settings.global_get_boolean(["webcam", "flipV"])),                                                                                         │ │ │
│ │ │      262 |                 bool(self._settings.global_get_boolean(["webcam",                                                                                                     │ │ │
│ │ │ Suggestion: Remove usage of settings paths under webcam.flipV.                                                                                                                   │ │ │
│ │ │                                                                                                                                                                                  │ │ │
│ │ │ STG-REM-0009                                                                                                                                                                     │ │ │
│ │ │ The webcam.rotate90 settings path has been removed in OctoPrint 1.9.0.                                                                                                           │ │ │
│ │ │ Breaking since: 1.9.0                                                                                                                                                            │ │ │
│ │ │ File: octoprint_pinozcam/camera.py:77                                                                                                                                            │ │ │
│ │ │       76 |             bool(self._settings.global_get_boolean(["webcam", "flipV"])),                                                                                             │ │ │
│ │ │   >   77 |             bool(self._settings.global_get_boolean(["webcam", "rotate90"])),                                                                                          │ │ │
│ │ │       78 |         )                                                                                                                                                             │ │ │
│ │ │ Suggestion: Remove usage of settings paths under webcam.rotate90.                                                                                                                │ │ │
│ │ │                                                                                                                                                                                  │ │ │
│ │ │ STG-REM-0009                                                                                                                                                                     │ │ │
│ │ │ The webcam.rotate90 settings path has been removed in OctoPrint 1.9.0.                                                                                                           │ │ │
│ │ │ Breaking since: 1.9.0                                                                                                                                                            │ │ │
│ │ │ File: octoprint_pinozcam/camera.py:262                                                                                                                                           │ │ │
│ │ │      261 |                 bool(self._settings.global_get_boolean(["webcam", "flipV"])),                                                                                         │ │ │
│ │ │   >  262 |                 bool(self._settings.global_get_boolean(["webcam",                                                                                                     │ │ │
│ │ │      263 |                                                         "rotate90"])))                                                                                                │ │ │
│ │ │ Suggestion: Remove usage of settings paths under webcam.rotate90.                                                                                                                │ │ │
│ │ ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯ │ │
│ │ ╭───────────────────────────────────────────────────────────────────────────── Security (3 findings) ──────────────────────────────────────────────────────────────────────────────╮ │ │
│ │ │ SEC-API-0004                                                                                                                                                                     │ │ │
│ │ │ Plugin implements BlueprintPlugin but does not override is_blueprint_csrf_protected(). Until OctoPrint 2.0.0, the default implementation returns False, leaving the blueprint's  │ │ │
│ │ │ endpoints exposed to CSRF attacks.                                                                                                                                               │ │ │
│ │ │ File: octoprint_pinozcam/__init__.py:33                                                                                                                                          │ │ │
│ │ │       32 |                                                                                                                                                                       │ │ │
│ │ │   >   33 | class PinozcamPlugin(ConfirmMixin,                                                                                                                                    │ │ │
│ │ │       34 |                      MaskMixin,                                                                                                                                       │ │ │
│ │ │ Suggestion: Override is_blueprint_csrf_protected() to return True and exempt only the endpoints that must not use CSRF protection via the                                        │ │ │
│ │ │ @octoprint.plugin.BlueprintPlugin.csrf_exempt() decorator.                                                                                                                       │ │ │
│ │ │                                                                                                                                                                                  │ │ │
│ │ │ SEC-XSS-0004                                                                                                                                                                     │ │ │
│ │ │ PNotify invocation passing a non-literal value as text without text_escape: true. PNotify v2 (the version shipped with OctoPrint) renders text as raw HTML by default, so any    │ │ │
│ │ │ attacker-controllable data flowing into it (push messages, backend errors, printer responses, etc.) results in DOM XSS.                                                          │ │ │
│ │ │ File: octoprint_pinozcam/static/js/pinozcam.js:822                                                                                                                               │ │ │
│ │ │      821 |             if (problems.length) {                                                                                                                                    │ │ │
│ │ │   >  822 |                 new PNotify({title: 'Not saved',                                                                                                                      │ │ │
│ │ │      823 |                              text: problems.join('  '), type: 'error'});                                                                                              │ │ │
│ │ │ Suggestion: Set text_escape: true in the PNotify options so the value is rendered as plain text. If the value must contain HTML, escape any user-controllable substring via      │ │ │
│ │ │ _.escape(...) before interpolating it into the markup.                                                                                                                           │ │ │
│ │ │                                                                                                                                                                                  │ │ │
│ │ │ SEC-XSS-0004                                                                                                                                                                     │ │ │
│ │ │ PNotify invocation passing a non-literal value as text without text_escape: true. PNotify v2 (the version shipped with OctoPrint) renders text as raw HTML by default, so any    │ │ │
│ │ │ attacker-controllable data flowing into it (push messages, backend errors, printer responses, etc.) results in DOM XSS.                                                          │ │ │
│ │ │ File: octoprint_pinozcam/static/js/pinozcam.js:1087                                                                                                                              │ │ │
│ │ │     1086 |                 self.showTabFor(self.firstInvalidField());                                                                                                            │ │ │
│ │ │   > 1087 |                 new PNotify({title: "PiNozCam setting not saved",                                                                                                     │ │ │
│ │ │     1088 |                              text: problems.join("  "), type: "error"});                                                                                              │ │ │
│ │ │ Suggestion: Set text_escape: true in the PNotify options so the value is rendered as plain text. If the value must contain HTML, escape any user-controllable substring via      │ │ │
│ │ │ _.escape(...) before interpolating it into the markup.                                                                                                                           │ │ │
│ │ ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯ │ │
│ │ ╭──────────────────────────────────────────────────────────────────────────── Deprecation (2 findings) ────────────────────────────────────────────────────────────────────────────╮ │ │
│ │ │ STG-DEP-0002                                                                                                                                                                     │ │ │
│ │ │ The webcam.snapshot settings path is deprecated since OctoPrint 1.9.0 and only kept reachable via a compatibility overlay. Please use the webcam system introduced with 1.9.0,   │ │ │
│ │ │ this compatibility layer will be removed in a future release.                                                                                                                    │ │ │
│ │ │ Deprecated since: 1.9.0                                                                                                                                                          │ │ │
│ │ │ File: octoprint_pinozcam/__init__.py:518                                                                                                                                         │ │ │
│ │ │      517 |             return True                                                                                                                                               │ │ │
│ │ │   >  518 |         if self._settings.global_get(["webcam", "snapshot"]):                                                                                                         │ │ │
│ │ │      519 |             return True                                                                                                                                               │ │ │
│ │ │ Suggestion: Migrate away from webcam.snapshot. Please use the webcam system introduced with 1.9.0, this compatibility layer will be removed in a future release.                 │ │ │
│ │ │                                                                                                                                                                                  │ │ │
│ │ │ STG-DEP-0002                                                                                                                                                                     │ │ │
│ │ │ The webcam.snapshot settings path is deprecated since OctoPrint 1.9.0 and only kept reachable via a compatibility overlay. Please use the webcam system introduced with 1.9.0,   │ │ │
│ │ │ this compatibility layer will be removed in a future release.                                                                                                                    │ │ │
│ │ │ Deprecated since: 1.9.0                                                                                                                                                          │ │ │
│ │ │ File: octoprint_pinozcam/camera.py:430                                                                                                                                           │ │ │
│ │ │      429 |         snapshot = None                                                                                                                                               │ │ │
│ │ │   >  430 |         snapshot_url = self._settings.global_get(["webcam", "snapshot"])                                                                                              │ │ │
│ │ │      431 |         if not snapshot_url:                                                                                                                                          │ │ │
│ │ │ Suggestion: Migrate away from webcam.snapshot. Please use the webcam system introduced with 1.9.0, this compatibility layer will be removed in a future release.                 │ │ │
│ │ ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯ │ │
│ │ ╭───────────────────────────────────────────────────────────────────────────── Packaging (2 findings) ─────────────────────────────────────────────────────────────────────────────╮ │ │
│ │ │ PKG-0001                                                                                                                                                                         │ │ │
│ │ │ Plugin uses the legacy octoprint_setuptools.create_plugin_setup_parameters() in setup.py and likely still relies on the old packaging method.                                    │ │ │
│ │ │ File: setup.py:358                                                                                                                                                               │ │ │
│ │ │      357 |                                                                                                                                                                       │ │ │
│ │ │   >  358 | setup_parameters = octoprint_setuptools.create_plugin_setup_parameters(                                                                                               │ │ │
│ │ │      359 |     identifier=plugin_identifier,                                                                                                                                     │ │ │
│ │ │ Suggestion: Migrate your plugin to pyproject.toml packaging. See https://docs.octoprint.org/en/dev/plugins/migration_pyproject_toml.html.                                        │ │ │
│ │ │                                                                                                                                                                                  │ │ │
│ │ │ PKG-0002                                                                                                                                                                         │ │ │
│ │ │ setup.cfg contains universal = 1, which produces wheels for both Python 2 and 3. Python 2 reached end-of-life in January 2020 and OctoPrint dropped Python 2 support in version  │ │ │
│ │ │ 1.4.0.                                                                                                                                                                           │ │ │
│ │ │ File: setup.cfg:3                                                                                                                                                                │ │ │
│ │ │        2 | #                                                                                                                                                                     │ │ │
│ │ │   >    3 | # `universal = 1` tags the wheel py2.py3-none-any. Both halves are wrong:                                                                                             │ │ │
│ │ │        4 | # the code is Python 3 only, and setup.py's external runtime dependency is                                                                                            │ │ │
│ │ │ Suggestion: Remove the [bdist_wheel] section with universal = 1 from setup.cfg, or delete setup.cfg entirely if it contains nothing else.                                        │ │ │
│ │ ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯ │ │
│ ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯ │
```

PR #1481, opened 2026-09-06
Add XY Nozzle Alignment

chrisns:

XY Nozzle Alignment (`nozzlealign`)

It measures the XY offset between the two nozzles of a dual extruder, using a camera that stands on the print bed looking up, and writes the result to the firmware with `M218`.

The camera can go anywhere on the bed. Nothing about its position is remembered between runs. One run homes the machine, sweeps the bed at Z90 until the toolhead comes into the picture, closes in on it, finds the nozzle bore, sweeps Z for the sharpest view, then steers each nozzle onto the same pixel. Both nozzles on one pixel means the image scale and the lens distortion cancel, so only the difference between the two machine positions matters. The plugin writes the offset, reads it back, and puts the previous value back if the firmware kept something else. A full run takes about four minutes.

https://github.com/chrisns/OctoPrint-NozzleAlign

Yes, and heavily. The code was written by Claude, Anthropic's coding agent, in a long session that I directed against my own printer. I set the goal, chose between approaches, physically re-aimed and focused the camera when the pictures were not good enough, and rejected the versions that did not work on the machine.

I am telling you this plainly because your guide says completely vibe coded plugins are not accepted, and you are entitled to judge this one on that basis. If that means you close this, I understand.

What I can offer against that bar: the README explains the method and every wrong turn that shaped it, including the one that matters most. The first working version measured the same physical nozzle twice, because under T0 the nozzle over the lens is T1's, raised out of use by the lift mechanism. The 0.13 mm it reported was the tilt of that lift, not a tool offset. The plugin now proves which nozzle it is looking at by commanding the tool change and checking that the other bore appears near the same pixel. There are 54 tests that run with no printer and no camera, against a simulated printer with the firmware's real tool change semantics, axis backlash, and a camera that renders both nozzles.

No. It is AGPL-3.0 and free.

No. It talks to the printer over its serial connection and fetches frames from a camera URL the user supplies, both local. It uses no external service, holds no account, and collects no data. There is no tracking code.

#### Further notes

**Safety.** The routine drives the nozzle down onto a camera that stands on the bed. Every Z move passes one floor check, every travel move and every tool change happens at a safe height, the Stop button sends `M410` and wakes the waiting thread, and a failed or stopped run parks the head high, selects T0 and homes. The floor is a setting, and the README tells the user to check it before the first run.

**Firmware.** `M218` is undocumented on Snapmaker and its toolhead module silently resets an X offset outside 26 +/- 1.2 mm to the default, so the plugin refuses to send a value outside that band, rounds to the two decimals the module keeps, and verifies the read-back.

**Tested on** OctoPrint 1.11.8, Python 3.10, Windows, against a Snapmaker A350 dual extruder. It uses the `octoprint.plugin.softwareupdate.check_config` hook, keyed on the plugin identifier and pointing at its own repository. No code runs at install time and the plugin makes no change to the user's system.


jacopotediosi:
@chrisns please stick to the PR template

chrisns:
Redrafted. Sorry I made the pr outside the github ui so missed the template

jacopotediosi:
Thanks, I will review this one in the next few days. I still have a backlog because I'm currently working on another plugin, so please expect a brief delay.

chrisns:
Thanks. No rush!

jacopotediosi:
@chrisns, I want to thank you for your contribution, but I'm afraid I don't have good news here.

There are quite a few things about this plugin that leaves me concerned. Below is a short list of some issues I noticed during a 30-minute review:
- The plugin uses the old installation system instead of `pyproject.toml`.
- The SimpleApi endpoints aren't protected by any kind of `Permission` check.
- `settings.py` contains more than 70 settings, and over half of them aren't configurable from the UI (e.g., the README says "Set the snapshot URL, and the stream URL if you have one", but `stream_url` can only be configured by editing the configuration YAML directly - same goes for `bed_x_min` and the other `bed_*` settings).
- The `bed_*` settings define the bed dimensions and are used in `G1` commands, so if the user doesn't adjust them they can exceed the printer's limits. There's no validation of those values against OctoPrint's Printer Profile.
- There's a fair amount of dead code, e.g. the `motion_min_fraction` setting is never referenced anywhere in the codebase, and the functions `motion_fraction`, `shift_between`, `build_pixel_map_from_shifts` and `offset_command` are never called.
- The plugin's settings page doesn't work: the settings template is registered with `custom_bindings=True`, so it's up to the plugin's js to do the bindings, but it doesn't, and as a result every input stays empty. Looking at the git history, it has never worked. settings don't work

Overall, the code is a mess; some parts are so tangled that I find it hard to believe a human ever read back over them. The [`_search_bed`](https://github.com/chrisns/OctoPrint-NozzleAlign/blob/48be427a741dfde6b7d96a74f891a9718cb8bf7e/octoprint_nozzlealign/routine.py#L264-L499) function has 5 nested functions, 6 levels of indentation and 50+ branches.

I think we should reject this PR, both because the code quality doesn't meet the standards we expect from OctoPrint plugins and because the plugin is entirely vibecoded, and therefore inadmissible under our AI usage policy and the accountability guarantees we require from maintainers.

chrisns:
Thats a very generous review I didn't deserve and fair conclusion.
I'll work to make improvements aligned to your points but no expectation that you'll waste your time looking at it again.
Thanks for maintaining the quality bar 🙏

jacopotediosi:
Thank you for understanding

